PR #1, opened 2018-07-03
Add simplified RpcServer

sisou:
Adds the RPC server to the /lib directory.


lib/rpc-server.js
```diff
@@ -0,0 +1,116 @@
...
+            _ok(message, result) {
+                this._replyTo(message, 'OK', result);
+            }
+
+            /**
+             * @param {MessageEvent} message
+             * @param {CustomError} error
+             */
+            _error(message, error) {
+                this._replyTo(message, 'error',
```
danimoh: Why is OK in capital letters but error not?

PR #4, opened 2018-07-04
Add Key and KeyStore

sisou:


src/lib/KeyStoreIndexeddb.js
```diff
@@ -0,0 +1,193 @@
...
+
+            request.onerror = () => reject(request.error);
+
+            request.onupgradeneeded = event => {
+                /** @type {IDBDatabase} */
+                const db = request.result;
+
+                if (event.oldVersion < 1) {
+                    // Version 1 is the first version of the database.
+                    const store = db.createObjectStore(KeyStore.DB_KEY_STORE_NAME, { keyPath: 'userFriendlyAddress' });
```
Bettelstab: no auto incremented id then?

src/lib/KeyStoreIndexeddb.js
```diff
@@ -0,0 +1,193 @@
...
+        const key = await this._getPlain(userFriendlyAddress);
+        const result = await Key.loadEncrypted(key.encryptedKeyPair, passphrase);
+        result.type = key.type;
+
+        return result;
+    }
+
+    /**
+     * @param {string} userFriendlyAddress
+     * @returns {Promise.<1|2>}
```
Bettelstab: could be replaced by enum (see max/sign-tx branch)

src/lib/KeyStoreIndexeddb.js
sisou: We decided against adding another identifier, because we only have the one-key-to-one-address relation currently. Future keys/addresses that are derived from seed phrases will be stored differently.

src/lib/KeyStoreIndexeddb.js
```diff
@@ -0,0 +1,194 @@
+/**
```
danimoh: Same reviews as for AccountStoreIndexeddb.js apply

src/lib/KeyStoreIndexeddb.js
```diff
@@ -0,0 +1,194 @@
...
+            putTx.onsuccess = () => resolve(putTx.result);
+            putTx.onerror = reject;
+        });
+    }
+
+    /**
+     * @param {string} userFriendlyAddress
+     * @returns {Promise}
+     */
+    async remove(userFriendlyAddress) {
```
danimoh: normalize again

src/lib/KeyStoreIndexeddb.js
```diff
@@ -0,0 +1,194 @@
...
+    putPlain(keyEntry) {
+        return this._putPlain(keyEntry);
+    }
+
+    /**
+     * @param {object} keyEntry
+     * @returns {Promise<any>}
+     */
+    async _putPlain(keyEntry) {
+        const db = await this.connect();
```
danimoh: if address is used as key, it should be normalized (e.g. no spaces, all caps)

src/lib/AccountStoreIndexeddb.js
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * @param {string} dbName
+     * @constructor
+     */
+    constructor(dbName = 'accounts') {
+        this._dbName = dbName;
+        /** @type {IDBDatabase|undefined} */
+        this._db;
+        this._connected = false;
```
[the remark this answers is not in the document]
sisou: Is a closed DB falsey?

src/lib/Key.js
```diff
@@ -0,0 +1,250 @@
...
+        this.type = type;
+    }
+
+    /**
+     * Sign a generic message
+     * @param {string} message - A utf-8 string
+     * @returns {object}
+     */
+    signMessage(message) {
+        message = 'nimiq_msg_' + message;
```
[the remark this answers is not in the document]
sisou: To prevent the user from accidentally signing arbitrary data (like transactions), we currently only allow strings and prepend a message identifier. Ethereum and Bitcoin do the same (in proper implementations). These prefixes are hidden from the user usually when displaying the signed data.

src/lib/AccountStoreIndexeddb.js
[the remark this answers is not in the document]
danimoh: no, but you would set it to null when closing the db

sisou:
@danimoh Review addressed in da446febac6e9046249f03c593707202d1545e28.

PR #5, opened 2018-07-04
I18n

svub:
i18n lib v1 + Jasmine tests setup.


sisou:
All the TS errors need to be surpressed (in the Jasmine lib) or fixed.

tests/jasmine-3.1.0/boot.js
```diff
@@ -0,0 +1,134 @@
+/**
```
danimoh: Do we really want to have all this jasmine code in our repo?

src/lib/i18n.js
```diff
@@ -0,0 +1,88 @@
...
+        this._current = language;
+        this.translateDom();
+    }
+
+    currentLanguage() {
+        return this._current;
+    }
+}
+
+var i18n = new I18n();
```
danimoh: mh, global i18n object vs global I18n class with static methods. I think I prefer the latter

PR #12, opened 2018-07-06
Extract a common RequestApi

sisou:
Introduces a common `PageApi` class, from which page API classes should inherit. This allows common operations to be DRY and allows to hook into the `resolve` event to write the cookie on iOS/Safari.

### DIFF
--- a/src/request/sign-transaction/SignTransactionApi.js
+++ b/src/request/sign-transaction/SignTransactionApi.js
@@ -23,8 +23,8 @@ class SignTransactionApi extends RequestApi {
             ? new SignTransactionWithPassphrase(txRequest)
             : new SignTransactionWithPin(txRequest);
 
-        handler.on('result', /** @param {any} result */ (result) => this.resolve(result));
-        handler.on('error', /** @param {Error} error */ (error) => this.reject(error));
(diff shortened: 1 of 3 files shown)


src/pages/PageApi.js
```diff
@@ -0,0 +1,51 @@
+class PageApi {
```
Bettelstab: naming: RequestApi to keep consistent with name change of pages folder to requests

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -17,15 +19,12 @@ class SignTransactionApi {
...
-            const handler = keyType === EncryptionType.HIGH
-                ? new SignTransactionWithPassphrase(txRequest)
-                : new SignTransactionWithPin(txRequest);
-
-            handler.on('result', /** @param {any} result */ (result) => resolve(result));
-            handler.on('error', /** @param {Error} error */ (error) => reject(error));
-        });
+        handler.on('result', /** @param {any} result */ (result) => this.resolve(result));
+        handler.on('error', /** @param {Error} error */ (error) => this.reject(error));
     }
```
svub: No round brackets needed for lambdas with on parameter. :P

PR #16, opened 2018-07-06
IFrame API

sisou:
expose `list()` and `migrateAccountsToKeys()` as methods in an iframe.

Adds `BrowserDetection` and `CookieJar` utility libraries.

Sets a cookie flag when `migrateAccountsToKeys()` is requested under iOS/Safari, so that popups can handle the migration when loaded the next time.

### DIFF
--- a/src/lib/CookieJar.js
+++ b/src/lib/CookieJar.js
@@ -16,7 +16,7 @@ class CookieJar {
         // Legacy cookie
         if (listFromAccountStore) {
             const match = document.cookie.match(new RegExp('accounts=([^;]+)'));
-            if (match) {
+            if (match && match[1]) {
                 const decoded = decodeURIComponent(match[1]);
                 return JSON.parse(decoded);
             }
@@ -25,7 +25,7 @@ class CookieJar {
 
         const match = document.cookie.match(new RegExp('k=([^;]+)'));
 
-        if (match) {
+        if (match && match[1]) {
             return this._decodeCookie(match[1]);
         }
 
--- a/src/lib/CookieJar.js
+++ b/src/lib/CookieJar.js
@@ -10,7 +10,7 @@ class CookieJar {
     }
 
     /**
-     * @param {boolean} [listFromAccountStore] - Deprecated, used for reading keys from old database
+     * @param {boolean} [listFromAccountStore] - @deprecated Only for database migration
      */
     static eat(listFromAccountStore) {
         // Legacy cookie
--- a/src/lib/CookieJar.js
+++ b/src/lib/CookieJar.js
@@ -40,8 +40,8 @@ class CookieJar {
 
         for (const keyInfo of keys) {
             const address = Nimiq.Address.fromUserFriendlyAddress(keyInfo.userFriendlyAddress);
-            const encodedKey = /** @type {string} */ (address.toBase64());
-            str += keyInfo.type.toString() + encodedKey.slice(0, 27);
+            const encodedAddress = /** @type {string} */ (address.toBase64());
+            str += keyInfo.type.toString() + encodedAddress.slice(0, 27);
         }
 
         return str;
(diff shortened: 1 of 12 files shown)


src/lib/CookieJar.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * @param {string} str
+     * @returns {KeyInfo[]}
+     */
+    static _decodeCookie(str) {
+        if (!str) return [];
+
+        if (str.length % 28 !== 0) throw new Error('Malformed cookie');
```
Bettelstab: maybe add custom error class?

src/request/RequestApi.js
```diff
@@ -39,6 +39,11 @@ class RequestApi {
      * @param {object} request
      */
     async request(request) {
+        // Deprecated, only for database migration
+        // TODO Maybe only check on iOS/Safari?
+        // Would require to load the BrowserDetection class for every request.
+        await this._checkForMigrationFlag();
```
danimoh: BrowserDetection is used anyways in resolve?

src/request/RequestApi.js
```diff
@@ -73,4 +84,39 @@ class RequestApi {
...
+    }
+
+    /**
+     * @deprecated Only for database migration
+     */
+    static async doMigrateAccountsToKeys() {
+        const accountStore = AccountStore.instance;
+        const keyStore = KeyStore.instance;
+
+        const keys = await accountStore.dangerouslistPlain();
```
danimoh: should be dangerousListPlain (capital L)

src/request/RequestApi.js
```diff
@@ -73,4 +84,39 @@ class RequestApi {
...
+                encryptedKeyPair: key.encryptedKeyPair,
+                userFriendlyAddress: key.userFriendlyAddress,
+                // Translate between old text type and new number type
+                type: /** @type {EncryptionType} */ (key.type === 'high' ? EncryptionType.HIGH : EncryptionType.LOW)
+            };
+            await keyStore.putPlain(keyEntry);
+        }
+
+        // FIXME Uncomment after/for testing
+        // await accountStore.drop();
```
danimoh: Might want to test that keyStore has been written correctly

src/request/RequestApi.js
```diff
@@ -73,4 +84,39 @@ class RequestApi {
...
+        return true;
```
danimoh: Can also delete old accounts cookie here

src/request/iframe/IFrameApi.js
```diff
@@ -0,0 +1,41 @@
+class IFrameApi {
```
danimoh: Might rename the RequestApi to PopupApi becase that's really what it is? And then move doMigrateAccounts to keys to KeyStore as it gets called from IframeApi also?

src/lib/CookieJar.js
danimoh: Mh, I dont know. That error should never occur

sisou: It's also not a user error, but a system error, if it _does_ happen.

src/request/RequestApi.js
sisou: I trust that it throws an exception if writing fails.

src/request/RequestApi.js
```diff
@@ -40,9 +40,9 @@ class RequestApi {
      */
     async request(request) {
         // Deprecated, only for database migration
-        // TODO Maybe only check on iOS/Safari?
-        // Would require to load the BrowserDetection class for every request.
-        await this._checkForMigrationFlag();
+        if (BrowserDetection.isIos() || BrowserDetection.isSafari() && this._hasMigrateFlag()) {
```
danimoh: Have to put (BrowserDetection.isIos() || BrowserDetection.isSafari()) into parentheses

PR #18, opened 2018-07-06
Improve Nimiq types

Bettelstab:


src/lib/Key.js
```diff
@@ -232,7 +240,7 @@ class Key { // eslint-disable-line no-unused-vars
             key = Nimiq.BufferUtils.fromAscii(key);
         }
 
-        await this.keyPair.lock(key);
+        this.keyPair.lock(key);
```
danimoh: why without the await?

src/lib/Key.js
```diff
@@ -247,16 +255,20 @@ class Key { // eslint-disable-line no-unused-vars
             key = Nimiq.BufferUtils.fromAscii(key);
         }
 
-        await this.keyPair.unlock(key);
+        this.keyPair.unlock(key);
```
danimoh: why without the await?

src/lib/KeyStoreIndexeddb.js
```diff
@@ -88,6 +90,7 @@ class KeyStore {
      * @param {Key} key
      * @param {Uint8Array | string} passphrase
      * @param {Uint8Array | string} [unlockKey]
+     * @returns {Promise}
```
danimoh: Promise<void>
(also for other methods in this class)

PR #23, opened 2018-07-08
Add additional checks when receiving a message in the RpcServer

sisou:


src/lib/RpcServer.js
```diff
@@ -61,6 +61,15 @@ class RpcServer { // eslint-disable-line no-unused-vars
...
             _receive(message) {
                 try {
+                    // Cannot reply to a message that has no source window
+                    if (!message.source) return;
+
+                    // Ignore messages without a command
+                    if (!message.data.command) return;
+
+                    // Ignore messages without an ID (except 'ping' commands)
+                    if (!message.data.id && message.data.command !== 'ping') return;
```
svub: maybe ping should have an ID too than we can drop this exception

PR #24, opened 2018-07-08
Add ESLint errors for incomplete JSDocs

sisou:


src/lib/I18n.js
```diff
@@ -22,10 +23,11 @@ class I18n { // eslint-disable-line no-unused-vars
     static translateDom(dom = document.body, enforcedLanguage) {
         const language = enforcedLanguage ? this.getClosestSupportedLanguage(enforcedLanguage) : this.language;
 
+        /* eslint-disable-next-line valid-jsdoc */ // Multi-line descriptions are not valid JSDoc, apparently
         /**
          * @param {string} tag
-         * @param {(element: HTMLElement, translation: string) => void} callback
-         *    - callback(element, translation) for each matching element
+         * @param {(element: HTMLElement, translation: string) => void} callback - callback(element, translation) for
+         * each matching element
```
danimoh: If it's not valid jsdoc why do it?

sisou:
FYI, `@returns` statements have been made optional for async functions a few days ago with this commit: https://github.com/eslint/eslint/commit/196c1021c070e52259f2ceccc4f141bdb7456c60. This means that we have to manually check if all async functions have the proper return type (`Promise<TheType>`). Sadly, this has not been made configurable.

If we would turn on `requireReturns: true`, then **_all_** methods would require a `@returns` annotation, even if they return only `void`.

sisou:
Ups...

PR #25, opened 2018-07-09
Dependency validation tooling

sisou:
This adds a script `dependencyValidator.js` under the `/tools` folder, that parses each requests' API class file and builds the dependency tree from it, by parsing the `/* global ... */` statements in each JS file (required by the ESLint `no-undef` rule, after moving the globals out of the `.eslintrc.json` file).

After the dependency list was assembled, the pathes to the dependencies are compared with the `script` tags found in each requests' `index.html` file. Missing dependencies and unneeded script tags are then displayed on the console.


src/lib/I18n.js
```diff
@@ -1,3 +1,5 @@
+/* global TRANSLATIONS */ // eslint-disable-line no-unused-vars
```
sisou: This is required to add the dependency to the translations themselves. (Ironically, ESLint says it's an unused variable xD)

src/common.js
```diff
@@ -1,3 +1,7 @@
+/* global Nimiq */
+/* global KeyStore */
+/* global RpcServer */
+
```
Bettelstab: I would prefer a one-liner (global Nimiq, Keystore...)

sisou: This could become a very long line, e.g. here: https://github.com/nimiq/keyguard-next/pull/25/files#diff-941ff894131b0eb2fea1412f106d8d2eR1

This would break with the 120 character limit.

I like them better under each other, just like classic import statements.

danimoh: Yes, multiple lines is better

PR #29, opened 2018-07-10
Implement message signing on Key

sisou:
This PR improves the `signMessage()` function on the Key class. It does not yet implement the request type.

### DIFF
--- a/Nimiq.d.ts
+++ b/Nimiq.d.ts
@@ -41,7 +41,35 @@ declare namespace Nimiq {
         static toHex(buf: Uint8Array): string
     }
 
-    class SerialBuffer extends Uint8Array {}
+    class SerialBuffer extends Uint8Array {
+        constructor(bufferOrArrayOrLength: any)
+        subarray(start: number, end: number): Uint8Array
+        readPos: number
+        writePos: number
+        reset(): void
+        read(length: number): Uint8Array
+        write(array: any): void
+        readUint8(): number
+        writeUint8(value: number): void
+        readUint16(): number
+        writeUint16(value: number): void
+        readUint32(): number
+        writeUint32(value: number): void
+        readUint64(): number
+        writeUint64(value: number): void
+        readVarInt(): number
+        writeVarInt(value: number): void
+        static varUintSize(value: number): number
+        readFloat64(): number
+        writeFloat64(value: number): void
+        readString(length: number): string
+        writeString(value: string, length: number): void
+        readPaddedString(length: number): string
+        writePaddedString(value: string, length: number): void
+        readVarLengthString(): string
+        writeVarLengthString(value: string): void
+        static varLengthStringSize(value: string): number
+    }
 
     class Synchronizer {}
     class MultiSynchronizer {}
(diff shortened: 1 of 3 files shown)


Nimiq.d.ts
```diff
@@ -41,7 +41,35 @@ declare namespace Nimiq {
...
 
-    class SerialBuffer extends Uint8Array {}
+    class SerialBuffer extends Uint8Array {
+        constructor(bufferOrArrayOrLength: any)
+        subarray(start: number, end: number): Uint8Array
+        readPos: number
+        writePos: number
+        reset(): void
+        read(length: number): Uint8Array
+        write(array: any): void
```
Bettelstab: any => SerialBuffer?

Nimiq.d.ts
```diff
@@ -41,7 +41,35 @@ declare namespace Nimiq {
         static toHex(buf: Uint8Array): string
     }
 
-    class SerialBuffer extends Uint8Array {}
+    class SerialBuffer extends Uint8Array {
+        constructor(bufferOrArrayOrLength: any)
```
Bettelstab: any =>  SerialBuffer | number ?

Nimiq.d.ts
```diff
@@ -41,7 +41,35 @@ declare namespace Nimiq {
...
+        subarray(start: number, end: number): Uint8Array
+        readPos: number
+        writePos: number
+        reset(): void
+        read(length: number): Uint8Array
```
Bettelstab: Unint8Array => SerialBuffer?

Nimiq.d.ts
sisou: I took the annotations directly from the core, and they are using `{*}`. So we have to honor that here.

Nimiq.d.ts
sisou: No, the core is returning a `Uint8Array` here.

Nimiq.d.ts
sisou: Well, it has the same options as a Uint8Array: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array

src/lib/Key.js
```diff
@@ -78,14 +78,48 @@ class Key { // eslint-disable-line no-unused-vars
...
     signMessage(message) {
-        message = `nimiq_msg_${message}`;
-        const msgBytes = Utf8Tools.stringToUtf8ByteArray(message);
-        const signature = Nimiq.Signature.create(this.keyPair.privateKey, this.keyPair.publicKey, msgBytes);
-        return { message, signature };
+        /**
+         * Adding a prefix to the message makes the calculated signature recognisable as
+         * a Nimiq specific signature. This prevents misuse where a malicious request can
+         * sign arbitrary data (e.g. a transaction) and use the signature to impersonate
+         * the victim. (https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign)
```
danimoh: :+1:

src/lib/Key.js
```diff
@@ -78,14 +78,48 @@ class Key { // eslint-disable-line no-unused-vars
...
+
+        const bufLength = 1 /* prefixBytes length value */
+            + prefixLength
+            + 1 /* msgBytes length value */
+            + msgLength;
+
+        // Construct buffer
+        const buf = new Nimiq.SerialBuffer(bufLength);
+        buf.writeUint8(prefixLength);
+        buf.write(prefixBytes);
```
danimoh: There is also SerialBuffer.writeVarLengthStringth

sisou: Yes, but it only supports ASCII characters and we want to support the full UTF-8 range.

sisou: Also, I need to know the length of the prefix string anyway to be able to size the buffer (yes, hardcoding would work but that makes it more inflexible).

PR #31, opened 2018-07-10
Sign Transaction request

sisou:
Supports basic transactions, no styling yet.


src/request/sign-transaction/SignTransaction.js
```diff
@@ -11,29 +11,44 @@ class SignTransaction { // eslint-disable-line no-unused-vars
...
+                recipient: tx.recipient.toUserFriendlyAddress(),
+                value: Nimiq.Policy.satoshisToCoins(tx.value),
+                fee: Nimiq.Policy.satoshisToCoins(tx.fee),
+                validityStartHeight: tx.validityStartHeight,
+                signature: signatureProof.signature.serialize(),
+                hash: tx.hash().toBase64(),
+            };
+        }
+
+        if (txRequest.type === TransactionType.EXTENDED) {
```
svub: extended not implemented as it seems. suggest to throw an exception "not implemented"

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -4,26 +4,56 @@
...
 class SignTransactionApi extends PopupApi { // eslint-disable-line no-unused-vars
     /**
      * @param {TransactionRequest} txRequest
      */
     async onRequest(txRequest) {
+        // Validation
+        if (!txRequest) throw new Error('Invalid request');
+
+        if (txRequest.type === TransactionType.EXTENDED) {
+            throw new Error('Extended transaction signing is not yet implemented');
```
svub: how about using a switch case here?

svub: for basic, extended, and else

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -4,26 +4,56 @@
...
             throw new AmountTooSmallError();
         }
 
         if (txRequest.network !== Nimiq.GenesisConfig.NETWORK_NAME) {
             throw new NetworkMissmatchError(txRequest.network, Nimiq.GenesisConfig.NETWORK_NAME);
         }
+        // End validation
 
+        // Normalization
         txRequest.value = Nimiq.Policy.coinsToSatoshis(txRequest.value);
```
svub: can we know for sure that the request value is not in satoshis already?

src/request/sign-transaction/index.html
```diff
@@ -31,17 +33,68 @@
     <link rel="stylesheet" href="../../common.css">
     <link rel="stylesheet" href="../../components/PinInput.css">
     <link rel="stylesheet" href="../../components/PassphraseInput.css">
+    <link rel="stylesheet" href="SignTransaction.css">
 </head>
 <body>
+    <div class="header-top">
+        <div class="nimiq-app-name">
+            <nimiq-logo>NIMIQ KEYGUARD</nimiq-logo>
```
svub: <nimiq-logo> -- not valid HTML

src/request/sign-transaction/index.html
```diff
@@ -31,17 +33,68 @@
...
-        <!-- For SignTransactionWithPin -->
+        <!-- For SignTransactionWithPin and content for SignTransactionWithPassphrase -->
         <div id="transaction-data" class="page">
+            <h1 data-i18n="sign-tx-header">Authorize Transaction</h1>
+            <h2 data-i18n="sign-tx-byline">Please confirm the following transaction:</h2>
+
+            <div class="transaction">
+                <div class="center">
+                    <div id="sender-identicon"></div>
+                    <i class="arrow material-icons">arrow_forward</i>
```
svub: do we use material icons?

src/request/sign-transaction/SignTransaction.js
sisou: The check and thus exception throwing is at the start of the request in `SignTransactionApi`.

src/request/sign-transaction/SignTransactionApi.js
sisou: No, but it's defined in the API to be in NIM. Maybe better API design would be to expect in sat, that way an error would mean sending less, not sending more?

svub: yep, agree.

src/request/sign-transaction/SignTransaction.js
```diff
@@ -1,8 +1,69 @@
...
+/* global Identicon */
+
 class SignTransaction { // eslint-disable-line no-unused-vars
+    /**
+     * @param {TransactionRequest} txRequest
+     * @returns {HTMLDivElement}
+     */
+    fillTransactionDetails(txRequest) {
+        /** @type {HTMLDivElement} */
+        const $transaction = (document.querySelector('#transaction-data .transaction'));
```
danimoh: are the parentheses needed?

src/request/sign-transaction/SignTransaction.js
```diff
@@ -1,8 +1,69 @@
...
+        }
+
+        $senderAddress.textContent = txRequest.sender;
+        $recipientAddress.textContent = txRequest.recipient;
+
+        $value.textContent = Nimiq.Policy.satoshisToCoins(txRequest.value).toString();
+        if (txRequest.fee > 0) {
+            $fee.textContent = Nimiq.Policy.satoshisToCoins(txRequest.fee).toString();
+            const $feeSection = $transaction.querySelector('.fee-section');
+            if ($feeSection) $feeSection.classList.remove('display-none');
```
danimoh: why is this check necessary?

src/request/sign-transaction/SignTransaction.js
```diff
@@ -11,29 +72,44 @@ class SignTransaction { // eslint-disable-line no-unused-vars
...
+
+            return {
+                type: TransactionType.BASIC,
+                sender: tx.sender.toUserFriendlyAddress(),
+                signerPubKey: signatureProof.publicKey.serialize(),
+                recipient: tx.recipient.toUserFriendlyAddress(),
+                value: Nimiq.Policy.satoshisToCoins(tx.value),
+                fee: Nimiq.Policy.satoshisToCoins(tx.fee),
+                validityStartHeight: tx.validityStartHeight,
+                signature: signatureProof.signature.serialize(),
```
danimoh: why unserialize and then serialize the proof again?

src/request/sign-transaction/SignTransaction.js
```diff
@@ -11,29 +72,44 @@ class SignTransaction { // eslint-disable-line no-unused-vars
...
+
+            const signatureProof = Nimiq.SignatureProof.unserialize(new Nimiq.SerialBuffer(tx.proof));
+
+            return {
+                type: TransactionType.BASIC,
+                sender: tx.sender.toUserFriendlyAddress(),
+                signerPubKey: signatureProof.publicKey.serialize(),
+                recipient: tx.recipient.toUserFriendlyAddress(),
+                value: Nimiq.Policy.satoshisToCoins(tx.value),
+                fee: Nimiq.Policy.satoshisToCoins(tx.fee),
```
danimoh: why not leave them as satoshis?

src/request/sign-transaction/SignTransactionApi.js
danimoh: Sticking to satoshis is generally a good idea.
I was also confused by value and fee being in satoshis in the `txRequest` in `_doSignTransaction` but in NIM here

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -4,26 +4,60 @@
...
      * @param {TransactionRequest} txRequest
      */
     async onRequest(txRequest) {
+        // Validation
+        if (!txRequest) throw new Error('Invalid request');
+
+        if (txRequest.type === TransactionType.EXTENDED) {
+            throw new Error('Extended transaction signing is not yet implemented');
+        } else if (txRequest.type === TransactionType.BASIC) {
+            if (txRequest.sender !== txRequest.signer) {
```
danimoh: might want to normalize

src/request/sign-transaction/SignTransaction.css
```diff
@@ -0,0 +1,58 @@
...
+.transaction .label {
+    margin-bottom: 1.25rem;
+}
+
+.transaction .address {
+    padding: 1rem 0;
+    display: inline;
+}
+
+.passphrase-input {
```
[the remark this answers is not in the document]
sisou: Firstly, the passphrase-input is not in `.transaction`. Second, the CSS is unique to this request, thus does not leak into other requests and is thus safe to unscope (except when we minify all CSS into only one single file). I scoped it into the request container now.

src/request/sign-transaction/SignTransaction.js
sisou: Yes, for type-casting for typescript, as otherwise the `querySelector` result can be `null`.

src/request/sign-transaction/SignTransaction.js
sisou: Because I forgot to typecast the `$feeSection`. Fixed it.

src/request/sign-transaction/SignTransaction.js
sisou: Because the `proof` is stored in a `Nimiq.Transaction` object as an `Uint8Array`, and thus needs to be deserialized to be a `Nimiq.SignaturProof` (line 84). To transfer it via a postMessage, it needs to be a primitive type again, and postMessage can work with Uint8Arrays, so to transfer over RPC we serialize the signature again into an Uint8Array.

src/request/sign-transaction/SignTransaction.js
sisou: It's a basic API decision. Originally we designed the API to be as user-friendly as possible, thus it always used NIM for values. We can change this, but then need to make it the same for all interfaces. I for one am in favour of using NIM for values, but it can be a majority decision.

src/request/sign-transaction/SignTransactionApi.js
sisou: I moved the normalization to before these checks.

src/request/PopupApi.js
```diff
@@ -55,11 +59,15 @@ class PopupApi { // eslint-disable-line no-unused-vars
...
+        return new Promise(async (resolve, reject) => {
             this._resolve = resolve;
             this._reject = reject;
 
-            this.onRequest(request);
+            try {
+                await this.onRequest(request);
+            } catch (e) {
+                this.reject(e);
+            }
```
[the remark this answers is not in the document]
sisou: I like it =)

src/request/PopupApi.js
```diff
@@ -38,6 +38,10 @@ class PopupApi { // eslint-disable-line no-unused-vars
 
         /** @type {Function} */
         this._reject = () => { throw new Error('Method not defined'); };
+
+        window.addEventListener('beforeunload', () => {
+            // this.reject(new Error('Keyguard popup closed'));
+        });
```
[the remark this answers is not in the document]
sisou: Yes. Commented for testing and accidentally commited.

client/index.html
```diff
@@ -6,18 +6,44 @@
         iframe {
             display: none;
         }
+
```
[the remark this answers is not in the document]
sisou: I respectfully disagree. It's not a demo in the same sense that component demos are demos. This is a tool to test the keyguard and all it's request in it's totality, not just a single request or component. It belongs to the keyguard-client, which itself is contained in this `/client` directory and does not leak anything out to the rest of the keyguard repository.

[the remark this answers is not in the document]
Bettelstab: In my understanding it's a demo of using the keyguard client. So somehow it should be in some demo folder.
I vote for moving the whole keyguard client code to the accounts repo, as it is the only place where it is used. Or we could put it in a npm package and make an own repo for that.

Keyguard.d.ts
```diff
@@ -19,36 +19,38 @@ type AccountEntry = AccountInfo & {
...
+    recipientType?: Nimiq.Account.Type
     signer: string
+    signerLabel?: string
     value: number
     fee: number
     network: string
     validityStartHeight: number
+    extraData?: string,
+    mockKeyType?: EncryptionType
 }
```
Bettelstab: extraData should only exist with TransactionType.EXTENDED. Why not two dedicated types?

src/request/sign-transaction/SignTransaction.js
Bettelstab: One argument for using satoshis as unit is that if one accidentally sends 1000 satoshis when his intention was to send 1000 NIM, it's not the same problem as the other way around.

Bettelstab: However, I think coins can be a vary valid choice too. We should decide this apart from this PR.

src/request/sign-transaction/SignTransaction.js
```diff
@@ -11,29 +92,63 @@ class SignTransaction { // eslint-disable-line no-unused-vars
...
-            validityStartHeight: tx.validityStartHeight,
-            signature: signatureProof.signature.serialize(),
-            extraData: Utf8Tools.utf8ByteArrayToString(tx.data),
-            hash: tx.hash().toBase64(),
-        };
+        // The check for a valid transaction type is already done in onRequest in SignTransactionApi,
+        // but Typescript complains if this function has a path with no return.
+        // Thus logically, this should never be hit.
+        // The alternative would be to remove the check for the extended type and just run that if type is not basic.
+        throw new Error('Invalid transaction type');
```
Bettelstab: Probably you can avoid this if you type txRequest as union type as I did before.

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -4,26 +4,71 @@
...
     /**
      * @param {TransactionRequest} txRequest
      */
     async onRequest(txRequest) {
-        if (Nimiq.Policy.coinsToSatoshis(txRequest.value) < 1) {
+        // Validation
+        if (!txRequest) throw new Error('Invalid request');
+
+        // These checks throw an exception on failure
+        AddressUtils.isUserFriendlyAddress(txRequest.sender);
```
Bettelstab: ...which is quite strange IMHO

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -4,26 +4,71 @@
...
+                throw new Error('Sender must be signer for basic transactions');
+            }
+
+            if (txRequest.extraData) {
+                throw new Error('Cannot add extraData to basic transaction');
+            }
+
+            break;
+        case TransactionType.EXTENDED:
+            // Validate extended transactions
```
Bettelstab: what exactly has to be validated?

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -4,26 +4,71 @@
...
 
-        txRequest.value = Nimiq.Policy.coinsToSatoshis(txRequest.value);
-        txRequest.fee = Nimiq.Policy.coinsToSatoshis(txRequest.fee);
-
-        // get key from keystore
-        const keyStore = KeyStore.instance;
-        const keyType = await keyStore.getType(txRequest.sender);
+        // Get signer key type
+        let keyType;
+        if (txRequest.mockKeyType) keyType = txRequest.mockKeyType;
```
Bettelstab: what about this mockKeyType? What is it for?

Keyguard.d.ts
sisou: I had type problems doing that. You are welcome to check out the branch and try to do it.

src/request/sign-transaction/SignTransactionApi.js
sisou: I couldn't think of any, that's why it's still empty ;)

src/request/sign-transaction/SignTransactionApi.js
sisou: For testing from the KeyguardClient, as long as we do not have stored keys.

src/request/sign-transaction/SignTransactionApi.js
sisou: There's also `AddressUtils.isValidAddress()` which returns a boolean. But it uses the above internally, and I didn't want to create a huge `if()` statement or three checks, thus I thought I'd just use the underlying error-throwing function.

client/index.html
```diff
@@ -45,6 +47,24 @@
...
+                    fee: 0.00138,
+                    network: 'test',
+                    validityStartHeight: 100000,
+                    extraData: 'Finally, I can move my stash!',
+                    mockKeyType: EncryptionType.HIGH,
+                });
+            }
         </script>
 </body>
 </html>
```
svub: Can't this entire file be moved to demos? It's even titled "KeyguardClient Demo" ;)

src/request/sign-transaction/SignTransactionApi.js
svub: Is there no way to put this into testing code? It pollutes the general code if it's for testing purpose only.

client/index.html
sisou: Will be removed in a separate PR.

styppo:
Merged 06d04d8aad58abff04a10e7868745b0f9c394e06.

PR #32, opened 2018-07-10
Add KeyStore tests, some fixes

danimoh:


src/lib/AccountStoreIndexeddb.js
```diff
@@ -22,26 +22,24 @@ class AccountStore {
      * @param {string} dbName
      * @constructor
      */
-    constructor(dbName = 'accounts') {
+    constructor(dbName = AccountStore.ACCOUNT_DATABASE) {
         this._dbName = dbName;
         this._dropped = false;
-        /** @type {IDBDatabase | null} */
-        this._db = null;
+        this._dbPromise = null;
```
sisou: Not too happy that it doesn't have a type anymore. You probably didn't notice now, but there was most likely no auto-completion anymore when writing `this._dbPromise. ...`. I think it was better when there was a clear variable that had the `IDBDatabase` type.

sisou: Also, just like in the `KeyStore`, you don't need to initialize it here I think.

src/lib/KeyStoreIndexeddb.js
```diff
@@ -221,18 +223,16 @@ class KeyStore {
...
         });
 
-        // FIXME Uncomment after/for testing
+        // FIXME Uncomment after/for testing (and also adapt KeyStoreIndexeddb.spec.js)
         // await accountStore.drop();
 
         if (BrowserDetection.isIos() || BrowserDetection.isSafari()) {
             // Delete migrate cookie
-            document.cookie = 'migrate=0;expires=0';
+            document.cookie = 'migrate=0; expires=Thu, 01 Jan 1970 00:00:01 GMT;';
```
sisou: What's wrong with `expires=0`? It works the exact same way.

src/lib/KeyStoreIndexeddb.js
```diff
@@ -203,7 +205,7 @@ class KeyStore {
      * stored the existing account labels. Both the 'accounts' database and cookie are
      * deleted afterwards.
      *
-     * @returns {Promise<boolean>}
+     * @returns {Promise<void>}
```
sisou: Why change it?

tests/DummyData.spec.js
```diff
@@ -18,18 +18,24 @@ Dummy.encryptionPassword = 'password';
...
+Dummy.keyDatabaseEntries = [
+    Object.assign({}, Dummy.keyInfo[0], {encryptedKeyPair: Dummy.encryptedKeyPairs[0]}),
+    Object.assign({}, Dummy.keyInfo[1], {encryptedKeyPair: Dummy.encryptedKeyPairs[1]})
+];
+
 /** @type {AccountInfo[]} */
 Dummy.deprecatedAccountInfo = Dummy.keyInfo.map((keyInfo, i) => ({
     userFriendlyAddress: keyInfo.userFriendlyAddress,
-    type: keyInfo.type === 1? 'low' : 'high',
+    type: keyInfo.type === EncryptionType.LOW? 'low' : 'high',
```
sisou: Missing space before the `?`.

tests/DummyData.spec.js
```diff
@@ -87,28 +93,39 @@ Dummy.accountStore = new AccountStore(Dummy.DUMMY_ACCOUNT_DATABASE_NAME);
...
         });
     }
 
     beforeAll(async () => {
         Nimiq.GenesisConfig.test();
         await Promise.all([
             Nimiq.WasmHelper.doImportBrowser(),
-            createDummyAccountStore()
+            createDummyAccountStore(),
+            createDummyKeyStore()
```
sisou: I think the creation and dropping of the database should be done `beforeEach`/`afterEach` test to have a clean slate. Jasmine executes the tests of the same `describe` after each other.

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+    it('connection can be opened and closed', async () => {
+        const db = await KeyStore.instance.connect();
+        expect(KeyStore.instance._dbPromise).toBeTruthy();
+        expect(db.name).toBe(Dummy.DUMMY_KEY_DATABASE_NAME);
+        await KeyStore.instance.close();
+        expect(KeyStore.instance._dbPromise).toBeNull();
+    });
+
+    it('can get a plain key', async () => {
```
sisou: Since you are testing two keys, should rename to "can get plain keys".

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
+describe('KeyStore', () => {
+    it('is a singleton', () => {
+        const instance1 = KeyStore.instance;
+        const instance2 = KeyStore.instance;
+        expect(instance1).toBe(instance2);
+    });
+
+    it('connection can be opened and closed', async () => {
```
sisou: Naming should follow the "it ..." syntax, for example **it** `can open and close a connection`.

tests/lib/AccountStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('AccountStore', () => {
+    it('is a singleton', () => {
+        const instance1 = AccountStore.instance;
+        const instance2 = AccountStore.instance;
+        expect(instance1).toBe(instance2);
+    });
+
+    it('connection can be opened and closed', async () => {
```
sisou: Naming should follow the "it ..." syntax, for example **it** `can open and close a connection`.

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    it('can get a plain key', async () => {
+        const [key1, key2] = await Promise.all([
+            KeyStore.instance._getPlain(Dummy.keyInfo[0].userFriendlyAddress),
+            KeyStore.instance._getPlain(Dummy.keyInfo[1].userFriendlyAddress)
+        ]);
+        expect(key1).toEqual(Dummy.keyDatabaseEntries[0]);
+        expect(key2).toEqual(Dummy.keyDatabaseEntries[1]);
+    });
+
+    it('can get and decrypt a key', async () => {
```
sisou: Again, you are testing two, so plural would be correct.

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+            KeyStore.instance.get(Dummy.keyInfo[0].userFriendlyAddress, Dummy.encryptionPassword),
+            KeyStore.instance.get(Dummy.keyInfo[1].userFriendlyAddress, Dummy.encryptionPassword)
+        ]);
+        expect(key1.getPublicInfo()).toEqual(Dummy.keyInfo[0]);
+        expect(key2.getPublicInfo()).toEqual(Dummy.keyInfo[1]);
+        expect(Nimiq.BufferUtils.equals(key1.exportPlain(), Dummy.keyPairs[0])).toBe(true);
+        expect(Nimiq.BufferUtils.equals(key2.exportPlain(), Dummy.keyPairs[1])).toBe(true);
+    });
+
+    it('can get the type of a key', async () => {
```
sisou: Plural.

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+        // check that we can't get a removed key by address
+        const removedKeys = await Promise.all([
+            KeyStore.instance._getPlain(Dummy.keyInfo[0].userFriendlyAddress),
+            KeyStore.instance._getPlain(Dummy.keyInfo[1].userFriendlyAddress)
+        ]);
+        expect(removedKeys[0]).toBeUndefined();
+        expect(removedKeys[1]).toBeUndefined();
+
+        // readd the keys to restore the state before the test
```
sisou: `re-add`

sisou: Would not be necessary if the DB setup would be run `beforeEach/afterEach`.

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+        await Promise.all([
+            KeyStore.instance.putPlain(Dummy.keyDatabaseEntries[0]),
+            KeyStore.instance.putPlain(Dummy.keyDatabaseEntries[1])
+        ]);
+    });
+
+    it('can add and update keys', async () => {
+        // first clear database
```
sisou: This test could be combined with the "delete" test, to avoid manually deleting and adding keys.

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+        let currentKeys = await KeyStore.instance.list();
+        expect(currentKeys.length).toBe(0);
+
+        // add a plain key
+        await KeyStore.instance.putPlain(Dummy.keyDatabaseEntries[0]);
+        currentKeys = await KeyStore.instance.list();
+        expect(currentKeys.length).toBe(1);
+
+        // add an unencrypted key
```
sisou: I think you mean an _encrypted_ key.

tests/lib/AccountStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,25 @@
...
+        const db = await AccountStore.instance.connect();
+        expect(AccountStore.instance._dbPromise).toBeTruthy();
```
sisou: Instead of using `AccountStore.instance` everywhere, you could store it in an `accountStore` variable instead.

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+        expect(plainKey1).toEqual(Dummy.keyDatabaseEntries[0]);
+        expect(key2.getPublicInfo()).toEqual(Dummy.keyInfo[1]);
+        expect(Nimiq.BufferUtils.equals(key2.keyPair.serialize(), Dummy.keyPairs[1])).toBe(true);
+
+        // encryption of keys is not deterministic, thus reencrypted key bytes differ from Dummy.encryptedKeyPairs[1]
+        let plainKey2 = await KeyStore.instance._getPlain(Dummy.keyInfo[1].userFriendlyAddress);
+        expect(plainKey2.encryptedKeyPair).not.toEqual(Dummy.encryptedKeyPairs[1]);
+
+        // update the key2 database entry with the original Dummy.keyDatabaseEntries[1]
+        await KeyStore.instance.putPlain(Dummy.keyDatabaseEntries[1]);
```
sisou: Instead of using `KeyStore.instance` everywhere, you could store it in a `keyStore` variable.

src/lib/AccountStoreIndexeddb.js
danimoh: Well, this._dbPromise. ... is not used anywhere, just await this._dbPromise.
However, readded the type annotation though.

I'm initializing it, to give it the explicit type | null

The reason I replaced this._db by this._dbPromise is that this._dbPromise gets set immediately while this._db only got set after the database was opened. Thus, the approach with this._db lead to problems when simultaneously having multiple connect calls (or any of the other calls that call connect) where it would create multiple database connections but only store the last one. The other ones remain open and prevent the database from being able to be deleted.

src/lib/KeyStoreIndexeddb.js
danimoh: Did you try and confirm it?
MDN states that `expires=` takes a date-in-GMTString-format and the example also uses `Thu, 01 Jan 1970 00:00:01 GMT;` to delete a cookie

src/lib/KeyStoreIndexeddb.js
danimoh: The method always returns `true`, so there is no need for a return value

tests/DummyData.spec.js
danimoh: Well, linting is not enabled on the test files.
Should we enable it?

tests/DummyData.spec.js
danimoh: Well, it would be slower. I make sure myself that all tests return the database to the initial state, if they changed something which only applies to add and remove tests

tests/lib/KeyStoreIndexeddb.spec.js
danimoh: The test name combines to "KeyStore connection can be opened and closed".
But your suggestion also works

tests/lib/KeyStoreIndexeddb.spec.js
danimoh: thought about it but prefer testing them separately

tests/lib/KeyStoreIndexeddb.spec.js
danimoh: No, the key gets encrypted when put into the database but is unencrypted before that

tests/lib/AccountStoreIndexeddb.spec.js
danimoh: But using `.instance` better resembles how the AccountStore / KeyStore are used in the keyguard

src/lib/KeyStoreIndexeddb.js
sisou: > Invalid dates, like the value 0, represent a date in the past and mean that the resource is already expired.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires

sisou: `max-age=0` also works.

tests/DummyData.spec.js
sisou: Best practice is to always create and tear-down the environment for each test. That way no test is infuenced by other tests, they are independent.

tests/lib/KeyStoreIndexeddb.spec.js
sisou: Yes, but the encryption is done in the `Key` class, thus what arrives in the `KeyStore` (which is the topic of this test file) is an encrypted key.

src/lib/KeyStoreIndexeddb.js
```diff
@@ -14,20 +14,23 @@ class KeyStore {
...
-        return new Promise((resolve, reject) => {
+        this._dbPromise = new Promise((resolve, reject) => {
             const request = window.indexedDB.open(KeyStore.DB_NAME, KeyStore.DB_VERSION);
 
             request.onsuccess = () => {
-                /** @type {IDBDatabase | null} */
-                this._db = request.result;
-                resolve(this._db);
+                resolve(request.result);
             };
```
svub: Quite some overlap (aka c&p) with account store > extract common code in super class?

src/lib/KeyStoreIndexeddb.js
```diff
@@ -181,10 +186,12 @@ class KeyStore {
...
-    close() {
-        if (!this._db) return;
-        this._db.close();
-        this._db = null;
+    /** @returns {Promise<void>} */
+    async close() {
+        if (!this._dbPromise) return;
+        const db = await this._dbPromise;
+        db.close();
+        this._dbPromise = null;
```
svub: DRY code...

src/lib/KeyStoreIndexeddb.js
svub: Tested in Safari and Firefox? Is it not following the standard to safe a bytes?

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+            KeyStore.instance._getPlain(Dummy.keyInfo[1].userFriendlyAddress)
+        ]);
+        expect(removedKeys[0]).toBeUndefined();
+        expect(removedKeys[1]).toBeUndefined();
+
+        // re-add the keys to restore the state before the test
+        await Promise.all([
+            KeyStore.instance.putPlain(Dummy.keyDatabaseEntries[0]),
+            KeyStore.instance.putPlain(Dummy.keyDatabaseEntries[1])
+        ]);
```
svub: could be removed when re-initializing the DB for each test.

tests/lib/KeyStoreIndexeddb.spec.js
svub: "encrypt and put key in DB" :P

tests/lib/KeyStoreIndexeddb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+        await KeyStore.instance.putPlain(Dummy.keyDatabaseEntries[1]);
+        plainKey2 = await KeyStore.instance._getPlain(Dummy.keyInfo[1].userFriendlyAddress);
+        expect(plainKey2.encryptedKeyPair).toEqual(Dummy.encryptedKeyPairs[1]);
+    });
+
+    it('can migrate accounts from deprecated AccountStore', async () => {
+        spyOnProperty(document, 'cookie', 'set').and.callFake((/** @type {string} */ cookie) => {
+            expect(cookie.startsWith('migrate') || cookie.startsWith('accounts')).toBe(true);
+            expect(cookie.endsWith('expires=Thu, 01 Jan 1970 00:00:01 GMT')).toBe(true);
+        });
```
svub: cool :) :+1:

src/lib/KeyStoreIndexeddb.js
svub: I guess because the account store is deprecated. 
Are there going to be more stores? If no, ok.

src/lib/AccountStoreIndexeddb.js
```diff
@@ -113,25 +114,25 @@ class AccountStore {
         });
     }
 
-    close() {
-        if (!this._db) return;
-        this._db.close();
-        this._db = null;
+    async close() {
+        if (!this._dbPromise) return;
```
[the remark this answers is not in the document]
danimoh: nah, it's already closed then, so why throw an exception

src/lib/KeyStoreIndexeddb.js
danimoh: exactly, it's deprecated and thus I think it's good the way it is

src/lib/KeyStoreIndexeddb.js
danimoh: The cookie is immediately deleted anyways, so I gonna follow the standard

src/lib/KeyStoreIndexeddb.js
danimoh: intended here because AccountStore is deprecated

tests/lib/KeyStore.spec.js
```diff
@@ -0,0 +1,137 @@
...
+        expect(Nimiq.BufferUtils.equals(key2.keyPair.serialize(), Dummy.keyPairs[1])).toBe(true);
+
+        // encryption of keys is not deterministic, thus reencrypted key bytes differ from Dummy.encryptedKeyPairs[1]
+        let plainKey2 = await KeyStore.instance._getPlain(Dummy.keyInfo[1].userFriendlyAddress);
+        expect(plainKey2.encryptedKeyPair).not.toEqual(Dummy.encryptedKeyPairs[1]);
+
+        // update the key2 database entry with the original Dummy.keyDatabaseEntries[1]
+        await KeyStore.instance.putPlain(Dummy.keyDatabaseEntries[1]);
+        plainKey2 = await KeyStore.instance._getPlain(Dummy.keyInfo[1].userFriendlyAddress);
+        expect(plainKey2.encryptedKeyPair).toEqual(Dummy.encryptedKeyPairs[1]);
```
sisou: Is this block really necessary anymore, since the database is recreated anyway and `putPlain` is tested already before in this test?

danimoh: it's a test whether existing keys can be updated

PR #35, opened 2018-07-11
Translation validation tool

sisou:
Scans all files for i18n and collects all dict keys and their default translation (english). Then compares those to the translation file and notifies about missing keys and missmatching english translations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,9 +10,10 @@
     "checkdeps": "node tools/dependencyValidator.js",
     "test-coverage": "karma start karma.conf.coverage.js",
     "coverage": "nyc instrument ./src ./instrumented && npm run test-coverage",
-    "pr": "npm run typecheck && npm run lint && npm run test && npm run checkdeps",
+    "checklangs": "node tools/translationValidator.js",
+    "pr": "npm run typecheck && npm run lint && npm run test && npm run checkdeps && npm run checklangs",
     "watch": "tsc-watch --pretty --onSuccess \"npm run watchonsuccess\"",
-    "watchonsuccess": "npm run lintfix && npm run checkdeps"
+    "watchonsuccess": "npm run lintfix && npm run checkdeps && npm run checklangs"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -8,9 +8,10 @@
     "lint": "eslint src client tools",
     "lintfix": "eslint --fix src client tools",
     "checkdeps": "node tools/dependencyValidator.js",
-    "pr": "npm run typecheck && npm run lint && npm run test && npm run checkdeps",
+    "checklangs": "node tools/translationValidator.js",
+    "pr": "npm run typecheck && npm run lint && npm run test && npm run checkdeps && npm run checklangs",
     "watch": "tsc-watch --pretty --onSuccess \"npm run watchonsuccess\"",
-    "watchonsuccess": "npm run lintfix && npm run checkdeps"
+    "watchonsuccess": "npm run lintfix && npm run checkdeps && npm run checklangs"
```
svub: would prefer to run lintfix when I need it (aka on demand) and have "lint only" being used by default -- because lintfix modifies "my files" on every save and thus gives visual studio code the hick-ups sometimes.

tools/translationValidator.js
```diff
@@ -0,0 +1,101 @@
+const fs = require('fs');
+const { spawnSync } = require('child_process');
+
+// const funcs = require('./functions');
+
```
svub: remove commented out code here and below

package.json
Bettelstab: yep, however this is not what this pr is about ;)

svub:
Ah, and I think the check for of tag text content to dict English translation will be annoying because we need to keep the HTML pages and the dict in sync = not DRY. Plus, all the English in the HTML is extra data to be downloaded as it's downloaded in the HTML and in the dict = twice.
So, having the English sentences twice is so so, but enforcing it is bad

sisou:
@svub in a future update, when the language files are separated by language, the english dict can even be generated by this tool. But I agree, I was not too sure if I should enforce this, so I can put it as a warning only (if english in-DOM and english in-DICT do not match).

PR #36, opened 2018-07-11
Create and tear-down databases beforeEach/afterEach

sisou:


sisou:
Synchonous execution issues.

PR #39, opened 2018-07-12
Create and tear-down databases beforeEach/afterEach

sisou:


sisou:
Fixed in https://github.com/nimiq/keyguard-next/pull/32 last commit.

PR #40, opened 2018-07-12
defer scripts in iframe

sisou:
Because of the new way in that i18n is loaded, it is now auto-executed in `runKeyguard`, but that only works when `document.body` is already filled and not `null`. Thus have to defer all scripts.

### DIFF
--- a/src/request/iframe/index.html
+++ b/src/request/iframe/index.html
@@ -2,23 +2,23 @@
 <html>
     <head>
         <meta charset="UTF-8">
-        <script src="https://cdn.nimiq.com/web-offline.js"></script>
-        <script src="../../lib/BrowserDetection.js"></script>
-        <script src="../../lib/AccountStore.js"></script> <!-- @deprecated Only for database migration -->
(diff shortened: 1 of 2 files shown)


src/request/iframe/index.html
```diff
@@ -2,23 +2,23 @@
...
+        <script defer src="../../lib/KeyStore.js"></script>
+        <script defer src="../../lib/Utf8Tools.js"></script>
+        <script defer src="../../lib/RpcServer.js"></script>
+        <script defer src="../../lib/Key.js"></script>
+        <script defer src="../../lib/AddressUtils.js"></script>
+        <script defer src="../../lib/EncryptionType.js"></script>
+        <script defer src="../../lib/CookieJar.js"></script>
+        <script defer src="../../common.js"></script>
+        <script defer src="../../errors.js"></script>
         <!-- TODO move I18n init to PopupApi and remove this -->
```
svub: using "import"? Or c&p the code in?

PR #47, opened 2018-07-12
possibility to add window target in client

Bettelstab:


client/KeyguardClient.js
```diff
@@ -129,9 +136,11 @@ class KeyguardClient {
      * @param {any[]} [args]
      */
     async _startPopup(requestName, args) {
+        const requestUrl = `${this._keyguardSrc}/request/${requestName}/`;
+
         const $popup = window.open(
-            `${this._keyguardSrc}request/${requestName}/`,
-            'Nimiq Keyguard Popup',
+            requestUrl,
+            this._targetName,
```
danimoh: Mdn says `The name should not contain whitespace.` (https://developer.mozilla.org/en-US/docs/Web/API/Window/open)

svub: yep, I remember it breaks in some browsers if you do

svub: the `name` is the identifier for this "frame" - the window title will be set by the HTML page loaded inside.

src/common.js
```diff
@@ -37,6 +37,4 @@ async function runKeyguard(RequestApiClass, options) { // eslint-disable-line no
```
svub: Hasn't `I18n` been moved to PopupAPI?

sisou: That PR is still pending (#48).

PR #49, opened 2018-07-13
Iframe api tests

danimoh:


tests/lib/IframeApi.spec.js
```diff
@@ -0,0 +1,96 @@
...
+
+    it('can list key info from cookies on iOS', async () => {
+        spyOn(BrowserDetection, 'isIos').and.returnValue(true);
+
+        const listedKeys = await iframeApi.list();
+
+        expect(CookieJar.eat).toHaveBeenCalledWith(undefined);
+        expect(AccountStore.instance.list).not.toHaveBeenCalled();
+        expect(KeyStore.instance.list).not.toHaveBeenCalled();
+        expect(listedKeys).toEqual(Dummy.keyInfo);
```
svub: doesn't this only work when the cookie is set before? What happens when this method is called for the very first time?

tests/lib/IframeApi.spec.js
```diff
@@ -0,0 +1,96 @@
...
+        spyOn(BrowserDetection, 'isIos').and.returnValue(true);
+        spyOnProperty(document, 'cookie', 'set').and.callFake((/** @type {string} */ cookie) => {
+            expect(cookie.startsWith('migrate=1;')).toBe(true);
+            cookieSet = true;
+        });
+
+        await iframeApi.migrateAccountsToKeys();
+
+        expect(cookieSet).toBe(true);
+        expect(KeyStore.instance.doMigrateAccountsToKeys).not.toHaveBeenCalled();
```
svub: the cookie kicks off migration once the popup is opended again. what happens if the other cookie is used in the meantime to list the accounts? it will be old accounts until the upgrade happens, right? if the popup wont be opened, the list in the cookie will never be updated, or? 
If yes, then maybe should open the popup to migrate on iOS devices.

tests/lib/IframeApi.spec.js
sisou: If no cookie present, list() should return an empty array.

tests/lib/IframeApi.spec.js
sisou: You are right, we might have to delete the accounts cookie already when setting the migrate signal.

The thing is: this whole thing depends on the iframe being able to write a cookie, that the popup can then read. If that is not possible on iOS, the  we have to use a popup anyway.

src/request/iframe/IFrameApi.js
```diff
@@ -35,6 +35,7 @@ class IFrameApi { // eslint-disable-line no-unused-vars
         if (BrowserDetection.isIos() || BrowserDetection.isSafari()) {
             // Set migrate flag cookie
             document.cookie = 'migrate=1;max-age=31536000';
+            return;
```
sisou: :+1: I missed this, thanks!

tests/lib/IframeApi.spec.js
danimoh: yes, it is implemented like that. I'll add a test for that in an extra PR

tests/lib/IframeApi.spec.js
danimoh: It's okay to use the old accounts cookie until the accounts get migrated in a popup as the accounts cookie still is up to date as everything that changes the accounts would require a popup anyways

PR #50, opened 2018-07-13
Tests-key-store: differentiate between account migration on iOS and n…

danimoh:
…on-iOS

### DIFF
--- a/tests/lib/KeyStore.spec.js
+++ b/tests/lib/KeyStore.spec.js
@@ -111,20 +111,55 @@ describe('KeyStore', () => {
         expect(plainKey2.encryptedKeyPair).toEqual(Dummy.encryptedKeyPairs[1]);
     });
 
-    it('can migrate accounts from deprecated AccountStore', async () => {
+    it('can migrate accounts from deprecated AccountStore on non-iOS', async () => {
         // clear key store and fill account store
         await Promise.all([
             Dummy.Utils.deleteDummyKeyStore(),
             Dummy.Utils.createDummyAccountStore()
         ]);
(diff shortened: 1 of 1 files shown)


tests/lib/KeyStore.spec.js
```diff
@@ -111,20 +111,55 @@ describe('KeyStore', () => {
...
+        let migrationCookieDeleted = false,
+            accountsCookieDeleted = false;
+        spyOnProperty(document, 'cookie', 'set').and.callFake((/** @type {string} */ cookie) => {
+            if (cookie.startsWith('migrate=')) {
+                migrationCookieDeleted = true;
+            } else if (cookie.startsWith('accounts=')) {
+                accountsCookieDeleted = true;
+            }
+            expect(migrationCookieDeleted || accountsCookieDeleted).toBe(true);
+            expect(cookie.includes('expires=Thu, 01 Jan 1970 00:00:01 GMT')).toBe(true);
```
sisou: I'd put this expect before the if-else condition.

danimoh: What's the difference?

danimoh:
Setting of the migration cookie is tested in the IframeApi specs and for the PopupApi (that triggers doMigrateAccountsToKeys) there are no tests yet.

tests/lib/KeyStore.spec.js
sisou: The difference is only a technicality: that you test first if the `set` is a deletion, before setting the `...deleted` variable.

svub: it's also nice to have all expects together (code style / readability). No strong opinion though

PR #51, opened 2018-07-13
AccountStore: if database doesn't exist list no accounts instead of f…

danimoh:
…ailing

### DIFF
--- a/src/lib/AccountStore.js
+++ b/src/lib/AccountStore.js
@@ -45,9 +45,10 @@ class AccountStore {
 
             request.onerror = () => reject(request.error);
             request.onupgradeneeded = () => {
+                // account database doesn't exist
                 this._dropped = true;
-                request.result.close();
-                reject(new Error('Account database does not exist'));
+                request.transaction.abort();
+                resolve(null);
             };
         });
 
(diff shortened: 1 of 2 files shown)


sisou:
What is the reason for this change?

sisou:
I know what it _does_, but _why_ does it need to be changed?

src/lib/AccountStore.js
```diff
@@ -59,6 +60,7 @@ class AccountStore {
      */
     async list() {
         const db = await this.connect();
+        if (!db) return [];
```
svub: I'd expect it to return `[]` in a promise though right away

src/lib/AccountStore.js
```diff
@@ -96,6 +98,7 @@ class AccountStore {
      */
     async dangerousListPlain() {
         const db = await this.connect();
+        if (!db) return [];
```
svub: same here. I always find it weird/ugly if a method returns a promise sometimes and an object other times. :/

svub: So my suggestion is to move this `if` on line down and go `if (!db) resolve([]);`

sisou: But it's an `async` function, so it returns a promise anyway, no?

danimoh: yes, it returns a promise anyway and thus I'd say `return [];` is easier to read

danimoh:
In my opinion the API is nicer if it doesn't throw if nothing unexpected happened.
E.g. if I call the method to list the accounts but there are none, then it should just return none and not throw.

PR #54, opened 2018-07-14
i18n and privacy agent for create request

svub:
Using PrivacyAgent component
And made "everything" i18n

PR #56, opened 2018-07-17
Small style fixes

danimoh:


src/request/create/CreateLow.js
```diff
@@ -46,17 +46,17 @@ class CreateLow {
         this._pinInput.on(
             PinInput.Events.PIN_ENTERED,
             /** @param {number} pin */ async pin => {
-                $confirmMessage.classList.add('hidden');
-                $notMatchingMessage.classList.add('hidden');
+                $confirmMessage.classList.add('display-none');
```
sisou: I believe the `hidden` class is used here on purpose to not make the content jump when unhiding it.

src/request/create/index.html
```diff
@@ -42,7 +42,7 @@
     <link rel="stylesheet" href="../../components/PassphraseInput.css">
     <link rel="stylesheet" href="../../components/RecoveryWords.css">
     <link rel="stylesheet" href="../../components/ValidateWords.css">
-    <link rel="stylesheet" href="./privacy-agent.css">
+    <link rel="stylesheet" href="../../components/PrivacyAgent.css">
```
sisou: This is already fixed in Sven's PR.

sisou: Or rather the CSS is already added some lines above.

src/request/create/privacy-agent.css
```diff
@@ -1,52 +0,0 @@
-.privacy-agent {
```
sisou: @svub did you not remove this file?

src/request/create/CreateLow.js
danimoh: To keep the no-jumping behaviour the message is absolutely positioned in an always visible container.

I changed the visibility:hidden to display:none two avoid that:
- both messages take space in the layout while at most one is visible at the same time
- the messages are not rendered at the same position
- the bottom message has no margin to the pinpad dots

src/request/create/index.html
danimoh: You mean #54 ?
Apparently removal of this file was forgotten there

Issue #84, opened 2018-09-10
Enable back-buttons in create flow


sisou:
The page-header back-buttons in the create flow should be made functional, including the first back-button, which should take the user back to the AccountManager's wallet-type selection page.

Bettelstab:
Solved in #86, not merged yet

Bettelstab:
First back button needs https://github.com/nimiq/rpc/tree/max/recoverOnHistoryBack to be merged.

sisou:
RPC PR now pending: https://github.com/nimiq/rpc/pull/3

PR #88, opened 2018-09-19
Login flow

sisou:
Support for switching to Word-import or Wallet-creation is not yet implemented.


src/nimiq-style.css
```diff
@@ -124,18 +125,24 @@
...
+}
+
 /***************
```
Bettelstab: icon-keyfile-secure, icon-keyfile-insecure, icon-keyfile-locked werden 2x definiert?!

src/request/import-file/ImportFileApi.js
```diff
@@ -1,136 +1,217 @@
...
         // Events
-        fileImportComponent.on(FileImport.Events.IMPORT, this._onFileImported.bind(this));
-        passphraseInput.on(PassphraseInput.Events.PASSPHRASE_ENTERED, this._onPassphraseEntered.bind(this));
-        pinInput.on(PinInput.Events.PIN_ENTERED, this._onPinEntered.bind(this));
+        fileImport.on(FileImport.Events.IMPORT, this._onFileImported.bind(this));
+        passphraseBox.on(PassphraseBox.Events.SUBMIT, this._onPassphraseEntered.bind(this));
+        passphraseBox.on(PassphraseBox.Events.CANCEL, () => this._goToImportPage());
+        passphraseSetterBox.on(PassphraseSetterBox.Events.SUBMIT, this._onPassphraseEntered.bind(this));
+        passphraseSetterBox.on(PassphraseSetterBox.Events.SKIP, () => this._onPassphraseEntered(null));
+        backArrow.addEventListener('click', () => this._goToImportPage());
```
Bettelstab: Is this a special case or can we remove it after merging #84 ?

sisou: I think it's covered by #84 and must be removed.

src/nimiq-style.css
sisou: Where else is it defined?

Bettelstab: in line 48-51

sisou: Yes, this is how all icons are defined. At the top are general styles for all icons, and then the specific sizes and actual backgrounds are defined individually. I think you may have been confused by how GitHub collapses the changes. Extend the file to see how it works.

Bettelstab: Ok, but weight and height do get overwritten. Anyway

sisou: Yes, because most icons have the same dimensions, so it makes sense to declare it once for all. But some icons are special and have other sizes, so we overwrite it.

You are welcome to suggest a different method.

PR #92, opened 2018-09-26
Add payment-info line to SignTransaction CheckoutLayout

### LINKED ISSUE #83: Payment info-line above page during checkout
As shown in the mockups in Zeplin, a so-called payment info-line should be displayed above the page for the duration of the checkout flow. This is currently only shown in the AccountManager, but no yet in the Keyguard.

sisou:
Resolves #83.

Only problem remaining is how to hide the info-line when the user goes back or cancels the signing. The Keyguard shortly shows the loading animation again before redirecting back to the AccountsManager. In that time, the page is hidden because it is not a CSS target anymore, but the payment-info-line lives outside the page, so is not hidden. And CSS can only be written to affect later elements in the DOM, so the info-line cannot react to a page being hidden (the page comes after the info-line)...

Only possibility I see is instead of relying solely on the CSS target, to also add a `loading` class or similar to the body or `#app` element... @Bettelstab?

### DIFF
--- a/src/components/PaymentInfoLine.css
+++ b/src/components/PaymentInfoLine.css
@@ -1,0 +1,21 @@
+.payment-info-line {
+    width: 100vw;
+    max-width: 56.25rem;
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    color: white;
+    font-size: 1.5rem;
+    line-height: 1;
+    margin-bottom: 1.5rem;
+    text-transform: uppercase;
+    letter-spacing: 0.15rem;
+}
+
+.payment-info-line .amount {
+    font-weight: bold;
+}
+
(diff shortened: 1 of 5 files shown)


Bettelstab:
There are several possibilities how to handle this payment line:
- Copy & Paste HTML ;(
- Make it a component and add it at the start of each page
- Place it after the page and use CSS (order property, #app is flex anyway) to make it appear before the page. (Could be the smartest)
- Or yours :)

sisou:
This place is the only place where it's needed, in the checkout layout.

I thought about doing the flex reverse order as well. I don't like it too much as it is unexpected order, but I currently think it's still the cleanest solution.

sisou:
And I also would agree with making it a component, although it's only needed once. But we'll never know.

I'll refactor it.

PR #93, opened 2018-09-26
Import and use local RpcServer bundle

### LINKED ISSUE #77: Build and use RPC Server only
Make a build of RPC Server (https://github.com/nimiq/rpc) without all the unneeded code (client, postMessage) , so it is easier to review.

sisou:
Resolves #77.

This uses a reduced RpcServer-only build from the @nimiq/rpc package. Additionally, I removed unneeded code as suggested in #77.

### DIFF
--- a/src/lib/RpcServer.es.js
+++ b/src/lib/RpcServer.es.js
@@ -1,0 +1,426 @@
+var ResponseStatus;
+(function (ResponseStatus) {
+    ResponseStatus["OK"] = "ok";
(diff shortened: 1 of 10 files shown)


src/lib/RpcServer.es.js
```diff
@@ -0,0 +1,329 @@
+// @ts-nocheck
+/* eslint-disable */
+
+var ResponseStatus;
```
Bettelstab: Suggestion: Add a comment which says where this file comes from.

PR #115, opened 2018-11-16
Use cdn.nimiq-testnet.com for Nimiq lib

sisou:
And update CSP.

### DIFF
--- a/src/request/create/index.html
+++ b/src/request/create/index.html
@@ -4,13 +4,15 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <!-- TODO: Don't use nimiq-network.com for releases -->
     <meta http-equiv="Content-Security-Policy"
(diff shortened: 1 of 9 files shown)


src/request/create/index.html
```diff
@@ -4,13 +4,15 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <!-- TODO: Don't use nimiq-network.com for releases -->
     <meta http-equiv="Content-Security-Policy"
-          content="default-src 'self' 'unsafe-eval' https://cdn.nimiq.com https://cdn.nimiq-network.com;
-          img-src http: https: blob: data:; child-src 'self' blob:; style-src 'self';">
+          content="default-src 'self' 'unsafe-eval' https://cdn.nimiq-testnet.com;
```
svub: I'm confused. the PR says "using cdn.nimiq-network.com" but `nimiq-testnet.com` is used here and in other changed files below.

svub: actually everywhere...

sisou: Ups, I always confuse those two when writing the url when not actively thinking about it. Already renamed the PR, will fox the commit message later.

PR #117, opened 2018-11-21
Change Passphrase Request

nibhar:


src/request/change-passphrase/ChangePassphrase.js
```diff
@@ -0,0 +1,158 @@
...
+            try {
+                const passphrase = phrase ? Nimiq.BufferUtils.fromAscii(phrase) : undefined;
+                const key = await KeyStore.instance.get(this._request.keyInfo.id, passphrase);
+                if (!key) {
+                    this._reject(new Error('No key'));
+                }
+                this._key = key;
+                /** @type {PassphraseSetterBox} */ (this._setPassphraseBox).reset();
+                window.location.hash = ChangePassphrase.Pages.SET_PASSPHRASE;
+                /** @type {PassphraseSetterBox} */ (this._setPassphraseBox).focus();
```
sisou: If the code from the `_create()` function was to be moved into the `constructor`, then you wouldn't need all the type annotation, because the variables would not be able to be `undefined`.

PR #118, opened 2018-11-21
Unify request types

Bettelstab:
Get rid of duplicate type definitions


client/types/KeyguardRequests.d.ts
```diff
@@ -0,0 +1,125 @@
+declare namespace KeyguardRequests {
```
sisou: What do you think about calling it `KeyguardRequest` (singular), just like `KeyguardCommand`? In the code it would then be more correct when referring to a single request type.

client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@nimiq/keyguard-client",
-  "version": "0.0.20",
-  "description": "Nimiq Keyguard client library",
+  "version": "0.0.21",
+  "description": "Nimiq Keyguard client library dev test",
```
sisou: Is this `dev test` supposed to be here?

types/KeyguardRequests.backup
```diff
@@ -71,7 +71,7 @@ type SignMessageResult = {
     signature: Uint8Array
```
sisou: Since you renamed this file to `.backup`, is it still necessary or can it be removed?

client/types/KeyguardRequests.d.ts
Bettelstab: sounds reasonable :)

client/package.json
Bettelstab: nope ;)

types/KeyguardRequests.backup
Bettelstab: removed!

client/types/KeyguardRequest.d.ts
```diff
@@ -0,0 +1,125 @@
...
+        address: Uint8Array
+    }
+
+    type KeyguardRequest = CreateRequest
+        | ImportRequest
+        | SimpleRequest
+        | SignTransactionRequest
+        | DeriveAddressRequest
+        | SignMessageRequest
+}
```
nibhar: missing end of file line break

PR #121, opened 2018-11-24
Store if a key uses a 6-digit PIN to adjust passphrase inputs

### LINKED ISSUE #103: Can't sign transactions with 6-character low-legacy account
After importing a low-security legacy "miner account", the sign-transaction PassphraseBox does not allow 6-character passphrase.

sisou:
Resolves #103

### DIFF
--- a/src/lib/CookieJar.js
+++ b/src/lib/CookieJar.js
@@ -38,7 +38,9 @@ class CookieJar { // eslint-disable-line no-unused-vars
      * @returns {string}
      */
     static _encodeCookie(keys) {
-        return keys.map(keyInfo => `${keyInfo.type}${keyInfo.encrypted ? 1 : 0}${keyInfo.id}`).join('');
+        return keys.map(
+            keyInfo => `${keyInfo.type}${keyInfo.encrypted ? 1 : 0}${keyInfo.hasPin ? 1 : 0}${keyInfo.id}`,
+        ).join('');
     }
 
     /**
@@ -48,16 +50,17 @@ class CookieJar { // eslint-disable-line no-unused-vars
     static _decodeCookie(str) {
         if (!str) return [];
 
-        if (str.length % 14 !== 0) throw new Error('Malformed cookie');
+        if (str.length % 15 !== 0) throw new Error('Malformed cookie');
 
-        const keys = str.match(/.{14}/g);
+        const keys = str.match(/.{15}/g);
         if (!keys) return []; // Make TS happy (match() can potentially return NULL)
 
         return keys.map(key => {
             const type = /** @type {Key.Type} */ (parseInt(key[0], 10));
             const encrypted = key[1] === '1';
-            const id = key.substr(2);
-            return new KeyInfo(id, type, encrypted);
+            const hasPin = key[2] === '1';
+            const id = key.substr(3);
+            return new KeyInfo(id, type, encrypted, hasPin);
         });
     }
 }
--- a/src/lib/Key.js
+++ b/src/lib/Key.js
@@ -4,10 +4,13 @@ class Key {
     /**
      * @param {Uint8Array} secret
      * @param {Key.Type} [type]
+     * @param {boolean} [hasPin]
      */
-    constructor(secret, type = Key.Type.BIP39) {
+    constructor(secret, type = Key.Type.BIP39, hasPin = false) {
         this._secret = secret;
         this._type = type;
+        /** @type {boolean} */
+        this._hasPin = !!hasPin;
     }
 
     /**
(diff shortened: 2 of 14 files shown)


src/lib/Key.js
```diff
@@ -4,10 +4,13 @@ class Key {
...
      * @param {Uint8Array} secret
      * @param {Key.Type} [type]
+     * @param {boolean} [hasPin]
      */
-    constructor(secret, type = Key.Type.BIP39) {
+    constructor(secret, type = Key.Type.BIP39, hasPin = false) {
         this._secret = secret;
         this._type = type;
+        /** @type {boolean} */
+        this._hasPin = !!hasPin;
```
nibhar: are the double exclamation marks required? Looks to me it is typed correctly so no need to force the boolean?

It does force the boolean as some sort of input validation which generally is never a bad thing, however I think there is no need for it here (also the other inputs aren't validated as well).

Curious what the others think.

src/lib/Key.js
```diff
@@ -62,6 +65,17 @@ class Key {
...
 
+    /**
+     * @type {boolean}
+     */
+    get hasPin() {
+        return this._hasPin;
+    }
+
+    set hasPin(hasPin) {
+        this._hasPin = !!hasPin;
```
nibhar: same as before.

src/lib/KeyStore.js
```diff
@@ -81,7 +81,7 @@ class KeyStore {
     async getInfo(id) {
         /** @type {?KeyRecord} */
         const keyRecord = await this._get(id);
-        return keyRecord ? new KeyInfo(keyRecord.id, keyRecord.type, keyRecord.encrypted) : null;
+        return keyRecord ? new KeyInfo(keyRecord.id, keyRecord.type, keyRecord.encrypted, keyRecord.hasPin) : null;
```
svub:
```suggestion
        return keyRecord ? KeyInfo.fromObject(keyRecord) : null;
```

src/lib/KeyStore.js
```diff
@@ -151,7 +152,12 @@ class KeyStore {
             .openCursor();
 
         const results = /** KeyRecord[] */ await KeyStore._readAllFromCursor(request);
-        return results.map(keyRecord => new KeyInfo(keyRecord.id, keyRecord.type, keyRecord.encrypted));
+        return results.map(keyRecord => new KeyInfo(
+            keyRecord.id,
+            keyRecord.type,
+            keyRecord.encrypted,
+            keyRecord.hasPin,
+        ));
```
svub: could also use `KeyInfo.fromObject(keyRecord)` here

src/lib/KeyStore.js
```diff
@@ -185,6 +191,7 @@ class KeyStore {
                 id: legacyKeyId,
                 type: Key.Type.LEGACY,
                 encrypted: true,
+                hasPin: key.type === 'low',
```
svub: How I get it `key.type` is a `Key.Type` (which is `0` or `1`) so it should never be `'low'` here. Is this legacy code?

src/lib/Key.js
svub: ha, same thought

svub: ha, same thought

src/lib/CookieJar.js
```diff
@@ -38,7 +38,9 @@ class CookieJar { // eslint-disable-line no-unused-vars
      * @returns {string}
      */
     static _encodeCookie(keys) {
-        return keys.map(keyInfo => `${keyInfo.type}${keyInfo.encrypted ? 1 : 0}${keyInfo.id}`).join('');
+        return keys.map(
+            keyInfo => `${keyInfo.type}${keyInfo.encrypted ? 1 : 0}${keyInfo.hasPin ? 1 : 0}${keyInfo.id}`,
```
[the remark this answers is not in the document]
sisou: As discussed on the phone, we will rather try to improve cookie storage capacity in the Accounts Manager, since that needs to transfer much more data in the cookie, thus will be the bottle neck.

Improving this for the sake of readability would be premature optimization.

src/lib/KeyInfo.js
```diff
@@ -6,14 +6,17 @@ class KeyInfo {
...
-    constructor(id, type, encrypted) {
+    constructor(id, type, encrypted, hasPin) {
         /** @private */
         this._id = id;
         /** @private */
         this._type = type;
         /** @private */
         this._encrypted = encrypted;
+        /** @private */
+        this._hasPin = hasPin;
```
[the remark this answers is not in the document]
sisou: I removed `!!` everywhere in this PR.

PR #123, opened 2018-11-25
Implement sign-message request

### LINKED ISSUE #41: Sign Message Request

sisou:
Resolves #41.


src/request/sign-message/SignMessage.js
```diff
@@ -0,0 +1,137 @@
...
+        // in the Accounts Manager and not return directly to the caller.
+        this._passphraseBox.on(PassphraseBox.Events.CANCEL, () => window.history.back());
+
+        /** @type {HTMLElement} */
+        const $appName = (document.querySelector('#app-name'));
+        $appName.textContent = request.appName;
+        /** @type {HTMLButtonElement} */
+        const $cancelLink = ($appName.parentNode);
+        $cancelLink.classList.remove('display-none');
+        $cancelLink.addEventListener('click', () => reject(new Error('CANCEL')));
```
nibhar: I suggest moving the lines concerning the cancel link into the api's `onRequest`. It is not for this specific page but for this specific request, and in case the handler gets re-used alongside another this code would run multiple times.

However we have multiple styles of the structure anyways. So this is really about how we want to do this in general. Maybe have a discussion about it with the rest.

src/translations/index.js
```diff
@@ -189,6 +194,10 @@ const TRANSLATIONS = {
         'sign-tx-to': 'an',
         'sign-tx-pay-with': 'Zahle mit',
 
+        'sign-msg-heading': 'Nachricht signieren',
+        'sign-msg-text': 'Bitte gib deine Passphrase ein, um die folgende Nachricht zu signieren:',
```
nibhar: We should have either 'Passphrase' or 'Passwort' here at some point.

PR #126, opened 2018-11-28
Return to Accounts to switch from Import to Create

nibhar:
Rejects with an Error specifically treated to redirect the user from Login to Create

### DIFF
--- a/src/request/import/ImportApi.js
+++ b/src/request/import/ImportApi.js
@@ -75,7 +75,7 @@ class ImportApi extends TopLevelApi {
             );
             handler.run();
         });
-        $createWalletLink.addEventListener('click', () => alert('Sorry, not implemented yet.'));
+        $createWalletLink.addEventListener('click', () => this.reject(new Error('GO_CREATE')));
 
         return {
             passphraseBox,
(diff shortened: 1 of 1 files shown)


src/request/import/ImportApi.js
```diff
@@ -75,7 +75,7 @@ class ImportApi extends TopLevelApi {
             );
             handler.run();
         });
-        $createWalletLink.addEventListener('click', () => alert('Sorry, not implemented yet.'));
+        $createWalletLink.addEventListener('click', () => this.reject(new Error('GO_CREATE')));
```
sisou: `GOTO_CREATE`

src/request/import/ImportApi.js
```diff
@@ -75,7 +75,7 @@ class ImportApi extends TopLevelApi {
             );
             handler.run();
         });
-        $createWalletLink.addEventListener('click', () => alert('Sorry, not implemented yet.'));
+        $createWalletLink.addEventListener('click', () => this.reject(new Error('GOTO_CREATE')));
```
Bettelstab: GOTO_CREATE seems like it should be a constant. But I guess it's only used here and in accounts?

sisou: Sebastian is working on making all errors constants currently in his `sebastian/errors` branch.

PR #139, opened 2018-12-04
Extract visual loading state setters into common methods

### LINKED ISSUE #133: Create function for setting app loading state
Move `document.body.classList.remove('loading');` and `document.body.classList.add('loading');` into (static) functions on the TopLevelApi.

sisou:
Resolves #133.

### DIFF
--- a/src/request/TopLevelApi.js
+++ b/src/request/TopLevelApi.js
@@ -118,12 +118,11 @@ class TopLevelApi { // eslint-disable-line no-unused-vars
         this._reject(error);
     }
 
-    static setLoading() {
-        document.body.classList.add('loading');
-    }
-
-    static removeLoading() {
-        document.body.classList.remove('loading');
+    /**
+     * @param {boolean} showLoading
+     */
+    static setLoading(showLoading) {
+        document.body.classList.toggle('loading', showLoading);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/request/TopLevelApi.js
```diff
@@ -118,9 +118,18 @@ class TopLevelApi { // eslint-disable-line no-unused-vars
...
     }
 
+    static setLoading() {
+        document.body.classList.add('loading');
+    }
+
+    static removeLoading() {
+        document.body.classList.remove('loading');
+    }
+
```
nibhar: As a suggestion for both in one function.
```javascript
static loading(loading=true) {
  document.body.classList.toggle('loading',loading);
}
```

Bettelstab: +1

Bettelstab: I would keep setLoading as name, though.

sisou: I disagree and actually made it two functions on purpose, to make it more explicit and not have to pass a boolean for removing it.

nibhar: I personally am in favour of one function for that kind of functionality. More so as `classList.toggle` exists.

PR #148, opened 2018-12-09
Parse incoming requests

nibhar:
Adds RequestParser from which ToplevelApi inherits. RequestParser provides parsing methods for the currently used properties of requests.
The Api classes of the specific requests now use those parse methods to parse the incoming request.

Also adds a test suit for the RequestParser.

### DIFF
--- a/src/lib/RequestParser.js
+++ b/src/lib/RequestParser.js
@@ -17,7 +17,7 @@ class RequestParser { // eslint-disable-line no-unused-vars
 
     /**
      * @param {any} path
-     * @param {string} name
+     * @param {string} name -  name of the property, used in error case only
      * @returns {string}
      */
     parsePath(path, name) {
@@ -36,7 +36,7 @@ class RequestParser { // eslint-disable-line no-unused-vars
 
     /**
      * @param {any} paths
-     * @param {string} name
+     * @param {string} name - name of the property, used in error case only
      * @returns {string[]}
      */
     parsePathsArray(paths, name) {
@@ -70,7 +70,7 @@ class RequestParser { // eslint-disable-line no-unused-vars
         if (typeof label !== 'string') {
             throw new Errors.InvalidRequestError('Label must be of type string');
         }
-        if(label.length === 0) {
+        if (label.length === 0) {
             return undefined;
         }
         return label;
@@ -84,10 +84,10 @@ class RequestParser { // eslint-disable-line no-unused-vars
         if (!keyId) {
             throw new Errors.InvalidRequestError('keyId is required');
         }
-        if(typeof keyId !== 'string') {
+        if (typeof keyId !== 'string') {
             throw new Errors.InvalidRequestError('keyId must be of type string');
         }
-        const keyInfo= await KeyStore.instance.getInfo(keyId);
+        const keyInfo = await KeyStore.instance.getInfo(keyId);
         if (!keyInfo) {
             throw new Errors.KeyNotFoundError();
         }
@@ -103,7 +103,7 @@ class RequestParser { // eslint-disable-line no-unused-vars
         if (!indicesArray || indicesArray.constructor !== Array) {
             throw new Errors.InvalidRequestError('indicesToDerive is required');
         }
-        if(indicesArray.length === 0) {
+        if (indicesArray.length === 0) {
             throw new Errors.InvalidRequestError('Indice array must not be empty');
         }
         indicesArray.forEach((/** @type {any} */index) => { // eslint-disable-line arrow-parens
@@ -152,8 +152,8 @@ class RequestParser { // eslint-disable-line no-unused-vars
             throw new Errors.InvalidRequestError('Sender and recipient must not match');
         }
 
-        const flags = request.flags || Nimiq.Transaction.Flag.NONE; // TODO verify
-        const data = request.data || new Uint8Array(0); // TODO verify
+        const flags = request.flags || Nimiq.Transaction.Flag.NONE; // TODO verify?
+        const data = request.data || new Uint8Array(0); // TODO verify?
 
         const networkId = request.networkId || Nimiq.GenesisConfig.NETWORK_ID;
         if (networkId !== Nimiq.GenesisConfig.NETWORK_ID) {
--- a/src/lib/RequestParser.js
+++ b/src/lib/RequestParser.js
@@ -1,6 +1,5 @@
 /* global Nimiq */
 /* global KeyStore */
-/* global Key */
 /* global Errors */
 /* global Utf8Tools */
 
@@ -189,6 +188,4 @@ class RequestParser { // eslint-disable-line no-unused-vars
         }
         return parsedUrl.origin;
     }
-
-
 }
--- a/src/lib/RequestParser.js
+++ b/src/lib/RequestParser.js
@@ -1,5 +1,6 @@
 /* global Nimiq */
 /* global KeyStore */
+/* global Key */
 /* global Errors */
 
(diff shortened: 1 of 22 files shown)


tests/lib/RequestParser.spec.js
```diff
@@ -0,0 +1,221 @@
...
+        }
+        expect(error).toEqual(new Errors.InvalidRequestError('keyId is required'));
+
+        keyId = {keyId: 75};
+        try {
+            const x = await requestParser.parseKeyId(keyId);
+        } catch(e) {
+            error = e;
+        }
+        expect(error).toEqual(new Errors.InvalidRequestError('keyId must be of type string'));
```
sisou: I thought you didn't want to test for specific error messages?

tests/lib/RequestParser.spec.js
```diff
@@ -0,0 +1,221 @@
...
+
+
+        keyId = '2ec615522906'; // dummy data
+        let parsedKeyInfo = await requestParser.parseKeyId(keyId);
+        expect(new KeyInfo(
+            '2ec615522906',
+            Key.Type.LEGACY,
+            true,
+            false,
+        )).toEqual(parsedKeyInfo);
```
sisou: The expect should be the other way around here:

expect _test result_ to be _known data_.

tests/lib/RequestParser.spec.js
```diff
@@ -0,0 +1,221 @@
...
+    });
+
+    it('can parse indiceArrays', () => {
```
sisou:
```suggestion
    it('can parse indicesArrays', () => {
```

tests/lib/RequestParser.spec.js
```diff
@@ -0,0 +1,221 @@
...
+        expect( () => { requestParser.parseIndicesArray(indicesArray); }).toThrow();
+        indicesArray = [1,2,3,4,5];
+        expect( () => { requestParser.parseIndicesArray(indicesArray); }).toThrow();
+        indicesArray = ["1","2","3"];
+        expect( () => { requestParser.parseIndicesArray(indicesArray); }).toThrow();
+        indicesArray = ["1'","2'","3'"];
+        expect(requestParser.parseIndicesArray(indicesArray)).toEqual(indicesArray);
+    });
+
+    it('can parse transaction', () => {
```
sisou:
```suggestion
    it('can parse transactions', () => {
```

tests/lib/RequestParser.spec.js
```diff
@@ -0,0 +1,221 @@
...
+            recipient: new Uint8Array([225, 253, 0, 255, 238, 105, 158, 173, 122, 16, 27, 203, 31, 16, 3, 178, 231, 105, 81, 188]),
+            sender: new Uint8Array([238, 61, 13, 183, 158, 200, 247, 106, 130, 61, 9, 123, 134, 82, 60, 95, 16, 71, 39, 70]),
+            senderType: 0,
+            validityStartHeight: 176450,
+            value: 545000000,
+        };
+        expect(requestParser.parseTransaction(transaction)).toEqual(
+            new Nimiq.ExtendedTransaction(
+                new Nimiq.Address(new Uint8Array([238, 61, 13, 183, 158, 200, 247, 106, 130, 61, 9, 123, 134, 82, 60, 95, 16, 71, 39, 70])), //sender
+                0, // senderType
```
sisou: Maybe use the Nimiq type directly? Otherwise we would kind of be testing the Nimiq lib, too. What do you think?
```suggestion
                Nimiq.Account.BASIC, // senderType
```

src/lib/RequestParser.js
```diff
@@ -0,0 +1,202 @@
+/* global Nimiq */
+/* global KeyStore */
+/* global Errors */
+/* global Utf8Tools */
+
+class RequestParser { // eslint-disable-line no-unused-vars
+    /**
+     * @param {any} appName
```
Bettelstab: one could improve type safety here by using unknown instead of any

src/request/change-passphrase/ChangePassphraseApi.js
```diff
@@ -27,40 +26,20 @@ class ChangePassphraseApi extends TopLevelApi { // eslint-disable-line no-unused
     }
 
     /**
+     *
      * @param {KeyguardRequest.SimpleRequest} request
```
Bettelstab: As we don't really know what comes in here, it would be more correct to use unknown as well.

src/lib/RequestParser.js
```diff
@@ -0,0 +1,202 @@
...
+            }
+        });
+        return indicesArray;
+    }
+
+    /**
+     * @param {any} request
+     * @returns {Nimiq.ExtendedTransaction}
+     */
+    parseTransaction(request) {
```
Bettelstab: request => transaction

src/request/change-passphrase/ChangePassphraseApi.js
```diff
@@ -27,40 +26,20 @@ class ChangePassphraseApi extends TopLevelApi { // eslint-disable-line no-unused
...
+        parsedRequest.appName = this.parseAppName(request.appName);
+        parsedRequest.keyInfo = await this.parseKeyId(request.keyId);
+        parsedRequest.keyLabel = this.parseLabel(request.keyLabel);
 
-        return /** @type {ParsedSimpleRequest} */ {
-            appName: request.appName,
-            keyInfo,
-            keyLabel: request.keyLabel,
-        };
+        return parsedRequest;
```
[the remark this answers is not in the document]
nibhar: It would be possible. It would also prevent you from doing additional checks as can be seen in DeriveAddressApi.js line 38.
I went with doing it in the same way for each of the requests.

src/lib/RequestParser.js
nibhar: I renamed it to object for now. Currently it is given the whole request in order to not have to build an additional object. In the future it might change, but transaction I feel is to restrictive.

src/lib/RequestParser.js
nibhar: tsc is not happy with unknown. I'll leave it as any for now. All tests except parseTransaction should work with unknown, too

src/lib/RequestParser.js
```diff
@@ -177,9 +165,12 @@ class RequestParser { // eslint-disable-line no-unused-vars
...
     parseMessage(message) {
-        if (message instanceof Uint8Array) return message;
-        if (typeof message === 'string') return Utf8Tools.stringToUtf8ByteArray(message);
-        throw new Errors.InvalidRequestError('Type of message must be a String or Uint8Array');
+        if (typeof message === 'string') message = Utf8Tools.stringToUtf8ByteArray(message);
+        if (!(message instanceof Uint8Array)) {
+            throw new Errors.InvalidRequestError('message must be a String or Uint8Array');
+        }
+        if (message.length < 255) return message;
+        throw new Errors.InvalidRequestError('message must not exceed 255 characters');
```
sisou: bytes, not characters.

src/lib/RequestParser.js
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * @param {any} address
+     * @returns {Nimiq.Address}
+     */
+    parseAddress(address) {
+        try {
+            const nqAddress = new Nimiq.Address(address);
+            return nqAddress;
+        } catch (error) {
+            throw new Errors.InvalidRequestError(`recipient must be a valid Nimiq Address (${error.message})`);
```
sisou: You need to pass the parameter name into the function, to be able to differenciate between sender and recipient in the error message.

PR #155, opened 2018-12-16
Upgrade encryption to V3 (ImageWallet standard)

sisou:
This PR upgrades the Nimiq core lib version to include a new V3 encryption standard.

The Nimiq core introduces a new parent class for `Entropy` and `PrivateKey`, called `Secret`. In this PR, the Keyguard's `Key` gets initialized with an instance of `Secret`, either an `Entropy` for BIP39 wallets or a `PrivateKey` for legacy single-address wallets. By testing `{Entropy|PrivateKey}.type`, the `Key` is able to determine which algorithms to use to derive address(es).

**The `Key.type` (formerly `{0|1}`) changes in this PR to represent `Nimiq.Secret.Type`: `{1|2}`, making this PR backwards-incompatible. All stored BIP39 wallets used for testing are now useless.**

It is furthermore not stored anymore as a separate property (in IndexedDB), if a key is stored encrypted, as this property is obvious from the stored secret`.bytelength`. It is still required to be passed into a `KeyInfo` instantiation, as the UI needs to be adapted depending on if a password is required. The `KeyStore` therefore gained a new static method `isEncrypted(keyRecord: KeyRecord)`.

Additionally, a key's `encrypted` attribute is removed from the `KeyInfoObject` to not be passed to outside the Keyguard anymore.


tests/lib/Key.spec.js
```diff
@@ -32,16 +33,16 @@ describe('Key', () => {
...
+        const key = new Key(Dummy.secrets[0]);
         const address = 'NQ71 CT4K 7R9R EHSB 7HY9 TSTP XNRQ L2RK 8U4U';
         expect(key.deriveAddress('m').toUserFriendlyAddress()).toEqual(address);
         expect(key.deriveAddress('m/0\'').toUserFriendlyAddress()).toEqual(address);
     });
 
     it('can derive addresses (BIP39)', () => {
-        const key = new Key(Dummy.keys[0], Key.Type.BIP39);
-        const address1 = 'NQ55 07AD F3U4 9FTD XP5N 1UMS 2FVT 3N69 Q0H9';
-        const address2 = 'NQ15 YMBC GHHD S4DY CPR5 LC74 HAKD ES3R 5B7P';
```
sisou: I had to update the addresses here, because we were using the `Dummy.keys[0]` for derivation here, which is supposed to be a legacy single-address wallet. Now, this doesn't work anymore because `Key` detects the type of secret and would derive the same legacy address for both paths, as in the test above.

package.json
```diff
@@ -32,7 +32,7 @@
   },
   "homepage": "https://github.com/nimiq/keyguard-next#readme",
   "devDependencies": {
-    "@nimiq/core-types": "^1.0.0",
+    "@nimiq/core": "nimiq-network/core",
```
sisou: The new core version that includes types is not yet released, but available in `master`, thus I'm using the Github path here, and not the NPM version.

src/common.js
```diff
@@ -13,7 +13,7 @@ async function loadNimiq() {
     // eslint-disable-next-line no-return-assign
     return __nimiqLoaded || (__nimiqLoaded = new Promise(async resolve => {
         // Load web assembly encryption library into browser (if supported)
-        await Nimiq.WasmHelper.doImportBrowser();
+        await Nimiq.WasmHelper.doImport();
```
sisou: This change is due to an API change in @nimiq/core.

package.json
sisou: As soon as the new version including the V3 encryption is released as stable, I will update all CDN links and the package version accordingly. This PR will not be merged before that.

src/lib/KeyStore.js
```diff
@@ -102,26 +116,39 @@ class KeyStore {
...
+            secret = await key.secret.exportEncrypted(passphrase);
+        } else {
+            const buf = new Nimiq.SerialBuffer(KeyStore.UNENCRYPTED_SECRET_SIZE);
+            const purposeId = key.secret.type === Nimiq.Secret.Type.ENTROPY
+                ? Nimiq.Entropy.PURPOSE_ID
+                : Nimiq.PrivateKey.PURPOSE_ID;
+
+            buf.writeUint32(purposeId);
+            key.secret.serialize(buf);
+            secret = buf.subarray(0, buf.byteLength);
```
nibhar: Seems odd, that there is no other way to get the `Uint8Array`.

src/request/import/ImportApi.js
```diff
@@ -189,30 +190,34 @@ class ImportApi extends TopLevelApi {
...
             let encryptionKey = null;
 
             if (passphrase !== null) {
                 encryptionKey = Utf8Tools.stringToUtf8ByteArray(passphrase);
             }
 
-            if (this._encryptedKey.length === Nimiq.CryptoUtils.ENCRYPTION_SIZE) {
-                // Make sure read position is at 0 even after a wrong passphrase
+            // Files: 56 = V3, 54 = V2/V1
+            if (this._encryptedKey.byteLength === 56 || this._encryptedKey.byteLength === 54) {
```
nibhar: I suggest defining constants for these values.

tests/DummyData.spec.js
```diff
@@ -8,21 +8,29 @@ beforeAll(async () => {
...
 
+/** @type {{[key: string]: any}} */
 const Dummy = {};
 
+/* To generate HEX arrays, you can use [...].map(n => '0x' + n.toString(16)) from a decimal array, and then remove the quotes. */
+
 Dummy.keys = [
     // eslint-disable-next-line max-len
-    Uint8Array.from([0x68, 0x2d, 0xf0, 0x35, 0xc3, 0x0e, 0x8c, 0x55, 0xbd, 0xa4, 0xab, 0x51, 0x72, 0x13, 0x79, 0x2c, 0x2a, 0xf4, 0xb4, 0xe9, 0xc2, 0x3a, 0xc2, 0xcf, 0x11, 0xbd, 0x28, 0xb2, 0x37, 0xa6, 0x54, 0x4e]),
+    Uint8Array.from([ 0x68, 0x2d, 0xf0, 0x35, 0xc3, 0x0e, 0x8c, 0x55, 0xbd, 0xa4, 0xab, 0x51, 0x72, 0x13, 0x79, 0x2c, 0x2a, 0xf4, 0xb4, 0xe9, 0xc2, 0x3a, 0xc2, 0xcf, 0x11, 0xbd, 0x28, 0xb2, 0x37, 0xa6, 0x54, 0x4e ]),
```
nibhar: Why are you adding the extra spaces? Seems like an unnecessary change to me and I would actually prefer it the way it was before.

demos/ExportWords.html
```diff
@@ -6,7 +6,7 @@
 </head>
 <body>
 <link href="../src/nimiq-style.css" rel="stylesheet">
-<script src="https://cdn.nimiq-network.com/branches/key-derivation/web-offline.js"></script>
+<script src="https://cdn.nimiq-network.com/branches/master/web-offline.js"></script>
```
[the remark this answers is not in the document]
nibhar: there are more of these in `ExportWords.html`.

src/lib/CookieJar.js
```diff
@@ -50,17 +52,19 @@ class CookieJar { // eslint-disable-line no-unused-vars
     static _decodeCookie(str) {
         if (!str) return [];
 
-        if (str.length % 15 !== 0) throw new Error('Malformed cookie');
+        if (str.length % 14 !== 0) throw new Error('Malformed cookie');
 
-        const keys = str.match(/.{15}/g);
+        const keys = str.match(/.{14}/g);
```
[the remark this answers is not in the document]
sisou: I think if the cookie is empty, because all keys where logged-out of, then it would maybe return NULL? I think the check for NULL is actually wanted now, and the comment about TS can be removed.

src/lib/KeyStore.js
```diff
@@ -102,26 +116,39 @@ class KeyStore {
...
+        return this.putPlain(keyRecord);
     }
 
     /**
      * @param {KeyRecord} keyRecord
      * @returns {Promise<void>}
      */
-    async _put(keyRecord) {
+    async putPlain(keyRecord) {
+        if (keyRecord.secret.byteLength !== 56 && keyRecord.secret.byteLength !== KeyStore.UNENCRYPTED_SECRET_SIZE) {
```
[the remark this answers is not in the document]
sisou: Yes, good point. You are of course correct, and it should be a constant as well.

src/lib/KeyStore.js
sisou: Well, `SeiralBuffer` _extends_ `Uint8Array`, so maybe we can even pass in the raw SerialBuffer, but not sure if that works.

tests/DummyData.spec.js
sisou: Haha, that's actually for readability in my editor, where I have automatic line-breaking and with the extra space, the values align better:
image

demos/ExportWords.html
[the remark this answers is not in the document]
sisou: We don't have the type declared anywhere in the Keyguard...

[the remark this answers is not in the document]
nibhar: You are right. I compared randomly with `demos/PassphraseInput.html` where the type is declared.

src/lib/KeyStore.js
sisou: But `secret` is typed to be a Uint8Array, and storing a `Nimiq.SerialBuffer` into the IndexedDB might produce unknown results, it's better to convert to Uint8Array, and this is the only way to do that.

nibhar: Yes. I looked it up myself, too. Was more about me being surprised.

src/lib/KeyStore.js
```diff
@@ -102,26 +116,39 @@ class KeyStore {
...
+            throw new Errors.KeyguardError('Invalid length: KeyRecord.secret');
```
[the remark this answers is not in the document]
sisou: I like the error type first better. @danimoh, @Bettelstab?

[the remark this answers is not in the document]
Bettelstab: ```KeyRecord.secret has invalid length.```

src/lib/KeyStore.js
```diff
@@ -20,6 +21,17 @@ class KeyStore {
...
 
+    /**
+     * @param {KeyRecord} keyRecord
+     * @returns {boolean}
+     */
+    static isEncrypted(keyRecord) {
+        // Because we are supporting legacy secrets which cannot be converted during migration,
+        // a KeyRecord can be both V2 and V3 encrypted.
+        return keyRecord.secret.byteLength === KeyStore.ENCRYPTED_SECRET_SIZE
+            || keyRecord.secret.byteLength === KeyStore.ENCRYPTED_SECRET_SIZE_V2;
```
Bettelstab: Which is V2 (wasn't that one never used live) and what about V3?

Bettelstab: In any case it should be documented :)

sisou: V3 is the new one, V2 is currently used, and V1 was only the default during betanet.

src/request/import/ImportApi.js
```diff
@@ -225,14 +231,13 @@ class ImportApi extends TopLevelApi {
     }
 
     /**
-     * @param { Nimiq.SerialBuffer } entropy
-     * @param { Key.Type } keyType
+     * @param {Nimiq.Entropy|Nimiq.PrivateKey} entropyOrPrivKey
      */
-    _onRecoveryWordsComplete(entropy, keyType) {
+    _onRecoveryWordsComplete(entropyOrPrivKey) {
```
Bettelstab: change name to secret?

src/request/import/ImportWords.js
```diff
@@ -147,8 +144,14 @@ class ImportWords {
      * @private
      */
     _onKeyTypeChosen(keyType, entropy) {
-        this._key = new Key(entropy.serialize(), keyType);
-        this._resolve(entropy.serialize(), keyType);
+        /** @type {Nimiq.Entropy|Nimiq.PrivateKey} */
+        let entropyOrPrivKey;
```
Bettelstab: change name to secret?

src/request/import/ImportWords.js
```diff
@@ -147,8 +144,14 @@ class ImportWords {
...
+        if (keyType === Nimiq.MnemonicUtils.MnemonicType.LEGACY) {
+            entropyOrPrivKey = entropy;
```
Bettelstab: Shouldn't it be the other way around?

src/request/remove-key/index.html
```diff
@@ -5,14 +5,14 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <meta http-equiv="Content-Security-Policy"
-          content="default-src 'self' 'unsafe-eval' https://cdn.nimiq-testnet.com;
+          content="default-src 'self' 'unsafe-eval' https://cdn.nimiq-network.com;
```
Bettelstab: Why those changes?

sisou: Because this branch has to use the master branch if the core repo, which is not yet released on the testnet CDN. Only on the CI-enabled nimiq-network CDN.

Bettelstab: Ok. Then you should change the string to be replaced in `/tools/build.sh` as well, I guess.

src/lib/KeyStore.js
```diff
@@ -62,16 +74,21 @@ class KeyStore {
             return null;
         }
 
-        if (!keyRecord.encrypted) {
-            return new Key(keyRecord.secret, keyRecord.type);
+        if (!KeyStore.isEncrypted(keyRecord)) {
+            const purposeId = new Nimiq.SerialBuffer(keyRecord.secret).readUint32();
+            const secret = purposeId === Nimiq.Entropy.PURPOSE_ID
```
nibhar: I am not sure about it, but it feels to me like there should be something like
```suggestion
                : (purposeId === Nimiq.PrivateKey.PURPOSE_ID 
                    ? new Nimiq.PrivateKey(keyRecord.secret.subarray(4))
                    : null; // I don't know
```

src/lib/KeyStore.js
```diff
@@ -102,26 +119,40 @@ class KeyStore {
...
-        const secret = !passphrase
-            ? key.secret
-            : await Nimiq.CryptoUtils.encryptOtpKdf(new Nimiq.SerialBuffer(key.secret), passphrase);
+        /** @type {Uint8Array} */
+        let secret;
+        if (passphrase) {
+            secret = await key.secret.exportEncrypted(passphrase);
+        } else {
+            const buf = new Nimiq.SerialBuffer(KeyStore.UNENCRYPTED_SECRET_SIZE);
+            const purposeId = key.secret.type === Nimiq.Secret.Type.ENTROPY
```
nibhar: Same as above, but here we are actually using `key.secret.type`.

src/lib/KeyStore.js
Bettelstab: Nope, because private keys are always encrypted.

src/lib/KeyStore.js
```diff
@@ -62,16 +74,21 @@ class KeyStore {
             return null;
         }
 
-        if (!keyRecord.encrypted) {
-            return new Key(keyRecord.secret, keyRecord.type);
+        if (!KeyStore.isEncrypted(keyRecord)) {
+            const purposeId = new Nimiq.SerialBuffer(keyRecord.secret).readUint32();
+            const secret = purposeId === Nimiq.Entropy.PURPOSE_ID
+                ? new Nimiq.Entropy(keyRecord.secret.subarray(4))
+                : new Nimiq.PrivateKey(keyRecord.secret.subarray(4));
```
Bettelstab: ...wait, am I wrong? :D

src/lib/KeyStore.js
sisou: I made it now use `keyRecord.type` instead of the purposeID, and also check for PrivateKey and create an Entropy in all other cases (as suggested here).

src/lib/KeyStore.js
sisou: Switched it around and added a comment about why the purposeID is stored.

sisou:
I now made sure that the purposeID is only written and not used for comparisons anymore in our code. Only `key.type` is used now, and `instanceof` whereever possible.

src/lib/KeyStore.js
```diff
@@ -62,16 +74,23 @@ class KeyStore {
             return null;
         }
 
-        if (!keyRecord.encrypted) {
-            return new Key(keyRecord.secret, keyRecord.type);
+        if (!KeyStore.isEncrypted(keyRecord)) {
+            // TODO: Compare stored type with purposeID to make sure?
```
nibhar: If someone was to use an Imagewallet from a different source, hence different `PURPOSE_ID`, what would be the expected behaviour of our code? Treat it as Entropy?

sisou: Yes, just like in the core, we would treat it as entropy. That's why I reversed the checks, to handle PrivateKey specially, but in all other cases just make an Entropy from it.

src/request/remove-key/index.html
Bettelstab: ...or do you want to change it back after release of core?

sisou: Yes, I'm changing it back to nimiq-testnet I think after core release.

PR #160, opened 2019-01-07
Configuration via environment variables

Bettelstab:
Makes the build process read and use environment variables ALLOWED_ORIGIN and CDN.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ You can configure the following values by either environment variables or config
 - KEYGUARD_ALLOWED_ORIGIN: The origin from which requests are accepted. '*' allows all origins. Be aware that slashes have to be masked by `\`. Defaults to 'https://accounts.nimiq-testnet.com'.
 - KEYGUARD_CDN: The CDN (content delivery network) from which the core library is served. Defaults to 'https://cdn.nimiq-testnet.com'.
 
-The most straight forward way is to use a configuration file, which has to be placed in `config` folder, and pass their name as argument to the build script. `yarn build local` uses `local.conf`. Some sample files are provided.
+The best way is to use a configuration file, which has to be placed in the `config` folder, and pass its name as an argument to the build script. `yarn build local` uses `local.conf`. Some sample files are provided.
 
 You can also set those values in your server configuration via environment variables. Please refer to your server's configuration, e.g. [https://httpd.apache.org/docs/2.4/env.html] for Apache.
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ yarn
 
 Then you can:
 
-- run the build script with `yarn build`.
+- run the build script with `yarn build [config]`.
 - run the tests with `yarn test`.
 - run the typechecker with `yarn typecheck`.
 - run the typecheck file watcher with `yarn watch`.
@@ -46,21 +46,25 @@ Then you can:
     - lib
     - components
     - request
+- types
 - tests
+- config
+- tools
 - demos
+- client
+    - src
+    - types
 ```
 
 ## Configuration
 
-You can configure the following values via environment variables, which are replaced by the build script.
+You can configure the following values by either environment variables or configuration files:
 - KEYGUARD_ALLOWED_ORIGIN: The origin from which requests are accepted. For Team Nimiq's Keyguard, this would be 'http://accounts.nimiq.com'. Be aware that slashes have to be masked by `\`. Defaults to `'*'`, which allows all origins.
 - KEYGUARD_CDN: The CDN (content delivery network) from which the core library is served. Defaults to 'https://cdn.nimiq-testnet.com'.
 
-For local development, you can set temporary environment variables (valid in your current terminal session) by e.g.
-```export KEYGUARD_ALLOWED_ORIGIN='https:\/\my-custom-accounts-instance.com'```
-on Unix-based systems or
-```set KEYGUARD_ALLOWED_ORIGIN='https:\/\my-custom-accounts-instance.com'```
-on Windows. For a persisted configuration on your webserver, please refer to the server's documentation, e.g. [https://httpd.apache.org/docs/2.4/env.html] for Apache.
+The most straight forward way is to use a configuration file, which has to be placed in `config` folder, and pass their name as argument to the build script. `yarn build local` uses `local.conf`. Some sample files are provided.
+
+You can also set those values in your server configuration via environment variables. Please refer to your server's configuration, e.g. [https://httpd.apache.org/docs/2.4/env.html] for Apache.
 
 ## I18n usage
 
--- a/README.md
+++ b/README.md
@@ -53,13 +53,13 @@ Then you can:
 ## Configuration
 
 You can configure the following values via environment variables, which are replaced by the build script.
-- CONFIG_ALLOWED_ORIGIN: The origin from which requests are accepted. For Team Nimiq's Keyguard, this would be 'http://accounts.nimiq.com'. Be aware that slashes have to be masked by `\`. Defaults to `'*'`, which allows all origins.
-- CONFIG_CDN: The CDN (content delivery network) from which the core library is served. Defaults to 'https://cdn.nimiq-testnet.com'.
+- KEYGUARD_ALLOWED_ORIGIN: The origin from which requests are accepted. For Team Nimiq's Keyguard, this would be 'http://accounts.nimiq.com'. Be aware that slashes have to be masked by `\`. Defaults to `'*'`, which allows all origins.
+- KEYGUARD_CDN: The CDN (content delivery network) from which the core library is served. Defaults to 'https://cdn.nimiq-testnet.com'.
 
 For local development, you can set temporary environment variables (valid in your current terminal session) by e.g.
-```export CONFIG_ALLOWED_ORIGIN='https:\/\my-custom-accounts-instance.com'```
+```export KEYGUARD_ALLOWED_ORIGIN='https:\/\my-custom-accounts-instance.com'```
 on Unix-based systems or
-```set CONFIG_ALLOWED_ORIGIN='https:\/\my-custom-accounts-instance.com'```
+```set KEYGUARD_ALLOWED_ORIGIN='https:\/\my-custom-accounts-instance.com'```
 on Windows. For a persisted configuration on your webserver, please refer to the server's documentation, e.g. [https://httpd.apache.org/docs/2.4/env.html] for Apache.
 
 ## I18n usage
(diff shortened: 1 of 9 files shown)


src/common.js
```diff
@@ -56,18 +56,11 @@ async function runKeyguard(RequestApiClass, options) { // eslint-disable-line no
...
-    function allowedOrigin() {
-        switch (window.location.origin) {
-        case 'https://keyguard-next.nimiq.com': return 'https://accounts.nimiq.com';
-        case 'https://keyguard-next.nimiq-testnet.com': return 'https://accounts.nimiq-testnet.com';
-        default: return '*';
-        }
-    }
+    /** @type {string} */
+    // eslint-disable-next-line no-undef
+    const allowedOrigin = CONFIG.ALLOWED_ORIGIN ? CONFIG.ALLOWED_ORIGIN : '*';
```
sisou: When `CONFIG` is not set, this will result in an `Cannot read ALLOWED_ORIGIN of undefined` error.

Better would be 
```
= CONFIG && CONFIG.ALLOWED_ORIGIN ? ...
```

However, since the build script replaces the thing, this won't work. 

I think you should use one variable for the allowed origin, not a nested object.

tools/build.sh
```diff
@@ -44,6 +37,8 @@ for DIR in src/request/*/ ; do
         cat $DIR/$url >> dist/request/$REQUEST/$JS_BUNDLE
     done
 
+    sed -i -e "s/CONFIG.ALLOWED_ORIGIN/$ALLOWED_ORIGIN/g" dist/request/$REQUEST/$JS_BUNDLE
```
sisou: Where does `$ALLOWED_ORIGIN` come from? It's not defined anywhere. Or is it an env variable?

In that case you need to include instructions on how to set this up.

tools/build.sh
```diff
@@ -1,12 +1,5 @@
 #!/bin/bash
```
sisou: I think this script should look for a `.env` file which includes the `export ALLOWED_ORIGIN = ... `.

Since we are building it only for deployments, our default config should even be part of the repo, I'd argue. But thus easily changeable by the community.

sisou:
What we could do on the CDN side of things, is replace the script tag in each index.html with a little script that adds the script tag to the end of the head. Thus we could have a variable in there that can be overwritten by the config.

tools/build.sh
sisou: I'm thinking the script should load the testnet config by default, and if run with `--mainnet` flag, it loads the mainnet config.

tools/build.sh
sisou: There's also no check here if it's set, so currently if not, it replaces everything with emptiness :(

tools/build.sh
Bettelstab: I think setting all variables per environment is the most flexible way, especially useful for deploying any combination of our apps' versions for testing purposes. So a dedicated config file would be just unnecessary overhead. That's more or less the feedback I got for https://github.com/nimiq/accounts/pull/32 (would also ready for a new shot soon which the same approach).

tools/build.sh
```diff
@@ -1,5 +1,21 @@
...
+    VALUE=$(ECHO "${!VARNAME}")
+
+    if [ "$VALUE" != "" ]; then
+        sed -i -e "s/$1/$VALUE/g" $3
+    fi
+}
+
+# ease configuration of allowed origin by adding quotes
+if [ "$KEYGUARD_ALLOWED_ORIGIN" != "" ]; then
+    KEYGUARD_ALLOWED_ORIGIN="'$KEYGUARD_ALLOWED_ORIGIN'"
```
sisou: I think you changed the variables names when you were writing it, but you have to decide on either `CONFIG_ALLOWED_ORIGIN` or `KEYGUARD_ALLOWED_ORIGIN`.

tools/build.sh
```diff
@@ -37,7 +53,7 @@ for DIR in src/request/*/ ; do
         cat $DIR/$url >> dist/request/$REQUEST/$JS_BUNDLE
     done
 
-    sed -i -e "s/CONFIG.ALLOWED_ORIGIN/$ALLOWED_ORIGIN/g" dist/request/$REQUEST/$JS_BUNDLE
+    replace_config_variable "CONFIG_ALLOWED_ORIGIN" "KEYGUARD_ALLOWED_ORIGIN" dist/request/$REQUEST/$JS_BUNDLE
```
sisou: Decide on a env variable name. KEYGUARD_ALLOWED_ORIGIN is not what you document as a name in the README.

tools/build.sh
```diff
@@ -91,9 +107,7 @@ for DIR in src/request/*/ ; do
         { print }
     ' $DIR/index.html > dist/request/${REQUEST}/index.html
 
-    if [ "$CDN" != "" ]; then
-        sed -i -e "s/https:\/\/cdn.nimiq-testnet.com/$CDN/g" dist/request/${REQUEST}/index.html
-    fi
+    replace_config_variable "https:\/\/cdn.nimiq-testnet.com" "KEYGUARD_CDN" dist/request/${REQUEST}/index.html
```
sisou: The readme says this should be called `CONFIG_CDN`.

src/common.js
```diff
@@ -56,18 +56,11 @@ async function runKeyguard(RequestApiClass, options) { // eslint-disable-line no
...
-    function allowedOrigin() {
-        switch (window.location.origin) {
-        case 'https://keyguard-next.nimiq.com': return 'https://accounts.nimiq.com';
-        case 'https://keyguard-next.nimiq-testnet.com': return 'https://accounts.nimiq-testnet.com';
-        default: return '*';
-        }
-    }
+    /** @type {string} */
+    // eslint-disable-next-line no-undef
+    const allowedOrigin = CONFIG_ALLOWED_ORIGIN || '*';
```
sisou: I think our accounts manager should be the default here, because the keyguard is a sensible little thing with all the keys in it etc. so I think making `*` the default is not good. The default should already be safe.

Wdyt, @danimoh @nibhar?

sisou: I think testnet. Mainnet should require a build flag, just like production build usually require a build flag.

Bettelstab: Problem is, however, that it should default to '*' for unbuilt versions, shouldn't it?

Bettelstab: Happy now? At first I wanted to replace the '*' in source code, but then we would loose our ability to use that as an entry point. If you don't think it's enough, we could reintroduce the check of own origin in src.

README.md
```diff
@@ -45,10 +48,26 @@ Then you can:
...
+    - types
 ```
 
+## Configuration
+
+You can configure the following values by either environment variables or configuration files:
+- KEYGUARD_ALLOWED_ORIGIN: The origin from which requests are accepted. '*' allows all origins. Be aware that slashes have to be masked by `\`. Defaults to 'https://accounts.nimiq-testnet.com'.
+- KEYGUARD_CDN: The CDN (content delivery network) from which the core library is served. Defaults to 'https://cdn.nimiq-testnet.com'.
+
+The most straight forward way is to use a configuration file, which has to be placed in `config` folder, and pass their name as argument to the build script. `yarn build local` uses `local.conf`. Some sample files are provided.
```
sisou:
```suggestion
The best way is to use a configuration file, which has to be placed in the `config` folder, and pass its name as an argument to the build script. `yarn build local` uses `local.conf`. Some sample files are provided.
```

README.md
```diff
@@ -45,10 +48,26 @@ Then you can:
...
+
+You can also set those values in your server configuration via environment variables. Please refer to your server's configuration, e.g. [https://httpd.apache.org/docs/2.4/env.html] for Apache.
```
sisou: I don't quite understand why I would set them in my server. Or do you mean local development server? Because those env variables are not read during runtime, only during build, right?

src/common.js
sisou: Ok, so for locally running it without a build it now uses `*`, but the build script uses the testnet config as the default. I think this is ok now.

tools/build.sh
```diff
@@ -113,9 +128,13 @@ LIST_CSS_TOPLEVEL="../../../node_modules/@nimiq/style/nimiq-style.min.css ../../
 # (since all urls are relative to request directories, we simply use the create request directory as the base)
 for url in $LIST_JS_COMMON; do
     cat src/request/create/$url >> dist/request/$JS_COMMON_BUNDLE
+    if [ "$KEYGUARD_ALLOWED_ORIGIN" != "" ]; then
+        replace_config_variable "CONFIG_ALLOWED_ORIGIN" "KEYGUARD_ALLOWED_ORIGIN" dist/request/$JS_COMMON_BUNDLE
+    fi
 done
 for url in $LIST_JS_TOPLEVEL; do
     cat src/request/create/$url >> dist/request/$JS_TOPLEVEL_BUNDLE
+    replace_config_variable "CONFIG_ALLOWED_ORIGIN" "KEYGUARD_ALLOWED_ORIGIN" dist/request/$JS_COMMON_BUNDLE
```
sisou: In the replace above for the common bundle, you have a check if KEYGUARD_ALLOWED_ORIGIN is set and not empty. Why not here?

tools/build.sh
```diff
@@ -1,11 +1,28 @@
...
-#    "alternative approach is too slow... :("
-#  exit 1
-# fi
+# execute config file given as parameter; default to testnet
+CONFIG_FILE=testnet
+if [ "$1" != "" ]; then
+    CONFIG_FILE="$1"
+fi
+
+source config/$CONFIG_FILE.conf
```
sisou: Add a check if config file exists. Or does `source` fail the script when the file does not exist?

README.md
Bettelstab: Exactly, it's a compile time config. So setting them in server configuration would make sense if you have a CI which builds on the server. I guess that should be clarified.

tools/build.sh
Bettelstab: Thanks, that other check became redundant as I introduced it generically in replace_config_variable().

tools/build.sh
Bettelstab: yep, done

README.md
Bettelstab: updated readme

tools/build.sh
```diff
@@ -1,11 +1,33 @@
...
-#  exit 1
-# fi
+# execute config file given as parameter; default to testnet
+CONFIG_FILE=testnet
+if [ "$1" != "" ]; then
+    CONFIG_FILE="$1"
+fi
+
+if [ ! -f config/$CONFIG_FILE.conf ]; then
+    echo "Config file not found!"
```
sisou: Would be nice to see the file path that was looked for but could not be found in the error message as well.

PR #168, opened 2019-02-06
Add ignored errors as references to ts-ignore comments

sisou:


sisou:
Because it's an "external" lib and will be changed according to @danimoh anyway.

PR #171, opened 2019-02-06
Improve error reporting

nibhar:
This PR adds an additional check to the event handlers for `unhandledrejection` and `error` to improve error reporting.
This should be an improvement over the last version, as it should not allow for `Cannot assign to read only property 'name' of object`

### DIFF
--- a/src/request/TopLevelApi.js
+++ b/src/request/TopLevelApi.js
@@ -1,4 +1,5 @@
 /* global BrowserDetection */
+/* global Errors */
 /* global KeyStore */
 /* global CookieJar */
 /* global I18n */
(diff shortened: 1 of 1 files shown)


nibhar:
Do you think that that code would ever execute, though? `event.error instance of Error` did not evaluate to `true` in that case.

sisou:
@nibhar is right, only `Error` have stacks, thus if the `reason` is not `instanceof Error`, there is no stack. Thus this special handling is only for all values passed into a `reject()`, that are not errors, and which will be processed correctly with these changes.

sisou:
Next time, you can merge single-commit PRs with the "Rebase and merge" option, to not clutter the graph with merge-lines

PR #178, opened 2019-02-09
Only autofocus inputs when document width is above threshold

### LINKED ISSUE #170: Prevent password input autofocus on mobile

nibhar:
This PR checks if the document is wieder than 600px before automatically focusing an input field.

Resolves #170

### DIFF
--- a/src/components/PassphraseInput.js
+++ b/src/components/PassphraseInput.js
@@ -27,7 +27,10 @@ class PassphraseInput extends Nimiq.Observable {
 
         this.$input.placeholder = placeholder;
 
-        this.$eyeButton.addEventListener('click', () => this._changeVisibility());
+        this.$eyeButton.addEventListener('click', () => {
+            this._changeVisibility();
+            this.focus();
+        });
 
         this._onInputChanged();
         this.$input.addEventListener('input', () => this._onInputChanged());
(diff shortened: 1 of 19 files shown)


src/components/PassphraseInput.js
```diff
@@ -27,7 +27,10 @@ class PassphraseInput extends Nimiq.Observable {
 
         this.$input.placeholder = placeholder;
 
-        this.$eyeButton.addEventListener('click', () => this._changeVisibility());
+        this.$eyeButton.addEventListener('click', () => {
+            this._changeVisibility();
+            this.focus();
```
sisou: Can we find out, if the input was focused before visibility was changed, to only re-focus when it had focus before? Or do you think it's expected to open the keyboard when I just click the visibility button on mobile?

src/request/export/ExportFile.js
```diff
@@ -48,12 +49,23 @@ class ExportFile extends Nimiq.Observable {
             alert('Wallet Files are not yet implemented.');
             this._finish();
         });
+        window.addEventListener('hashchange', event => {
+            const newUrl = new URL(event.newURL);
+            if (newUrl.hash === `#${ExportFile.Pages.EXPORT_FILE}`) {
+                this._downloadKeyFilePassphraseBox.reset();
+                if (TopLevelApi.getDocumentWidth() > Constants.MIN_WIDTH_FOR_AUTOFOCUS) {
+                    this._downloadKeyFilePassphraseBox.focus();
```
sisou: Why is this necessary?

src/components/PassphraseInput.js
nibhar: Functionality of the Eye Button is unchanged. I agree that we should talk about it.

src/request/export/ExportFile.js
nibhar: ExportFiles Password Box would not reset in case a user navigates to ExportWords and then back without it.

PR #189, opened 2019-02-11
Configurable network

Bettelstab:
Adds a config parameter to decide which network to connect to. (Was forgotten in previous config PR)
Also fixes a build tool bug (wrong replacement target) which had no effect.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -28,11 +28,9 @@ async function loadNimiq() {
         case Constants.NETWORK.TEST:
             Nimiq.GenesisConfig.test();
             break;
-
         case Constants.NETWORK.MAIN:
             Nimiq.GenesisConfig.main();
             break;
-
         default:
             throw new Errors.KeyguardError(ErrorConstants.Messages.INVALID_NETWORK_CONFIG);
         }
--- a/src/common.js
+++ b/src/common.js
@@ -1,8 +1,17 @@
 /* global Nimiq */
 /* global RpcServer */
+/* global Errors */
+/* global ErrorConstants */
+/* global Constants */
 
 /** @type {Promise<void>?} */
 let __nimiqLoaded = null;
+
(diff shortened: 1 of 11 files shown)


src/common.js
```diff
@@ -15,8 +24,18 @@ async function loadNimiq() {
...
+        switch (CONFIG.NETWORK) {
+        case Constants.NETWORK.TEST:
+            Nimiq.GenesisConfig.test();
+            break;
+
+        case Constants.NETWORK.MAIN:
+            Nimiq.GenesisConfig.main();
+            break;
+
+        default:
```
sisou: You can remove the empty lines between the cases (and default).

src/lib/ErrorConstants.js
```diff
@@ -17,6 +17,8 @@ const ErrorConstants = {
         CANCELED: 'CANCELED',
         // used to signal that a given keyId no longer exist in KG, to be treated by caller.
         KEY_NOT_FOUND: 'keyId not found',
+        // network name does not exist
+        INVALID_NETWORK_CONFIG: 'invalid network config',
```
sisou:
```suggestion
        INVALID_NETWORK_CONFIG: 'Invalid network config',
```

src/request/iframe/index.html
```diff
@@ -14,6 +14,7 @@
         <script defer bundle-common src="../../lib/CookieJar.js"></script>
         <script defer bundle-common src="../../lib/BrowserDetection.js"></script>
         <script defer bundle-common src="../../lib/AccountStore.js"></script>
+        <script defer bundle-common src="../../lib/Constants.js"></script>
```
sisou: I think, since the invalid-network-config error can happen now in ALL requests, the Constants.js needs to be added to all requests' index.html as well... sorry :P

Bettelstab: Should be already there everywhere :)

PR #190, opened 2019-02-11
fix css of #loading to make it cover the whole screen even in presenc…

### LINKED ISSUE #179: Mobile: Fix blurred area in loading state when keyboard is open
Edit: After entering a password in account creation, the keyboard is open. Then, when loading state is set, only the upper half of the remaining screen (without keyboard) is blurred. Seen on Android.

Bettelstab:
…e of dev console or keyboard. Fixes #179 
Tested on Desktop Chrome, Firefox, Safari, Android Chrome, iOS Safari

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -102,13 +102,15 @@
     flex-flow: column;
     justify-content: center;
     align-items: center;
-    position: absolute;
+    position: fixed;
     left: 0;
     top: 0;
+    right: 0;
+    bottom: 0;
     width: 100vw;
     max-width: 100vw;
     height: 100vh;
-    max-height: 100vh;
+    min-height: 100vh;
     z-index: 1;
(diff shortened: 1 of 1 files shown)


src/common.css
```diff
@@ -102,13 +102,15 @@
...
     justify-content: center;
     align-items: center;
-    position: absolute;
+    position: fixed;
     left: 0;
     top: 0;
+    right: 0;
+    bottom: 0;
     width: 100vw;
     max-width: 100vw;
```
nibhar: You changed `max-height` to `min-height`. For `max-width` it was not changed.
```suggestion
    min-width: 100vw;
```

Bettelstab:
> Change one `max` to `min` I guess.

nope! ^^

PR #200, opened 2019-02-14
Auto-incremented key ID

Bettelstab:
Instead of a hash, use an autoincremented number as database index in KeyStore to avoid collisions. Still save the hash and check each time before saving a key if it already exists.

This PR refactors tests a little bit to enable better typechecking there.

I'm going to work on the counterpart in accounts soon, too, which should be finished before merging this PR.


tests/lib/CookieJar.spec.js
```diff
@@ -13,14 +13,14 @@ describe('CookieJar', () => {
 
     it('can be filled with key info', () => {
         spyOnProperty(document, 'cookie', 'set').and.callFake((/** @type {string} */ cookie) => {
-            expect(cookie.includes(`k=${Dummy.keyInfoCookieEncoded}`)).toBe(true);
+            expect(cookie.valueOf().substr(0, Dummy.keyInfoCookieEncoded.length + 2)).toEqual(`k=${Dummy.keyInfoCookieEncoded}`);
```
sisou: Häää?

tests/DummyData.spec.js
```diff
@@ -201,14 +180,36 @@ Dummy.Utils = {
...
+    deprecatedAccountInfos,
+    cookie,
+    encryptionPassword,
+    encryptionPassword2,
+    keyInfoObjects,
+    deprecatedAccount2KeyInfoObject,
+    DUMMY_ACCOUNT_DATABASE_NAME,
+    DUMMY_KEY_DATABASE_NAME,
+    deprecatedAccountRecords
+}
```
sisou: Missing newline :laughing:

tests/lib/IframeApi.spec.js
```diff
@@ -88,8 +89,12 @@ describe('IframeApi', () => {
         expect(KeyStore.instance.migrateAccountsToKeys).toHaveBeenCalled();
 
         // check that keys have been copied correctly
-        const key1 = await KeyStore.instance._get(Dummy.keyInfos[0].id);
-        expect(key1).toEqual(Dummy.keyRecords[0]);
+        const ids = (await KeyStore.instance.list()).map(x => x.id);
```
sisou:
```suggestion
        const ids = (await KeyStore.instance.list()).map(record => record.id);
```

tests/lib/IframeApi.spec.js
```diff
@@ -88,8 +89,12 @@ describe('IframeApi', () => {
...
+        for (let id of ids) {
+            const keyRecord = await KeyStore.instance._get(id);
+            const expectedKeyRecord = /** @type {StoredKeyRecord} */(Dummy.storedKeyRecords().find(x => x.id === id));
```
sisou:
```suggestion
            const expectedKeyRecord = /** @type {StoredKeyRecord} */(Dummy.storedKeyRecords().find(record => record.id === id));
```

tests/lib/KeyStore.spec.js
```diff
@@ -27,25 +27,23 @@ describe('KeyStore', () => {
...
-        expect(key1.secret).toEqual(Dummy.secrets[0]);
-        expect(key1.hasPin).toEqual(Dummy.keyInfos[0].hasPin);
-
-        expect(key2.id).toEqual(Dummy.keyInfos[1].id);
-        expect(key2.type).toEqual(Dummy.keyInfos[1].type);
-        expect(key2.secret).toEqual(Dummy.secrets[1]);
-        expect(key2.hasPin).toEqual(Dummy.keyInfos[1].hasPin);
+
+        for (let [i, key] of keys.entries()) {
+            if (!key) throw new Error();
```
sisou:
```suggestion
            if (!key) throw new Error(`Key with id ${Dummy.keyInfos[i].id} not found!`);
```

tests/lib/KeyStore.spec.js
```diff
@@ -157,12 +155,51 @@ describe('KeyStore', () => {
...
+        currentKeys = await KeyStore.instance.list();
+        expect(currentKeys.length).toBe(1);
+
+        // add key again
+        await KeyStore.instance.put(new Key(Dummy.secrets[1]), passphrase);
+        currentKeys = await KeyStore.instance.list();
+        expect(currentKeys.length).toBe(1);
+
+        // add key again with different password
+        const passphrase2 = Nimiq.BufferUtils.fromAscii(Dummy.encryptionPassword2);
```
sisou: I suggest moving the "generation" of the password buffers to the top together and naming them `password1/2` instead of passphrase.

tests/lib/KeyStore.spec.js
```diff
@@ -157,12 +155,51 @@ describe('KeyStore', () => {
...
         // TODO: Expect Accounts DB to not be deleted
 
         await Dummy.Utils.deleteDummyAccountStore();
     });
 
+    it('doesn\'t store same key twice', async () => {
+        // first clear database
+        await Dummy.Utils.deleteDummyKeyStore();
+
+        let currentKeys = await KeyStore.instance.list();
```
sisou:
```suggestion
        let currentKeys = await KeyStore.instance.list();
        expect(currentKeys.length).toBe(0); // Just to be sure
```

src/lib/KeyStore.js
```diff
@@ -82,30 +82,30 @@ class KeyStore {
...
+        /** @type {?StoredKeyRecord} */
         const keyRecord = await this._get(id);
         return keyRecord ? KeyInfo.fromObject(keyRecord, KeyStore.isEncrypted(keyRecord)) : null;
     }
 
     /**
-     * @param {string} id
-     * @returns {Promise<?KeyRecord>}
+     * @param {number} id
+     * @returns {Promise<?StoredKeyRecord>}
```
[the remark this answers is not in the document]
Bettelstab: Yes, con mucho gusto!

src/lib/Key.js
```diff
@@ -5,15 +5,17 @@ class Key {
...
 
     /**
      * @param {Nimiq.Entropy|Nimiq.PrivateKey} secret
      * @param {boolean} [hasPin]
+     * @param {number?} [id]
      */
-    constructor(secret, hasPin = false) {
+    constructor(secret, hasPin = false, id = null) {
+        this._id = id;
         this._secret = secret;
```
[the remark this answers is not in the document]
Bettelstab: 2. I'll go with `id = undefined` then, so default parameters are not split between function declaration and jsdoc.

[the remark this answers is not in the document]
Bettelstab: ...or I'll leave it, as it breaks return type and I don't like putting number | undefined.

src/lib/KeyStore.js
```diff
@@ -52,9 +52,16 @@ class KeyStore {
...
+                switch (event.oldVersion) {
+                    case 1:
+                        // Version 1 was only used in Testnet, so we'll wipe it out.
+                        db.deleteObjectStore(KeyStore.DB_KEY_STORE_NAME);
+                        // eslint-disable-next-line no-fallthrough
+                    case 0:
+                        db.createObjectStore(KeyStore.DB_KEY_STORE_NAME, { keyPath: 'id', autoIncrement: true });
+                        break;
+                    default:
+                        break;
```
sisou: This is not how this works. This is how this works:

```javascript
if (event.oldVersion < 1) {
    // Version 1 is the first version of the database.
    db.createObjectStore(KeyStore.DB_KEY_STORE_NAME, { keyPath: 'id', autoIncrement: true });
}

if (event.oldVersion < 2) {
    // Version 2 changes the key id calculation, thus we drop and recreate the whole database
    db.deleteObjectStore(KeyStore.DB_KEY_STORE_NAME);
    db.createObjectStore(KeyStore.DB_KEY_STORE_NAME, { keyPath: 'id', autoIncrement: true });
}
```

src/lib/KeyStore.js
```diff
@@ -225,52 +233,67 @@ class KeyStore {
         }
     }
 
-    // eslint-disable-next-line valid-jsdoc
     /**
-     * @param {(AccountInfo|AccountRecord)[]} accounts
-     * @param {boolean} [withAccount]
-     * @returns {KeyguardRequest.KeyInfoObject[]|KeyguardRequest.LegacyKeyInfoObject[]|KeyRecord[]}
+     * @param {AccountRecord[]} accounts
+     * @returns {(KeyRecord)[]}
```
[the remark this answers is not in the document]
Bettelstab: no, forgot to remove them.

src/lib/KeyStore.js
```diff
@@ -225,52 +233,67 @@ class KeyStore {
...
+    static accountInfos2KeyInfos(accounts) {
+        return accounts.map((account, id) => {
+            const address = Nimiq.Address.fromUserFriendlyAddress(account.userFriendlyAddress);
 
             /** @type {KeyguardRequest.KeyInfoObject} */
             const keyObject = {
-                id: legacyKeyId,
-                type: 1, // Nimiq.Secret.Type.PRIVATE_KEY
+                id,
+                type: Nimiq.Secret.Type.PRIVATE_KEY,
```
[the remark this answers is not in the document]
Bettelstab: Ok. But then we can remove it again in that case.

src/lib/KeyStore.js
```diff
@@ -160,11 +167,17 @@ class KeyStore {
         const db = await this.connect();
         const transaction = db.transaction([KeyStore.DB_KEY_STORE_NAME], 'readwrite');
         const request = transaction.objectStore(KeyStore.DB_KEY_STORE_NAME).put(keyRecord);
-        return KeyStore._requestToPromise(request, transaction);
+
+        const dbKey = await KeyStore._requestToPromise(request, transaction);
```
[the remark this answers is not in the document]
Bettelstab: Yes. I'm adding a test for that.

tests/lib/CookieJar.spec.js
Bettelstab: Well, let's have an example. Assume `keyInfoCookieEncoded === '101;102` and the cookie starts with `k=101;102;203`. I wanted to avoid that this passes, but turns out I didn't... Damn.

Bettelstab:
Done with changes.

src/lib/Key.js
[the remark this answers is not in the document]
sisou: Still leaves comment 1, about the order of assignments.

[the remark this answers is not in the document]
Bettelstab: I could have sworn I already changed it...  now done.

src/lib/KeyStore.js
```diff
@@ -225,52 +240,63 @@ class KeyStore {
...
+        return accounts.map((account, index) => {
+            const address = Nimiq.Address.fromUserFriendlyAddress(account.userFriendlyAddress);
 
-            /** @type {KeyguardRequest.KeyInfoObject} */
-            const keyObject = {
-                id: legacyKeyId,
-                type: 1, // Nimiq.Secret.Type.PRIVATE_KEY
+            /** @type {KeyguardRequest.LegacyKeyInfoObject} */
+            const legacyKeyObject = {
+                id: index + 1,
```
sisou: Ok, I actually think forecasting the legacy accounts' `id`s like that is fine now, because even on Safari the migration will happen as soon as a user tries to create or import another account (because that is a top-level invocation of the Keyguard). So I think everything is fine here and this _should_ work nicely.

Bettelstab:
I'd like to include:
The client should proceed to accept requests with strings as keyId. These need to have the form `K{id}` (K for Keyguard). All results should be mapped this way, too. That way, there is no breaking change for Accounts (where I would have done this mapping otherwise).

I'll wait until #202 is merged to avoid merge conflicts.

client/src/PublicRequest.ts
```diff
@@ -96,20 +104,43 @@ export type DeriveAddressResult = {
...
     | SimpleRequest
     | SignTransactionRequest
     | DeriveAddressRequest
     | SignMessageRequest
     | RemoveKeyRequest;
 
-export type RpcResult = KeyResult
+export type IFrameRequest = EmptyRequest | DeriveAddressesRequest | ReleaseKeyRequest;
+
+export type Request = RedirectRequest | IFrameRequest | SimpleResult;
```
sisou: Is this supposed to include `SimpleRequest` instead of `SimpleResult`?

src/request/import/ImportWords.js
```diff
@@ -81,21 +83,20 @@ class ImportWords {
...
-            this._loginFileIcon.lock(`nq-${colorString}-bg`);
+            if (this._keys.entropy.addresses) {
+                const color = Iqons.getBackgroundColorIndex(
+                    new Nimiq.Address(
+                        // use color of first address as loginFile color
+                        this._keys.entropy.addresses[0].address,
+                    ).toUserFriendlyAddress(),
+                );
+                const colorString = LoginFile.CONFIG[color].name;
+                this._loginFileIcon.lock(`nq-${colorString}-bg`);
```
nibhar: This line should execute regardless of wether it is an entropy or a private key. In case of a private key the argument can be omitted, because it only closes the lock but does not change the color.

Bettelstab: Did I introduce a change here accidentally?

Bettelstab:
Tested (with https://github.com/nimiq/accounts/pull/144), seems to work fine.

client/yarn-error.log
```diff
@@ -0,0 +1,360 @@
+Arguments:
```
nibhar: This file should be removed.

src/request/import/ImportWords.js
nibhar: In case of `this._keys.entropy === null` but `this._keys.privateKey !==null` the lock will not close.

src/request/import/ImportWords.js
```diff
@@ -182,7 +183,6 @@ class ImportWords {
      */
     _onRecoveryWordsComplete(mnemonic, mnemonicType) {
         this._secrets = { entropy: null, privateKey: null };
-        this._keys = [];
```
nibhar: this should still reset.

tests/lib/AccountStore.spec.js
```diff
@@ -1,5 +1,6 @@
 /* global Dummy */
 /* global AccountStore */
+
```
nibhar: magnificent

Bettelstab:
Please check again.

src/request/import/ImportWords.js
```diff
@@ -30,8 +30,9 @@ class ImportWords {
 
         /** @type {{entropy: Nimiq.Entropy?, privateKey: Nimiq.PrivateKey?}} */
         this._secrets = { entropy: null, privateKey: null };
-        /** @type {KeyguardRequest.KeyResult[]} */
-        this._keys = [];
+        // eslint-disable-next-line max-len
+        /** @type {{entropy: Partial<KeyguardRequest.SingleKeyResult>, privateKey: Partial<KeyguardRequest.SingleKeyResult>}} */
+        this._keys = { entropy: {}, privateKey: {} };
```
sisou: The variable `this._keys` is misnamed, because it does not contain `Key`s (the class), but only (Single)`KeyResult`s (the object).

tests/DummyData.spec.js
```diff
@@ -9,155 +9,144 @@ beforeAll(async () => {
...
-        type: Dummy.keyInfos[1].type,
-        hasPin: Dummy.keyInfos[1].hasPin,
-    },
-];
+const keyInfoObjects = keyInfos.map(/** @param {KeyInfo} x */x => 
+    ({
+        id: x.id,
+        type: x.type,
+        hasPin: x.hasPin
+    }));
```
sisou: The last closing bracket should be on the next line.

tests/DummyData.spec.js
```diff
@@ -9,155 +9,144 @@ beforeAll(async () => {
...
 _purposeIdBuf.writeUint32(Nimiq.Entropy.PURPOSE_ID);
 const _purposeIdArray = Array.from(_purposeIdBuf.subarray(0, 4));
 
-/** @type {KeyRecord[]} */
-Dummy.keyRecords = [
-    Object.assign({}, Dummy.keyInfoObjects[0], { secret: Dummy.encryptedKeys[0] }),
-    Object.assign({}, Dummy.keyInfoObjects[1], { secret: new Uint8Array(_purposeIdArray.concat(Array.from(Dummy.keys[1]))) }),
+const keyRecords = () => [
+    Object.assign({}, {
+        type: keyInfoObjects[0].type,
```
sisou: Why use `assign` if not inheriting from anything? The copied properties are still references, when they are objects, no?

tests/DummyData.spec.js
```diff
@@ -9,155 +9,144 @@ beforeAll(async () => {
...
+const deprecatedAccountInfos = [
     {
         userFriendlyAddress: 'NQ71 CT4K 7R9R EHSB 7HY9 TSTP XNRQ L2RK 8U4U',
         type: 'high',
         label: 'Dummy account 1',
     },
 ];
 
-Dummy.deprecatedAccountCookie = [
+const deprecatedAccountCookie = [
```
sisou: Technically this is an array, so could be named in the plural.

tests/lib/CookieJar.spec.js
```diff
@@ -13,14 +13,15 @@ describe('CookieJar', () => {
 
     it('can be filled with key info', () => {
         spyOnProperty(document, 'cookie', 'set').and.callFake((/** @type {string} */ cookie) => {
-            expect(cookie.includes(`k=${Dummy.keyInfoCookieEncoded}`)).toBe(true);
+            const cookieValue = cookie.valueOf().split(';')[0];
```
sisou: Why does this need `valueOf()`?

client/src/InternalRequest.ts
```diff
@@ -0,0 +1,33 @@
...
+export * from './PublicRequest';
+
+export type KeyInfoObject = PublicToInternal<Public.KeyInfoObject>;
+export type LegacyKeyInfoObject = PublicToInternal<Public.LegacyKeyInfoObject>;
+export type SimpleRequest = PublicToInternal<Public.SimpleRequest>;
+export type RemoveKeyRequest = PublicToInternal<Public.RemoveKeyRequest>;
+export type SingleKeyResult = PublicToInternal<Public.SingleKeyResult>;
+export type KeyResult = SingleKeyResult[];
+export type DeriveAddressRequest = PublicToInternal<Public.DeriveAddressRequest>;
+export type DeriveAddressesRequest = PublicToInternal<Public.DeriveAddressesRequest>;
```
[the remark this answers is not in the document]
Bettelstab: I totally agree, however I was only consistent there with existing Keyguard commands.

client/src/KeyguardClient.ts
```diff
@@ -1,7 +1,11 @@
 import { RedirectRpcClient } from '@nimiq/rpc';
 import { RequestBehavior, RedirectRequestBehavior, IFrameRequestBehavior } from './RequestBehavior';
 import { KeyguardCommand } from './KeyguardCommand';
-import * as KeyguardRequest from './KeyguardRequest';
+import { Request, CreateRequest, ImportRequest, SimpleRequest, SignTransactionRequest, DeriveAddressRequest,
+    DeriveAddressesRequest, RemoveKeyRequest, ReleaseKeyRequest, EmptyRequest, RedirectRequest,
+    IFrameRequest, IFrameResult, ListResult, SimpleResult, DeriveAddressesResult, ListLegacyResult,
+    RedirectResult} from './PublicRequest';
```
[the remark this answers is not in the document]
Bettelstab: I don't want to write everytime Public.CreateRequest and so on.

client/src/KeyguardClient.ts
```diff
@@ -10,21 +14,56 @@ export class KeyguardClient {
...
+            object.keyId = KeyguardClient.parseId(object.keyId);
+        }
+        if (object && object.id) {
+            object.id = KeyguardClient.parseId(object.id);
+        }
+        return object;
+    }
+
+    // Not really well typed. Return type gets any. Feel free to improve.
+    private static internalToPublic(result: any): InternalToPublic<typeof result> {
```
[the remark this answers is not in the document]
Bettelstab: Precisely, I get back unknown from keyguard (like the keyguard gets a request of type unknown). It would be cleaner to use a response parser.
The result type gets, for redirect requests at least, lost anyway in the Observable.fire() currently. :/
I already started working on another PR which hopefully will solve some of those issues, but I felt there are still more important things to do.

src/request/iframe/IFrameApi.js
```diff
@@ -20,57 +20,59 @@ class IFrameApi {
...
      * @param {RpcState?} state
-     * @returns {Promise<boolean>}
+     * @returns {Promise<KeyguardRequest.SimpleResult>}
      */
     async hasKeys(state) {
         const keyInfos = await this._getKeys();
-        return keyInfos.length > 0;
+        return {
+            success: keyInfos.length > 0,
+        };
```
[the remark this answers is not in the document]
Bettelstab: Yes, because of that.

client/src/KeyguardClient.ts
```diff
@@ -10,21 +14,56 @@ export class KeyguardClient {
...
+        if (result instanceof Array) {
+            return result.map((x) => KeyguardClient.internalToPublic(x));
+        }
+        if (result && result.keyId) {
+            result.keyId = `K${result.keyId}`;
+        }
+        if (result && result.id) {
+            result.id = `K${result.id}`;
```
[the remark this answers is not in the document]
Bettelstab: KeyInfoObject and its legacy variant.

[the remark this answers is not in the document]
Bettelstab: KeyInfoObject and it's legacy variant.

client/src/KeyguardClient.ts
```diff
@@ -42,91 +81,96 @@ export class KeyguardClient {
         return this._redirectClient.init();
     }
 
-    public on(command: KeyguardCommand, resolve: (...args: any[]) => any, reject: (...args: any[]) => any) {
+    public on(
+        command: KeyguardCommand,
+        resolve: (result: any, state?: any) => any,
```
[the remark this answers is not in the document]
Bettelstab: I think types should be refactored in a way we can easily derive a command's result (and request) type. For now, let's stick with `RedirectResult`.

client/src/KeyguardClient.ts
```diff
@@ -42,91 +81,96 @@ export class KeyguardClient {
...
+        const internalRequest = KeyguardClient.publicToInternal(request);
+        this._redirectBehavior.request(this._endpoint, command, [ internalRequest ]);
+        // return value of redirect call is received in _onResolve()
+    }
+
+    private async _iframeRequest<T1 extends IFrameRequest, T2 extends IFrameResult>(
+        command: KeyguardCommand,
+        request?: T1,
+    ): Promise<T2> {
+        const args = request ? [ KeyguardClient.publicToInternal(request) ] : [];
```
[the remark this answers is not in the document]
Bettelstab: But what should the return result be then and how would it be handled? Because you don't want `args === [ null ]`.

client/src/KeyguardClient.ts
```diff
@@ -42,91 +81,96 @@ export class KeyguardClient {
         return this._redirectClient.init();
     }
 
-    public on(command: KeyguardCommand, resolve: (...args: any[]) => any, reject: (...args: any[]) => any) {
+    public on(
+        command: KeyguardCommand,
+        resolve: (result: any, state?: any) => any,
+        reject: (...args: any[]) => any,
```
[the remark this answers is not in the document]
Bettelstab: yep

src/request/TopLevelApi.js
```diff
@@ -159,9 +159,10 @@ class TopLevelApi extends RequestParser { // eslint-disable-line no-unused-vars
 
     /**
      * Overwritten by each request's API class
+     * Note: There is a bug with typescript not parsing 'template {KeyguardRequest.RedirectRequest} T', so we use any.
```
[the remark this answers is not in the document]
Bettelstab: Not sure, I'll check.

src/request/TopLevelApi.js
```diff
@@ -159,9 +159,10 @@ class TopLevelApi extends RequestParser { // eslint-disable-line no-unused-vars
...
      *
-     * @param {KeyguardRequest.Request} request
-     * @returns {Promise<ParsedRequest>}
+     * @param {any} request
+     * @returns {Promise<Parsed<any>>}
```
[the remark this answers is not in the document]
Bettelstab: That would only work if we make TopLevelApi generic and then make e.g. `CreateRequestApi extend TopLevelApi<CreateRequest>`, just more complicated with JSdoc annotation. Not a straightforward task.

client/src/KeyguardClient.ts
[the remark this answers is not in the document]
sisou: The function could return the empty array internally, instead of this outside check. But it's fine now.

src/request/TopLevelApi.js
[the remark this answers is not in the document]
sisou: Ok

PR #202, opened 2019-02-14
No ambient types

Bettelstab:
On client side, removes the need for ambient declaration for Nimiq and KeyguardRequest.


client/src/KeyguardRequest.ts
```diff
@@ -0,0 +1,135 @@
+import * as Nimiq from '@nimiq/core-web';
+
+// tslint:disable-next-line:interface-over-type-literal
```
nibhar: I think this is unnecessary, as the rule is disabled in `tslint.json`.

PR #203, opened 2019-02-14
Move parseRequest to TopLevelApi and load CryptoWorker if needed.

### LINKED ISSUE #143: Move CryptoUtils.getInstanceAsync into TopLevelApi and run always

nibhar:
This PR moves the request parsing call from the`${Request}Api` classes to `TopLevelApi` which all of the former extend.
After the parsing the necessity of `Nimiq.CryptoWorker.getInstanceAsync();` can be determined by the ParsedRequest and is only loaded when needed.

resolves #143


src/request/TopLevelApi.js
```diff
@@ -63,10 +65,23 @@ class TopLevelApi extends RequestParser { // eslint-disable-line no-unused-vars
...
             this._reject = reject;
 
+            if (!parsedRequest) { // should already be rejected here with an Errors.InvalidRequestError()
+                // this really should never happen
+                this.reject(new Errors.InvalidRequestError('Request was not successfully parsed'));
+                return;
+            }
+
+            if (!(/** @type {ParsedSimpleRequest} */(parsedRequest).keyInfo)
+                || /** @type {ParsedSimpleRequest} */(parsedRequest).keyInfo.encrypted) {
```
Bettelstab: What does that mean conceptually?

sisou: It means the CryptoWorker is only loaded for requests which do not contain a `keyInfo` in their request (like `create` and `import`) and for requests where the key is `encrypted`.

sisou: But I agree, a comment explaining this in the code would be good.

PR #206, opened 2019-02-15
Cookie monster service worker

### LINKED ISSUE #195: Implement cookie monster
Implement a cookie monster that eats cookies :cookie:
I.e.: a service worker that intercepts requests to the server and strips cookies.

Bettelstab:
Introduces a service worker which intercepts all outgoing requests and omits cookies.
(Note: Could be extended easily to cache data.)

Resolves #195.


src/common.js
```diff
@@ -12,6 +12,16 @@ const CONFIG = {
     NETWORK: 'test',
 };
 
+// Register service worker if necessary (and possible).
+// This file is always called from a ./request/*/ folder, hence the paths.
+if ('serviceWorker' in navigator) {
+    navigator.serviceWorker.register('../../ServiceWorker.js', {
+        scope: '../../',
+    }).then(reg => {
+        console.log(`Service worker has been registered for scope: ${reg.scope}`);
```
sisou: Can we make this a `console.debug`?

src/ServiceWorker.js
```diff
@@ -0,0 +1,63 @@
...
+    );
+});
+
+// The activate handler takes care of cleaning up old caches.
+self.addEventListener('activate', event => {
+    const currentCaches = [PRECACHE, RUNTIME];
+    // @ts-ignore
+    event.waitUntil(
+        caches.keys().then(cacheNames => cacheNames.filter(cacheName => currentCaches.indexOf(cacheName) === -1))
+            .then(cachesToDelete => Promise.all(cachesToDelete.map(cacheToDelete => caches.delete(cacheToDelete))))
```
sisou: Can we improve the formatting of these two long lines by breaking them up a little, maybe?

src/ServiceWorker.js
```diff
@@ -0,0 +1,63 @@
...
+            // @ts-ignore
+            .then(() => self.clients.claim()),
+    );
+});
+
+
+// Intercept fetch
+self.addEventListener('fetch', event => {
+    // @ts-ignore
+    if (event.request.url.startsWith(self.location.origin)) {
```
danimoh: From my understanding, whether cookies are send or is not determined by the whole origin.
From https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy:
```
Cookies use a separate definition of origins. A page can set a cookie for its own domain or any parent domain, as long as the parent domain is not a public suffix. Firefox and Chrome use the Public Suffix List to determine if a domain is a public suffix. Internet Explorer uses its own internal method to determine if a domain is a public suffix. The browser will make a cookie available to the given domain including any sub-domains, no matter which protocol (HTTP/HTTPS) or port is used. When you set a cookie, you can limit its availability using the Domain, Path, Secure and Http-Only flags. When you read a cookie, you cannot see from where it was set. Even if you use only secure https connections, any cookie you see may have been set using an insecure connection.
```

Thus, rather the `host` of the location should be checked, ignoring the protocol and port.
It shouldn't make a difference though, as our service worker is only active on `https` and the very only port we're using anyways. It's just a beautification.

Bettelstab:
please check again

src/ServiceWorker.js
```diff
@@ -0,0 +1,59 @@
...
+*/
+
+/* eslint-disable no-restricted-globals */
+
+const PRECACHE = 'v1';
+
+// A list of local resources we always want to be cached.
+/** @type {string[]} */
+const PRECACHE_URLS = [
+];
```
sisou: Could be one line.

src/ServiceWorker.js
```diff
@@ -27,34 +27,36 @@ const PRECACHE_URLS = [
...
-            .then(() => self.clients.claim()),
-    );
-});
-
+    // @ts-ignore Property 'waitUntil' does not exist on type 'Event'.ts(2339)
+    event.waitUntil(async () => {
+        const cacheNames = await caches.keys();
+        const cachesToDelete = cacheNames.filter(cacheName => currentCaches.indexOf(cacheName) === -1);
+        await Promise.all(cachesToDelete.map(cacheToDelete => caches.delete(cacheToDelete)));
+        // @ts-ignore Property 'clients' does not exist on type 'Window'.ts
```
danimoh: Cast self if possible

src/ServiceWorker.js
```diff
@@ -27,34 +27,36 @@ const PRECACHE_URLS = [
...
+        return self.clients.claim();
+    });
```
danimoh: `async` method needs to be invoked.

src/ServiceWorker.js
danimoh: Well, you just marked this as resolved....
I still think it should be changed.

src/ServiceWorker.js
```diff
@@ -0,0 +1,63 @@
...
+self.addEventListener('activate', event => {
+    const currentCaches = [PRECACHE, RUNTIME];
+    // @ts-ignore
+    event.waitUntil(
+        caches.keys().then(cacheNames => cacheNames.filter(cacheName => currentCaches.indexOf(cacheName) === -1))
+            .then(cachesToDelete => Promise.all(cachesToDelete.map(cacheToDelete => caches.delete(cacheToDelete))))
+            // @ts-ignore
+            .then(() => self.clients.claim()),
+    );
+});
```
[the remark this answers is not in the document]
danimoh: I think everything but the `fetch` event listener can actually be commented out.

src/ServiceWorker.js
```diff
@@ -27,34 +27,36 @@ const PRECACHE_URLS = [
...
-    event.waitUntil(
-        caches.open(PRECACHE)
-            .then(cache => cache.addAll(PRECACHE_URLS))
-            // @ts-ignore
-            .then(self.skipWaiting()),
-    );
+    // @ts-ignore Property 'waitUntil' does not exist on type 'Event'.ts
+    event.waitUntil(async () => {
+        const cache = await caches.open(PRECACHE);
+        cache.addAll(PRECACHE_URLS);
```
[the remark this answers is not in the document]
Bettelstab: > See comments.
> Also instead of the `@ts-ignore`s try to cast events to `ExtendableEvent`s and `self` to `ServiceWorkerGlobalScope` if these types exist in typescript.

They don't :(

src/ServiceWorker.js
```diff
@@ -48,11 +49,15 @@ self.addEventListener('activate', event => {
...
-    // forward request
-    // @ts-ignore Property 'respondWith' does not exist on type 'Event'.ts,
-    // Property 'request' does not exist on type 'Event'.ts
-    event.respondWith(fetch(event.request, {
-        // omit cookie transmission
-        credentials: 'omit',
-    }));
+    // @ts-ignore Property 'request' does not exist on type 'Event'.ts
+    const requestHost = new URL(event.request.url).host;
+    if (requestHost === location.host) {
```
danimoh: Please add a comment with the link https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy explaining, why we test on host instead of origin.

PR #207, opened 2019-02-15
FlippableHandler provides static function, loading state becomes explicit

nibhar:
This PR makes the Flippable Handler provide a static method instead of inheriting from it. There is no need for inheritance as all functionality is global and not on the instantiated object.

It also makes the loading state more explicit and moves the setting of textContent for the global-close button to TopLevelApi.

No style changes aside from functional ones have been made to the loading overlay. The opacity allows for observing the page build being in progress which might be undesirable.

### DIFF
--- a/src/components/FlippableHandler.js
+++ b/src/components/FlippableHandler.js
@@ -4,7 +4,7 @@ class FlippableHandler {
      * If other classes should be flipped to the backside they need to be added to FlippableHandler.css
      * @param {string} [classname = "flipped"] - Pages with this classname will be on the backside of the flip.
      */
-    static flippable(classname = 'flipped') {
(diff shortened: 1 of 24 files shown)


src/components/FlippableHandler.js
```diff
@@ -4,7 +4,7 @@ class FlippableHandler {
      * If other classes should be flipped to the backside they need to be added to FlippableHandler.css
      * @param {string} [classname = "flipped"] - Pages with this classname will be on the backside of the flip.
      */
-    constructor(classname = 'flipped') {
+    static flippable(classname = 'flipped') {
```
sisou: I think this should be renamed to something like `initFlippable` or `attach()`.

PR #209, opened 2019-02-16
Remove request API's onRequest

nibhar:
This PR removes the `onRequest()` method from the API classes of each requests.

Instead a getter on the Handler class is used to enable the TopLevelApi to execute all code necessary. 

A `onBeforeRun(parsedRequest)` hook is created to allow API classes to execute code before the handler's `run()` is called.

### DIFF
--- a/src/request/TopLevelApi.js
+++ b/src/request/TopLevelApi.js
@@ -129,6 +129,7 @@ class TopLevelApi extends RequestParser { // eslint-disable-line no-unused-vars
                 reject(new Errors.KeyguardError('Handler undefined'));
                 return;
             }
+
             try {
                 const handler = new this.Handler(parsedRequest, resolve, reject);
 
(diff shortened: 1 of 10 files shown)


src/request/TopLevelApi.js
```diff
@@ -125,31 +125,49 @@ class TopLevelApi extends RequestParser { // eslint-disable-line no-unused-vars
...
             });
 
-            this.onRequest(parsedRequest).catch(reject);
-            this.setGlobalCloseButtonText(`${I18n.translatePhrase('back-to')} ${parsedRequest.appName}`);
-            TopLevelApi.setLoading(false);
+            if (!this.Handler) {
+                reject(new Errors.KeyguardError('Handler undefined'));
+                return;
+            }
+            try {
```
sisou: Please add newline between the if and try block here.

src/request/TopLevelApi.js
```diff
@@ -125,31 +125,49 @@ class TopLevelApi extends RequestParser { // eslint-disable-line no-unused-vars
...
+                const handler = new this.Handler(parsedRequest, resolve, reject);
+
+                this.onBeforeRun(parsedRequest);
+
+                this.setGlobalCloseButtonText(`${I18n.translatePhrase('back-to')} ${parsedRequest.appName}`);
+
+                TopLevelApi.setLoading(false);
+
+                handler.run();
```
sisou: Isn't this supposed to be the other way around?
Hander.run() zuerst und danach setLoading(false)?

PR #211, opened 2019-02-16
Import File

sisou:
Also includes the `DownloadLoginFile` component, which is used after import-words.


src/request/export/ExportFile.js
```diff
@@ -31,7 +31,7 @@ class ExportFile extends Nimiq.Observable {
 
         /** @type {HTMLFormElement} */
         const $downloadKeyFilePassphraseBox = ($exportFilePage.querySelector('.passphrase-box'));
-        /** @type {HTMLFormElement} */
+        /** @type {HTMLAnchorElement} */
         const $downloadKeyFile = ($exportFilePage.querySelector('.download-key-file'));
```
sisou: I deliberately did not rename the variables here, because Sebastian is currently working on this file, and I did not want to change more than I had to.

src/request/import/Import.css
```diff
@@ -104,3 +123,32 @@ a#goto-create .nq-icon {
...
 }
 
+.page#download-file .page-body {
+    display: flex;
+    flex-direction: column;
+    justify-content: flex-end;
+}
+
+.page#unlock-account .nq-icon.lock-locked,
+.page#unlock-account .nq-icon.lock-unlocked {
```
danimoh: Move these style rules to where the other `.page#unlock-account` rules are

src/request/import/ImportFile.js
```diff
@@ -118,50 +142,30 @@ class ImportFile {
...
+            // If this code runs, the password was correct
+            /** @type {HTMLElement} */
+            (this.$unlockAccountPage.querySelector('.lock-locked')).classList.add('lock-unlocked');
 
+            const key = new Key(secret, this._flags.hasPin);
+            await KeyStore.instance.put(key, encryptionKey);
             return key;
         } catch (event) {
             console.error(event);
             TopLevelApi.setLoading(false);
```
danimoh: Why is this line in here?

src/request/remove-key/index.html
```diff
@@ -37,11 +37,14 @@
     <script defer bundle-toplevel src="../../translations/index.js"></script>
 
     <script defer src="../../lib/AutoComplete.js"></script>
+    <script defer src="../../lib/Iqons.js"></script>
+    <script defer src="../../lib/QrEncoder.js"></script>
```
danimoh: Needed here?

src/components/DownloadLoginFile.js
```diff
@@ -0,0 +1,257 @@
...
+/* global Errors */
+/* global Iqons */
+
+class DownloadLoginFile extends Nimiq.Observable {
+    /**
+     * @param {HTMLAnchorElement} [$el]
+     * @param {Uint8Array} [secret]
+     * @param {Nimiq.Address} [firstAddress]
+     */
+    constructor($el, secret, firstAddress) {
```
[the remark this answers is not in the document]
sisou: By using `[]` in the JSDoc, they _are_ marked as optional. Am I missing something?

src/components/DownloadLoginFile.js
```diff
@@ -0,0 +1,257 @@
...
+        if (secret.byteLength !== KeyStore.ENCRYPTED_SECRET_SIZE) {
+            throw new Errors.KeyguardError('Can only export encrypted Entropies');
+        }
+
+        const color = Iqons.getBackgroundColorIndex(firstAddress.toUserFriendlyAddress());
+        this.file = new LoginFile(Nimiq.BufferUtils.toBase64(secret), color);
+    }
+
+    /**
+     * TODO: Is this method even necessary?
```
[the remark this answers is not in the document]
sisou: Maybe we need it in Export, @nibhar?

src/components/DownloadLoginFile.js
```diff
@@ -0,0 +1,257 @@
...
+     * @returns {boolean}
+     */
+    _supportsNativeDownload() {
+        return typeof this.$el.download !== 'undefined';
+    }
+
+    /**
+     * @param {string} href
+     * @param {string} filename
+     * @returns {void}
```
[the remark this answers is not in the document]
sisou: Because I thought ESLint will yell at me for not including a return annotation in the JSDoc, but apparently it doesn't. I'm removing it.

src/components/DownloadLoginFile.js
```diff
@@ -0,0 +1,257 @@
...
+
+    _onDownloadEnd() {
+        this.fire(DownloadLoginFile.Events.DOWNLOADED);
+        window.removeEventListener('blur', this._onWindowBlur);
+        window.clearTimeout(this._blurTimeout);
+    }
+
+    _onWindowBlur() {
+        // wait for the window to refocus when the browser download dialog closes
+        this._listenOnce('focus', () => this._onDownloadEnd(), window);
```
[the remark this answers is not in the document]
sisou: Would require doing a `this._onDownloadEnd = this._onDownloadEnd.bind(this)` in the constructor, though, to have the same reference to the listener.

src/request/export/ExportFile.js
```diff
@@ -106,11 +106,13 @@ class ExportFile extends Nimiq.Observable {
      * @param {Key | null} key
      * @param {boolean} isProtected
      */
-    setKey(key, isProtected) {
+    setKey(key, isProtected) { // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
sisou: Because if I remove it, I will have to remove it from lots of other places, thereby adding lots of changes to this and other export-related files, creating merge-conflicts for the future Export PR.

src/request/export/ExportFile.js
```diff
@@ -106,11 +106,13 @@ class ExportFile extends Nimiq.Observable {
...
         this._key = key;
-        this._downloadKeyfile.setSecret(new Uint8Array(0), isProtected); // TODO
-        /** @type {HTMLElement} */(document.getElementById(ExportFile.Pages.EXPORT_FILE))
-            .classList.toggle('show-download', this._key !== null);
+        if (key) {
+            this._downloadKeyfile.setSecret(new Uint8Array(0), key.deriveAddress('m/44\'/242\'/0\'/0\''));
```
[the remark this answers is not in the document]
sisou: I have not touched this file other than for the purpose of removing linting errors because of class name changes.

src/request/export/index.html
```diff
@@ -36,8 +36,11 @@
     <script defer bundle-toplevel src="../../translations/index.js"></script>
 
     <script defer src="../../lib/AutoComplete.js"></script>
+    <script defer src="../../lib/Iqons.js"></script>
+    <script defer src="../../lib/QrEncoder.js"></script>
```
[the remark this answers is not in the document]
sisou: No, they are required by the DownloadLoginFile component.

src/request/import/ImportFile.js
sisou: The `setLoading`? Because there is a `setLoading(true)` above, that displays the loading indicator for the duration that it takes to decode the encrypted key. If the password is wrong, the `catch` block is executed and we want the loading overlay to go away again, so the user can try again.

src/request/remove-key/index.html
sisou: Yes, as dependencies for the LoginFile component.

sisou: (Unnecessary scripts would throw a `checkdeps` error during the Travis CI run.)

src/components/DownloadLoginFile.js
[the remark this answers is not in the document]
danimoh: Oh, perfect :+1:

src/components/DownloadLoginFile.css
```diff
@@ -0,0 +1,68 @@
...
+    position: absolute;
+    left: calc(50% - 5.5rem);
+    top: 25.5rem; /* To lie over the wallet indicator on the LoginFile*/
+    pointer-events: none;
+    display: none;
+    stroke: white;
+}
+
+.download-loginfile .long-touch-indicator.animate {
+    animation: long-touch-indicator-animation .7s linear forwards;
```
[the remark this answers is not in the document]
sisou: It looks alright when trying it out. It actually makes the animation go faster, which seems snappier. The user will hold until the overlay comes anyway, I think. I would not change this.

src/components/PassphraseBox.js
```diff
@@ -57,7 +57,7 @@ class PassphraseBox extends Nimiq.Observable {
         /** @type {{[i18nTag: string]: string}} */
         const buttonVersions = {
             'passphrasebox-continue': '<button class="submit" data-i18n="passphrasebox-continue">Continue</button>',
-            'passphrasebox-log-in': '<button class="submit" data-i18n="passphrasebox-log-in">Log in to your wallet</button>',
+            'passphrasebox-log-in': '<button class="submit" data-i18n="passphrasebox-log-in">Unlock</button>',
```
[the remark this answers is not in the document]
sisou: It is:
https://github.com/nimiq/keyguard-next/pull/211/files#diff-8ead4e7e2cc829b218deb44dd46c5739R85
https://github.com/nimiq/keyguard-next/pull/211/files#diff-8ead4e7e2cc829b218deb44dd46c5739R231

src/components/PassphraseBox.js
```diff
@@ -122,7 +122,7 @@ class PassphraseBox extends Nimiq.Observable {
      */
     _onInputChangeValidity(isValid) {
         this._isInputValid = isValid;
-        this.$el.classList.toggle('input-valid', isValid);
+        this.$el.classList.toggle('input-eligable', isValid);
```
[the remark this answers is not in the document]
sisou: In `PassphraseBox.css`. It was already changed to `input-eligable` in an earlier PR, but forgotten here. So this is actually just a bugfix.

src/components/DownloadLoginFile.js
```diff
@@ -0,0 +1,257 @@
...
+
+    _onHeldToDownload() {
+        this.$el.classList.add('held-to-download');
+    }
+
+    _onDownloadEnd() {
+        this.fire(DownloadLoginFile.Events.DOWNLOADED);
+        window.removeEventListener('blur', this._onWindowBlur);
+        window.clearTimeout(this._blurTimeout);
+    }
```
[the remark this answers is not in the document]
sisou: This looks really good and clean. The problem I found is that the window blur event listener is actually used by two start points. Once for the regular click (together with the 500ms wait for no-blur) and once for the secondary click, where no parallel wait is started.

I can't bring them into the same function without the possibility of starting and attaching multiple events, if use right-clicks multiple times (without actually clickin on safe-as). So I think the way of cancelable timeouts is actually not bad.

src/components/DownloadLoginFile.js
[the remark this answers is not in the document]
nibhar: I'd say leave it in. In case I don't use it, I can still delete it then.

src/request/import/ImportFile.js
```diff
@@ -2,7 +2,7 @@
 /* global Nimiq */
 /* global Key */
 /* global ImportWords */
-/* global FileImport */
+/* global FileImporter */
```
nibhar: why the name change?

src/request/import/ImportFile.js
```diff
@@ -118,50 +142,30 @@ class ImportFile {
...
+            const encryptionKey = Utf8Tools.stringToUtf8ByteArray(passphrase);
+
+            // Make sure read position is at 0 even after a wrong passphrase
+            this._encryptedKey.reset();
+
+            const secret = await Nimiq.Secret.fromEncrypted(this._encryptedKey, encryptionKey);
+
+            // If this code runs, the password was correct
+            /** @type {HTMLElement} */
+            (this.$unlockAccountPage.querySelector('.lock-locked')).classList.add('lock-unlocked');
```
nibhar: so this results in a `.lock-locked.lock-unlocked`?

src/request/import/ImportWords.js
```diff
@@ -83,7 +83,7 @@ class ImportWords {
         this._passwordSetter.on(PassphraseSetterBox.Events.ENTERED, () => {
             const color = Iqons.getBackgroundColorIndex(
                 new Nimiq.Address(
-                    /*
+                    /**
```
nibhar: Why the change?

src/translations/index.js
```diff
@@ -35,19 +35,21 @@ const TRANSLATIONS = {
         'import-heading-protect': 'Protect your wallet',
         'import-text-set-password': 'You can now set a password to encrypt your wallet on this device.',
         'import-upload-login-file': 'Upload your Login File',
+        'import-unlock-account': 'Unlock your account',
         'import-create-account': 'Create new Account',
 
         'import-file-lost-file': 'Lost your Key File? You can recover your account with your 24 Recovery Words.',
         'import-file-button-words': 'Login with Recovery Words',
         'import-file-heading-unlock': 'Unlock your Key File',
```
nibhar:
```suggestion
        'import-file-heading-unlock': 'Unlock your Login File',
```
Potentially unused. Still i just saw it there.
However there are more of them left in `translations/index.js`, so probably better to do in a separate wording PR.

src/translations/index.js
```diff
@@ -245,12 +247,9 @@ const TRANSLATIONS = {
         'identicon-selector-generate-new': 'Neu generieren',
         'identicon-selector-more-addresses': 'Mehr Adressen',
 
-        'downloadkeyfile-heading-protected': 'Dein Schlüsseldatei ist geschützt!',
-        'downloadkeyfile-heading-unprotected': 'Dein Schlüsseldatei ist nicht geschützt!',
-        'downloadkeyfile-safe-place': 'Lagere sie in einem sicheren Ort. Wenn du sie verlierst, '
-                                    + 'kann sie nicht wiederhergestellt werden!',
-        'downloadkeyfile-download': 'Schlüsseldatei herunterladen',
-        'downloadkeyfile-download-anyway': 'Trotzdem herunterladen',
+        'download-loginfile-download': 'LoginDatei herunterladen',
```
nibhar:
```suggestion
        'download-loginfile-download': 'Login Datei herunterladen',
```

src/request/import/ImportFile.js
sisou: I found the names `ImportFile` and `FileImport` to be too similar and confusing, thus I made this component into the `FileImporter`.

src/request/import/ImportFile.js
sisou: Yes. And since one is defined after the other, it works :D

src/components/DownloadLoginFile.js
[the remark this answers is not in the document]
danimoh: Hum, ok, then how about this:

```js
async download(fromContextMenu = false) {
    // cancel previous download listeners
    if (this._cancelDownload) {
        this._cancelDownload();
    }
    try {
        // If the window gets blurred by opening a download dialog, consider the download as finished,
        // when the window get's focused again. If no download dialog opens, consider the download
        // as successful after a short time.
        await new Promise((resolve, reject) => {
            if (!fromContextMenu) {
                // only add fallback timeout if not initiated from context menu, as "save as" always opens
                // a dialog.
                window.setTimeout(resolve, 500);
            }
            window.addEventListener('blur', resolve, { once: true });
            this._cancelDownload = reject;
        });

        // if window was blurred,wait for it to get focused again
        if (!document.hasFocus()) {
            await new Promise((resolve, reject) => {
                window.addEventListener('focus', resolve, { once: true });
                this._cancelDownload = reject;
            });
        }

        // download finished
        this.fire(DownloadLoginFile.Events.DOWNLOADED);
    catch (e) {}
    finally {
        this._cancelDownload = null;
    }
}
```

src/components/FileImporter.css
```diff
@@ -0,0 +1,74 @@
...
+
+.file-import .error-message {
+    font-size: 2rem;
+    color: var(--nimiq-red);
+    pointer-events: none;
+}
+
+.file-import input {
+    position: absolute;
+    top: -1000rem; /* out of the viewport */
```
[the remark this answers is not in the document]
danimoh: From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file: 
"Note: opacity is used to hide the file input instead of visibility: hidden or display: none, because assistive technology interprets the latter two styles to mean the file input isn't interactive."
Not sure though whether that's also an issue when using the `label` technique.

And I would use `vh` instead of `rem` as `vh` would also work for insanely huge screens :D

src/request/export/ExportFile.js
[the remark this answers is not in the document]
danimoh: Might be changed, when the `eligable` will be changed.

src/components/DownloadLoginFile.css
```diff
@@ -22,7 +22,7 @@
 }
 
 .download-loginfile .long-touch-indicator.animate {
-    animation: long-touch-indicator-animation .7s linear forwards;
+    animation: long-touch-indicator-animation 800ms linear forwards;
```
danimoh: So you actually changed it to 800 :D

src/components/FileImporter.css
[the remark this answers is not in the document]
sisou: How usable our interface is without touch, but only assistive technologies, is questionable anyway. I made it display:none now.

src/components/DownloadLoginFile.js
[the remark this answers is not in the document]
sisou: Ok, I managed to use this model for both the download and the long-touch handling. Please check last commits.

src/components/DownloadLoginFile.css
sisou: Had to change it back to 700ms because the indicator did not fill up completely with 800. I also added a comment explaining this for future generations.

src/request/import/ImportFile.js
nibhar: I do not really like it though. It is a binary mutually exclusive state.

nibhar:
```suggestion
            (this.$unlockAccountPage.querySelector('.lock-locked')).classList.replace('lock-locked', 'lock-unlocked');
```

src/components/DownloadLoginFile.js
```diff
@@ -116,81 +114,85 @@ class DownloadLoginFile extends Nimiq.Observable {
...
             this._onDownloadStart();
-        } else if (event.button === 2) { // secondary button
-            window.addEventListener('blur', this._onWindowBlur);
+        // } else if (event.button === 2) { // secondary button
+        //     // Big Problem with the secondary button listening:
+        //     // When user does right-click, but then does NOT save-as,
+        //     // then unfocues the window for whatever reason and then
+        //     // refocuses it, the download is considered finished
+        //     // and the request is returned.
+        //     this._onDownloadStart(true);
```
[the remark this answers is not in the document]
sisou: The user can just continue by clicking on the big fat blue "Download Login File" button, or click regularily on the image.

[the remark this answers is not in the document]
sisou: I will leave it as is, until the ServiceWorker can substitute a _real_ download.

src/components/DownloadLoginFile.js
```diff
@@ -116,81 +114,85 @@ class DownloadLoginFile extends Nimiq.Observable {
...
-
-    _onLongTouchComplete() {
-        this._hideLongTouchIndicator();
-        this.$el.classList.add('long-touch-downloaded');
-    }
+        try {
+            await new Promise((resolve, reject) => {
+                // Consider the long-touch successfull after LONG_TOUCH_DURATION
+                window.setTimeout(resolve, DownloadLoginFile.LONG_TOUCH_DURATION);
+                this.$loginfile.addEventListener('touchend', reject, { once: true });
```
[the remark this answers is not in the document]
sisou: I can add a `_cancelTouch` storage variable for the promise's reject function, which can then be called on touchstart. Would that fullfil that requirement?

src/components/DownloadLoginFile.js
```diff
@@ -119,16 +119,18 @@ class DownloadLoginFile extends Nimiq.Observable {
...
             this._onDownloadStart();
-        // } else if (event.button === 2) { // secondary button
-        //     // Big Problem with the secondary button listening:
-        //     // When user does right-click, but then does NOT save-as,
-        //     // then unfocues the window for whatever reason and then
-        //     // refocuses it, the download is considered finished
-        //     // and the request is returned.
-        //     this._onDownloadStart(true);
+        } else if (event.button === 2) { // secondary button
+            this._onDownloadStart(true);
```
danimoh: Might be a little bit simplified:
```js
if (!this._supportsNativeDownload()) return;
this._onDownloadStart(event.button === 2);
```

src/components/DownloadLoginFile.js
```diff
@@ -153,10 +155,8 @@ class DownloadLoginFile extends Nimiq.Observable {
 
             // If window gets blurred, wait for it to get focused again
             if (!document.hasFocus()) {
-                await new Promise((resolve, reject) => {
-                    window.addEventListener('focus', resolve, { once: true });
-                    this._cancelDownload = reject;
-                });
+                this.$el.classList.add('maybe-downloaded');
```
danimoh: Comment above is now not correct anymore

src/components/DownloadLoginFile.js
```diff
@@ -119,16 +119,18 @@ class DownloadLoginFile extends Nimiq.Observable {
      * @param {MouseEvent} event
      */
     _onMouseDown(event) {
+        /** @type {HTMLElement} */
+        const target = (event.target);
+        if (target.matches('.continue')) {
+            this._onDownloadEnd();
+            return;
```
danimoh: Is this needed?
The continue button already has a stopPropagation

src/components/DownloadLoginFile.js
sisou: This actually leads to a different result, has in the `_onDownloadStart` function is not called, when native download is not supported. In my existing code, the download-start function is still called for secondary-clicks, even if native download is not supported.

src/components/DownloadLoginFile.js
sisou: As discussed, this handling is for two different events: `mousedown` and `click`. I removed the `_onDownloadEnd()` call from this `mousedown` handler, as the `click` handler should cover both mobile and desktop usages of the Continue button.

PR #213, opened 2019-02-17
LoginFile adaptions

sisou:
4-times the pixels (double-sized, to look better on high-density screen such as phones), better wording and spacing.


src/lib/LoginFile.js
```diff
@@ -21,8 +21,8 @@ class LoginFile {
 
     static calculateQrPosition() {
         return {
-            x: 69,
-            y: 287,
+            x: 138,
+            y: 536,
```
danimoh: So, this is not `2 * 287` on purpose?

src/lib/LoginFile.js
```diff
@@ -31,7 +31,7 @@ class LoginFile {
     }
 
     filename() {
-        const walletName = 'Nimiq Login File '
+        const walletName = 'Nimiq LoginFile '
```
danimoh: Needs to be changed to `Login File` now again.

src/lib/LoginFile.js
```diff
@@ -115,11 +115,11 @@ class LoginFile {
         /* eslint-disable no-multi-spaces */
         const $canvas = QrEncoder.render({
             text: encodedSecret,
-            radius: 0.7,    // We encode 56 bytes. To keep within a smaller QR base size, we need to reduce
+            radius: 0.5,    // We encode 56 bytes. To keep within a smaller QR base size, we need to reduce
```
danimoh: Comment still says `.7`.
That sentence and this line can just be removed. `.5` is already the default.

src/lib/LoginFile.js
```diff
@@ -115,11 +115,11 @@ class LoginFile {
...
             ecLevel: 'M',   // the error-correction level to M. Thus we reduce the radius from .8 to .7
             fill: 'white',  // to reduce scanning issues.
             background: 'transparent',
-            size: Math.min(240, (window.innerWidth - 64)),
+            size: LoginFile.QR_SIZE,
         });
         /* eslint-enable no-multi-spaces */
         if (!$canvas) throw new Error('Cannot draw QR code');
```
danimoh: Why does that canvas get appended to a div?

src/lib/LoginFile.js
```diff
@@ -78,17 +78,17 @@ class LoginFile {
...
         await loaded;
-        this._ctx.drawImage(img, 58, 43, 198.6, 24);
+        this._ctx.drawImage(img, 116, 86, 397.2, 48);
```
danimoh: Could use `this._drawDataUrlImage` here too?

src/lib/LoginFile.js
sisou: Every element except the logo and the date moved up, because of spacing changes.

src/lib/LoginFile.js
sisou: Oh, must have overlooked this. Thanks.

src/lib/LoginFile.js
sisou: It doesn't, apparently, but I did not change that in this PR :stuck_out_tongue: 
Removed it now ;)

src/lib/LoginFile.js
```diff
@@ -102,28 +99,24 @@ class LoginFile {
 
     _drawWarningText() {
         const x = LoginFile.WIDTH / 2;
-        const y = LoginFile.HEIGHT - 32;
+        const y = LoginFile.HEIGHT - 86 - 2;
```
nibhar: If not doing `LoginFile.HEIGHT - 88` I would prefer a comment explaining the different parts of the calculation.

src/lib/LoginFile.js
```diff
@@ -102,28 +99,24 @@ class LoginFile {
...
      */
     _drawQrCode(encodedSecret) {
-        const $el = document.createElement('div');
-        /* eslint-disable no-multi-spaces */
         const $canvas = QrEncoder.render({
             text: encodedSecret,
-            radius: 0.7,    // We encode 56 bytes. To keep within a smaller QR base size, we need to reduce
-            ecLevel: 'M',   // the error-correction level to M. Thus we reduce the radius from .8 to .7
-            fill: 'white',  // to reduce scanning issues.
+            radius: 0.5,
```
nibhar: Why did the radius decrease even more? Changing it from .8 to .7 was worth a comment. Why is it now .5 without the comment?

src/lib/LoginFile.js
sisou: Only because I don't want to write that number.

src/lib/LoginFile.js
sisou: Because it looks bad in such a high resolution. The gaps in the radiuses are very visible in high res.

danimoh: Actually .5 is the maximum allowed value.
Values over .5 look glitchy.

sisou: Yes. It's not too bad with low resolution, as the missing pixels hide it somewhat, but sadly it became obvious now.

PR #217, opened 2019-02-19
update core-web and typescript

Bettelstab:
Updates core-web and typescript packages and fixes all typing issues after the update.

### DIFF
--- a/types/Keyguard.d.ts
+++ b/types/Keyguard.d.ts
@@ -5,8 +5,7 @@ interface Newable {
     new(...args: any[]): any
 }
 
-type DOMEvent = Event & {
-    target: Element
+interface Event {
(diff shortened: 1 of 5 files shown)


types/Keyguard.d.ts
```diff
@@ -5,8 +5,7 @@ interface Newable {
     new(...args: any[]): any
 }
 
-type DOMEvent = Event & {
-    target: Element
+interface Event {
     data: any
 }
```
sisou: If we are adding `data` to an `Event`, why not also add `target: Element`?

Bettelstab: Because it's already defined in Event (as EventTarget | null).

PR #224, opened 2019-02-22
Import words result fix

nibhar:
This PR changes Import Words to return an array of SingleKeyResults instead of an Object. 
The counter Part in Accounts that can deal with this array is still pending and only the first returned array element will get checked.

### DIFF
--- a/src/components/FlippableHandler.css
+++ b/src/components/FlippableHandler.css
@@ -39,3 +39,8 @@
 #rotation-container > .page:target {
     z-index: 3;
 }
+
+.nq-card.nq-blue-bg {
+    background-image: var(--nimiq-blue-bg);
+    color: white;
+}
--- a/src/request/import/ImportWords.js
+++ b/src/request/import/ImportWords.js
@@ -185,14 +185,16 @@ class ImportWords {
                     keyPath,
                     address: key.deriveAddress(keyPath).serialize(),
                 }));
-                this._keyResults.push({
+
+                /** @type {KeyguardRequest.SingleKeyResult?} */
+                const result = {
                     keyId: await KeyStore.instance.put(key, encryptionKey || undefined),
                     keyType: Nimiq.Secret.Type.ENTROPY,
                     addresses,
-                });
-                // this._keyResults.entropy.keyId = await KeyStore.instance.put(key, encryptionKey || undefined);
+                };
+                this._keyResults.push(result);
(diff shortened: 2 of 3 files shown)


src/request/import/ImportWords.js
```diff
@@ -31,8 +31,8 @@ class ImportWords {
         /** @type {{entropy: Nimiq.Entropy?, privateKey: Nimiq.PrivateKey?}} */
         this._secrets = { entropy: null, privateKey: null };
         // eslint-disable-next-line max-len
-        /** @type {{entropy: Partial<KeyguardRequest.SingleKeyResult>, privateKey: Partial<KeyguardRequest.SingleKeyResult>}} */
-        this._keyResults = { entropy: {}, privateKey: {} };
+        /** @type {KeyguardRequest.KeyResult} */
+        this._keyResults = [];
```
Bettelstab: should be renamed to singular form then probably

Bettelstab: or maybe not... I don't know.

nibhar: It is 1 or 2, so i think results is fine.

src/components/FlippableHandler.css
```diff
@@ -39,3 +39,8 @@
 #rotation-container > .page:target {
     z-index: 3;
 }
+
+.nq-card.nq-blue-bg {
```
nibhar: This is a fix because any `.nq-card.nq-blue-bg` currently has the style of `.nq-card` and not `.nq-card-bg`.

src/request/import/ImportWords.js
```diff
@@ -84,11 +83,12 @@ class ImportWords {
...
         this._passwordSetter.on(PassphraseSetterBox.Events.ENTERED, () => {
             let colorClass = '';
-            if (this._keyResults.entropy.addresses) {
+            if (this._secrets.entropy) {
+                const key = new Key(this._secrets.entropy, false);
                 const color = Iqons.getBackgroundColorIndex(
                     new Nimiq.Address(
                         // use color of first address as loginFile color
-                        this._keyResults.entropy.addresses[0].address,
+                        key.deriveAddress('m/44\'/242\'/0\'/0\'').serialize(),
```
Bettelstab: use Contants.DEFAULT_KEY_PATH :)

nibhar: Is already in the Export PR. Wanted to change it there once merged, but will do here as well now.

PR #225, opened 2019-02-22
Sign Message

sisou:


src/lib/Key.js
```diff
@@ -145,5 +139,5 @@ class Key {
     }
 }
 
-Key.MSG_PREFIX = new Nimiq.SerialBuffer(Nimiq.BufferUtils.fromAscii('Nimiq Signed Message:\n'));
-Key.MSG_PREFIX_LENGTH = Key.MSG_PREFIX.byteLength;
+Key.MSG_PREFIX = 'Nimiq Signed Message: ';
+Key.MSG_PREFIX_LENGTH = 22;
```
Bettelstab: Why not calculate this from MSG_PREFIX?

src/request/sign-message/SignMessage.js
```diff
@@ -0,0 +1,116 @@
...
+     */
+    async _onConfirm(request, resolve, reject, passphrase) {
+        TopLevelApi.setLoading(true);
+
+        const passphraseBuf = passphrase ? Utf8Tools.stringToUtf8ByteArray(passphrase) : undefined;
+
+        /** @type {Key | null} */
+        let key = null;
+        try {
+            /** @type {Key} */ // Request parsing already validates that the key exists
```
nibhar: As discussed in other PR's I think this should still be an explicit check.

PR #226, opened 2019-02-23
Change password overhaul

Bettelstab:


Bettelstab:
Not yet done: changed wording for legacy

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+class ChangePassword {
+    /**
+     * @param {Parsed<KeyguardRequest.SimpleRequest>} request
+     * @param {ChangePassword.resolve} resolve
+     * @param {reject} reject
+     */
+    constructor(request, resolve, reject) {
+        this._resolve = resolve;
+        this._request = request;
+        this._reject = reject;
```
sisou: Order of params is off. `_request` should be moved above ` _resolve`.

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+
+        /** @type {Key} */
+        // eslint-disable-next-line no-unused-expressions
+        this._key;
```
sisou: Could you make this `{Key?}` and assign it `null` instead of the eslint-disable?

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+
+        // Pages
+        /** @type {HTMLFormElement} */
+        this.$enterPassword = (document.getElementById(ChangePassword.Pages.ENTER_PASSWORD));
+        /** @type {HTMLFormElement} */
+        this.$setPassword = (document.getElementById(ChangePassword.Pages.SET_PASSWORD));
```
sisou: The `$enterPassword` and `$setPassword` page variables are only used within the constructor, thus they don't need to be assigned to the instance but can be local `const` only.

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+        this._progressIndicator = new ProgressIndicator(
+            document.querySelector(`#${ChangePassword.Pages.SET_PASSWORD} .progress-indicator`),
+            4,
+            2,
+        );
+        // eslint-disable-next-line no-new
+        new ProgressIndicator(
+            document.querySelector(`#${ChangePassword.Pages.DOWNLOAD_FILE} .progress-indicator`),
+            4,
+            4,
```
sisou: Please add a comment why this jumps from 2 to 4 and leaves out step 3 (because step 3 is used below for the password repetition).

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+        this._passwordGetter.on(PassphraseBox.Events.SUBMIT, this._unlock.bind(this));
+
+        this._passwordSetter.on(PassphraseSetterBox.Events.ENTERED, () => {
+            let colorClass = '';
+            if (this._key.secret instanceof Nimiq.Entropy) {
+                const color = Iqons.getBackgroundColorIndex(
+                    new Nimiq.Address(
+                        // use color of first address as loginFile color
+                        this._key.deriveAddress(Constants.DEFAULT_DERIVATION_PATH).serialize(),
+                    ).toUserFriendlyAddress(),
```
sisou: #229 adds `this._key.defaultAddress` to easily get the first address of a key. Reminder to replace this whole block with

```js
this._key.defaultAddress.toUserFriendlyAddress()
```

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+            if (this._key.secret instanceof Nimiq.PrivateKey) {
+                this._resolve({ success: true });
+                return;
+            }
+
+            // Prepare Login File for download
+
+            const firstAddress = new Nimiq.Address(
+                this._key.deriveAddress(Constants.DEFAULT_DERIVATION_PATH).serialize(),
+            );
```
sisou: Same here with the `key.defaultAddress`.

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+
+            const encryptedEntropy = await this._key.secret.exportEncrypted(passwordBytes);
+
+            downloadLoginFile.setEncryptedEntropy(
+                /** @type {Nimiq.SerialBuffer} */ (encryptedEntropy),
```
sisou: Maybe, instead of double-encryption, we could read the encrypted entropy from the database after it was stored and hand it to the LoginFile from there.

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+            if (e.message === 'Invalid key') {
+                TopLevelApi.setLoading(false);
+                this._passwordGetter.onPassphraseIncorrect();
+                return;
+            }
+            this._reject(new Errors.CoreError(e));
+            return;
+        }
+        if (!key) {
+            this._reject(new Errors.KeyNotFoundError());
```
sisou: _Technically_, key-existance is already checked during request parsing. I know that this check is also for Typescript, but it's still redundant and this code should never get triggered. I think we should make this error into something like a `ShouldNeverHappenError`, so that we are alerted when unexpected code paths are triggered. @nibhar?

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+        TopLevelApi.setLoading(false);
+    }
+
+    /**
+     * @param {string} phrase
+     */
+    async _finish(phrase) {
+        TopLevelApi.setLoading(true);
+        if (!this._key) {
+            this._reject(new Errors.KeyguardError('Bypassed Password'));
```
sisou:
```suggestion
            this._reject(new Errors.KeyguardError('Bypassed password setter'));
```
(Please don't use Githubs `Accept this suggestion` feature, but instead do this change yourself together with other changes in one commit.)

@nibhar This would also be a good place for the `ShouldNeverHappenError` ;)

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+        if (TopLevelApi.getDocumentWidth() > Constants.MIN_WIDTH_FOR_AUTOFOCUS) {
+            this._passwordSetter.focus();
+        }
+        TopLevelApi.setLoading(false);
+    }
+
+    /**
+     * @param {string} phrase
+     */
+    async _finish(phrase) {
```
sisou: Please rename `phrase` to `newPassword`.

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,211 @@
...
+        window.location.hash = ChangePassword.Pages.ENTER_PASSWORD;
+        if (TopLevelApi.getDocumentWidth() > Constants.MIN_WIDTH_FOR_AUTOFOCUS) {
+            this._passwordGetter.focus();
+        }
+    }
+
+    /**
+     * @param {string} password
+     */
+    async _unlock(password) {
```
sisou: The parameter should be named `oldPassword` instead, to differentiate between the passwords (for code clarity).

src/request/change-password/index.html
```diff
@@ -0,0 +1,175 @@
...
+    <script defer src="../../components/LoginFileIcon.js"></script>
+    <script defer src="../../components/ProgressIndicator.js"></script>
+    <script defer src="../../components/PassphraseSetterBox.js"></script>
+    <script defer src="ChangePassword.js"></script>
+    <script defer src="ChangePasswordApi.js"></script>
+    <script defer src="index.js"></script>
+
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Muli:400,600,700">
+    <link rel="stylesheet"
+        href="https://fonts.googleapis.com/css?family=Fira+Mono&text=0123456789ABCDEFGHJKLMNPQRSTUVXY">
```
sisou: There's really no need to have this in two lines, because no other index.html does it (the monospace font might even be unnecessary in this request? Is it even during change-password?)

src/request/change-password/index.html
```diff
@@ -0,0 +1,175 @@
...
+
+<body class="loading">
+    <div class="header-top">
+        <div class="nimiq-app-name">
+            <span class="nimiq-logo"></span>
+        </div>
+    </div>
+
+    <div id="app">
+        <div class="flex-grow"></div>
```
sisou: The `flex-grow` elements should be unnecessary now with recent `common.css` fixes.

src/request/change-password/index.html
```diff
@@ -0,0 +1,175 @@
...
+                <div class="progress-indicator"></div>
+                <a tabindex="0" class="page-header-back-button nq-icon arrow-left"></a>
+                <h1 class="nq-h1" data-i18n="change-password-set-password-heading">Create a new password</h1>
+            </div>
+
+            <div class="page-body nq-card-body">
+                <p data-i18n='change-password-set-password-text' class="nq-light-blue nq-text center">Secure your new Login File.</p>
+                <div class="flex-grow"></div>
+                <div class="nq-icon login-file-icon"></div>
+                <div class="flex-grow"></div>
```
sisou: The alignment can be managed in CSS without the need for `flex-grow` elements.

src/request/change-password/index.html
```diff
@@ -0,0 +1,175 @@
...
+                <a tabindex="0" class="page-header-back-button nq-icon arrow-left"></a>
+                <h1 class="nq-h1" data-i18n="change-password-download-login-file">Download new Login File</h1>
+            </div>
+
+            <div class="page-body nq-card-body">
+                <a class="download-login-file"></a>
+            </div>
+        </div>
+
+        <div id="unlock-account" class="page nq-card">
```
sisou: I beleve this page is unused.

src/request/change-password/index.html
```diff
@@ -0,0 +1,175 @@
...
+                <form class="password-box"></form>
+            </div>
+        </div>
+
+        <button class="global-close nq-button-s display-none">
+            <span class="nq-icon arrow-left"></span>
+            <span id="global-close-text"></span>
+        </button>
+
+        <div class="flex-grow"></div>
```
sisou: Same here: I think this is not necessary anymore.

src/translations/index.js
```diff
@@ -244,14 +250,14 @@ const TRANSLATIONS = {
         'export-button-words': 'Wiederherstellungswörter anzeigen',
         'export-button-file': 'Schlüsseldatei herunterladen',
         'export-finish': 'Export abschließen',
-        'export-more-options-heading': 'Zusätzliche Export Optionen',
+        'export-more-options-heading': 'Zusätzliche Exportoptionen',
```
sisou: Unrelated change.

src/components/PassphraseSetterBox.js
```diff
@@ -61,6 +64,10 @@ class PassphraseSetterBox extends Nimiq.Observable {
         /** @type {HTMLButtonElement} */
         ($el.querySelector('button.submit')).classList.add('nq-button', 'inverse', options.bgColor);
 
+        if (options.hideSkip) {
+            ($el.removeChild(/** @type {HTMLAnchorElement} */($el.querySelector('.password-skip'))));
+        }
```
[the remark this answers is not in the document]
Bettelstab: Sure. I did it like this to keep the htmlString in one piece. But I don't have a strong opinion about it.

src/request/change-password/ChangePassword.js
Bettelstab: But then I have to cast it later on, which adds much more noise.

sisou: But `_key` _is_ `null` until the account is unlocked, so this is even _technically_ wrong.

Bettelstab: Agreed. Found another solution.

src/request/change-password/ChangePassword.js
Bettelstab: Maybe it got lost in the meantime, which `shouldNeverHappen` of course.

src/request/change-password/index.html
Bettelstab: Sorry, for some reason my VS Code thinks it's a good idea.
You're right, no address, no Fira. I'll remove it.

src/request/change-password/ChangePassword.js
Bettelstab: How do I do that?

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,203 @@
...
+        }
+        this._loginFileIcon.lock(colorClass);
+        this._progressIndicator.setStep(3);
+    }
+
+    /**
+     * @param {string} newPassword
+     */
+    async _finish(newPassword) {
+        const passwordBytes = Utf8Tools.stringToUtf8ByteArray(newPassword);
```
sisou: I think you accidentally lost the `TopLevelApi.setLoading(true)` call here in you last commit.

src/request/change-password/ChangePassword.js
```diff
@@ -179,28 +136,63 @@ class ChangePassword {
...
-     * @param {string} phrase
+     * @param {string} newPassword
      */
-    async _finish(phrase) {
-        TopLevelApi.setLoading(true);
-        if (!this._key) {
-            this._reject(new Errors.KeyguardError('Bypassed Password'));
-            return;
-        }
+    async _finish(newPassword) {
```
sisou: This function does not actually _finish_ the request for entropies. Those continue to the LoginFile download page. Thus `_finish` might be misleading.

src/request/change-password/ChangePassword.js
Bettelstab: Actually, it was missing before already. Finish was never called. But yeah, it's missing.

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,212 @@
...
+            const colorString = LoginFile.CONFIG[color].name;
+            colorClass = `nq-${colorString}-bg`;
+        }
+        this._loginFileIcon.lock(colorClass);
+        this._progressIndicator.setStep(3);
+    }
+
+    /**
+     * Called after new password was entered second time.
+     * @param {string} newPassword
```
nibhar: In case we want to allow for no password as new password this becomes 
```suggestion
     * @param {string} [newPassword]
```
That would also mean, that the test for 
```
if (this.key.secret instanceof Nimiq.PrivateKey) {
```
is no longer enough as unencrypted Entropies are also exempt from Login Files.

src/request/change-password/ChangePassword.js
```diff
@@ -0,0 +1,212 @@
...
+        );
+
+        const encryptedEntropy = await this.key.secret.exportEncrypted(passwordBytes);
+
+        this._downloadLoginFile.setEncryptedEntropy(
+            /** @type {Nimiq.SerialBuffer} */(encryptedEntropy),
+            firstAddress,
+        );
+
+        this._downloadLoginFile.on(DownloadLoginFile.Events.DOWNLOADED, () => {
```
nibhar: what happens if I go back and fourth between the login file and the password?

I think the event handling here should go to the constructor.

src/request/change-password/index.html
```diff
@@ -62,53 +71,79 @@
...
+                    <svg class="nq-icon"><use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-arrow-left"/></svg>
+                </a>
+                <h1 data-i18n="change-password-heading" class="nq-h1">Confirm old password</h1>
             </div>
 
             <div class="page-body nq-card-body">
-                <p data-i18n="change-paragraph" class="nq-text">
-                    Before entering your new password, you must first unlock your account.
+                <p data-i18n="change-password-paragraph" class="nq-text nq-orange center">
+                    Changing the password will create a new Login File that replaces the current one.
```
nibhar: So is there no replacement text for this in case of a `Nimiq.PrivateKey`?

src/request/change-password/index.html
```diff
@@ -62,53 +71,79 @@
...
                 </a>
-                <h1 data-i18n="change-passphrase-heading" class="nq-h1">Change your password</h1>
+                <h1 class="nq-h1" data-i18n="change-password-set-password-heading">Create a new password</h1>
             </div>
 
             <div class="page-body nq-card-body">
-                <p data-i18n="change-set-paragraph" class="nq-text">
-                    Now please enter your new password and repeat it for confirmation.
-                </p>
+                <p data-i18n='change-password-set-password-text' class="nq-light-blue nq-text center">Secure your new Login File.</p>
```
nibhar: Also no replacement text?

PR #229, opened 2019-02-23
Sebastian/export

nibhar:
This PR updates the Export flow to the new user flow designs. Also included is an update to remove key where the handlers get used.

Missing:
- [x] Focus on input boxes.
- [x] language attributes

Not included are language attributes and localisation for the Login File intro page, as they will change again. Also not included is a request option to direct the user to the recovery words immediately.


src/request/remove-key/index.html
```diff
@@ -79,58 +85,253 @@
     </div>
 
     <div id="app">
-        <!-- Remove key-->
-        <div class="page nq-card" id="remove-key">
+        <div id="rotation-container">
```
nibhar: identical to `export/index.html` except the `id="remove-key" page.

src/request/export/ExportFile.js
```diff
@@ -92,71 +153,80 @@ class ExportFile extends Nimiq.Observable {
...
+        this._key.hasPin = false;
+        this._password = password ? Utf8Tools.stringToUtf8ByteArray(password) : undefined;
+        await KeyStore.instance.put(this._key, this._password);
+
+        const keyInfo = await KeyStore.instance.getInfo(this._key.id);
+        if (!keyInfo) {
+            this._reject(new Errors.KeyNotFoundError());
+            return;
+        }
+        this._request.keyInfo = keyInfo;
```
sisou: Why get it from the database? What is different, except the `encrypted` part, which could just as well be set directly on the request.keyInfo.

src/request/export/ExportFile.js
```diff
@@ -92,71 +153,80 @@ class ExportFile extends Nimiq.Observable {
...
+                this._key.defaultAddress,
+            );
+
+            this._downloadLoginFile.on(DownloadLoginFile.Events.DOWNLOADED, () => {
+                this._resolve({ success: true });
+            });
+
+            window.location.hash = ExportFile.Pages.LOGIN_FILE_DOWNLOAD;
+        } else {
+            this._reject(new Errors.KeyguardError('key or password missing'));
```
sisou:
```suggestion
            this._reject(new Errors.KeyguardError('Key or password missing'));
```

src/request/export/ExportFile.js
```diff
@@ -92,71 +153,80 @@ class ExportFile extends Nimiq.Observable {
...
-        /** @type {HTMLElement} */
-        const $app = (document.getElementById('app'));
-        $app.insertBefore($el, $app.children[1]);
-        return $el;
+    /**
+     * Used to set the key if already decrypted elsewhere. This will disable the password requirement
+     * for use cases where the unencrypted key is needed. In case the encrypted key is needed, the password
+     * must also be provided to bypass its requirement.
+     * Set to null to reenable password requirement.
+     * @param {Key | null} key
```
sisou:
```suggestion
     * @param {Key?} key
```

src/request/export/ExportWords.js
```diff
@@ -21,94 +22,112 @@ class ExportWords extends Nimiq.Observable {
...
+
         /** @type {Key | null} */
         this._key = null;
 
+        // pages
         /** @type {HTMLElement} */
-        const $privacyPage = document.getElementById(ExportWords.Pages.PRIVACY)
-                          || this._buildPrivacy();
-
+        this._$noRecovery = (document.getElementById(ExportWords.Pages.RECOVERY_WORDS_INTRO));
```
sisou:
```suggestion
        this._$noRecoveryPage = (document.getElementById(ExportWords.Pages.RECOVERY_WORDS_INTRO));
```

src/request/export/ExportWords.js
```diff
@@ -21,94 +22,112 @@ class ExportWords extends Nimiq.Observable {
...
     }
 
     run() {
-        this._privacyWarningPassphraseBox.reset();
-        window.location.hash = ExportWords.Pages.PRIVACY;
+        this._wordsPasswordBox.reset();
+        window.location.hash = ExportWords.Pages.RECOVERY_WORDS_INTRO;
         if (TopLevelApi.getDocumentWidth() > Constants.MIN_WIDTH_FOR_AUTOFOCUS) {
-            this._privacyWarningPassphraseBox.focus();
+            this._wordsPasswordBox.focus();
```
sisou: I guess this could lead to the keyboard opening on e.g. an iPad, which fullfills the min-width condition, when the input is however hidden because it's unencrypted or the key is already known. So I think this focus should be only executed when we know we need a password.

Or, better: The PassphraseBox should only forward the focus() to the input element, when its `options.hideInput` is false.

src/request/export/ExportWords.js
```diff
@@ -21,94 +22,112 @@ class ExportWords extends Nimiq.Observable {
...
+        } else if (password) {
+            passwordBuffer = Utf8Tools.stringToUtf8ByteArray(password);
+        }
+
         /** @type {Key?} */
         let key = null;
+
         try {
-            key = await KeyStore.instance.get(this._request.keyInfo.id, passphraseBuffer);
+            key = await KeyStore.instance.get(this._request.keyInfo.id, passwordBuffer);
```
sisou: If we already know the `_password`, then we most likely also already know the `this._key`. I think getting it from the database and running the decryption _again_ is unecessary when `this._key` is already available.

src/request/export/ExportWords.js
```diff
@@ -128,105 +148,23 @@ class ExportWords extends Nimiq.Observable {
                 words = Nimiq.MnemonicUtils.entropyToMnemonic(this._key.secret);
             } else {
                 this._reject(new Errors.KeyguardError('Unknown mnemonic type'));
```
sisou:
```suggestion
                 this._reject(new Errors.KeyguardError('Unknown secret type'));
```

src/request/export/ExportWords.js
```diff
@@ -21,94 +22,112 @@ class ExportWords extends Nimiq.Observable {
...
             return;
         }
 
-        this.setKey(key);
-        this.fire(ExportWords.Events.KEY_CHANGED, {
-            key,
-            isProtected: this._request.keyInfo.encrypted,
-        });
+        this.setKey(key, passwordBuffer);
+        this.fire(ExportWords.Events.KEY_CHANGED, key, passwordBuffer);
```
sisou: Is it really necessary to fire both the key and the password? I think the key should suffice. At least from ExportFile to ExportWords, only the key is enough. When ExportFile then uses the database-stored encrypted secret, instead of decrypting and encrypting, the key (and password) would not have to be fired from ExportWords.

src/request/export/ExportWords.js
```diff
@@ -117,8 +136,9 @@ class ExportWords extends Nimiq.Observable {
      * Used to set the key if already decrypted elsewhere. This will disable the passphrase requirement.
      * Set to null to re-enable passphrase requirement.
      * @param {Key | null} key
```
sisou:
```suggestion
      * @param {Key?} key
```

src/request/export/ExportWords.js
```diff
@@ -128,105 +148,23 @@ class ExportWords extends Nimiq.Observable {
...
-
-        <div class="page-body nq-card-body">
-            <div class="validate-words"></div>
-        </div>
-        `;
-        /** @type {HTMLElement} */
-        const $app = (document.getElementById('app'));
-        $app.insertBefore($el, $app.children[1]);
-        return $el;
+        this._$noRecovery.classList.toggle('key-active', this._key !== null);
```
sisou: What does `key-active` stand for and do? The classname is not descriptive enough.

src/common.css
```diff
@@ -176,10 +176,6 @@
     margin-right: 0.25rem;
 }
 
-.flex-grow {
-    flex-grow: 1;
-}
```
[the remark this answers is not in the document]
nibhar: https://github.com/nimiq/nimiq-style/blob/0ebfe54d8c67f6cff50093e03097f3329922b549/src/layout.css#L15
Is a part of nimiq-style, so nothing was lost here except for 4 lines.

src/common.css
```diff
@@ -32,7 +32,7 @@
     width: 100%;
     max-width: 52.5rem; /* 420px */
     min-height: 70.5rem; /* 564px */
-    /**
+     /**
```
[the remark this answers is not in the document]
nibhar: I removed the comment at some point. Re-added it later and apparently also an additional space.

src/components/ProgressIndicator.css
```diff
@@ -13,6 +16,10 @@
     margin: 0.5rem;
 }
 
+.nq-card-header .progress-indicator ~ * {
+    margin-top: 1rem;
+}
```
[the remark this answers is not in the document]
nibhar: In the presence of a ProgressIndicator within a `.nq-card-header`, all neighbours of said indicator get an additional margin. I think this is the desired behaviour?

src/components/RecoveryWords.js
```diff
@@ -68,9 +68,14 @@ class RecoveryWords extends Nimiq.Observable {
                 content.title = `word #${i + 1}`;
                 this.$fields.push(content);
 
+                const $wordNumber = document.createElement('span');
+                $wordNumber.classList.add('word-number');
+                $wordNumber.innerText = `${i <= 8 ? '0' : ''}${i + 1}`;
```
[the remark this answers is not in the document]
nibhar: It is either some `i + 1` or some `i - 1`. I think starting at 0 then makes more sense.

src/request/export/Export.css
```diff
@@ -1,50 +1,118 @@
...
+    display: flex;
+    flex-direction: column;
+    flex-grow: 1;
+    padding-bottom: 1rem;
 }
 
-.page#download-key-file .download-icon {
-    align-self: center;
+.page#recovery-words {
+    height: 70.5rem; /* 564px */
```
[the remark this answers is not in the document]
nibhar: I think we should do it in another PR as it will likely touch all the other requests, too.

src/request/export/Export.js
```diff
@@ -21,95 +22,38 @@ class Export {
...
+            this._exportFileHandler.run();
         } else {
-            this.exported.file = result.success;
-            this.$fileButton.classList.add('display-none');
-            window.location.hash = Export.Pages.MORE_EXPORT_OPTIONS;
+            this._exportWordsHandler.run();
         }
     }
 
+    _fileExportSuccessful() {
```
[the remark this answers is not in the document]
nibhar: Because the export request can not be resolved without the file export being resolved. It can resolve without the words portion of it resolving.

src/request/export/ExportFile.js
```diff
@@ -19,69 +24,135 @@ class ExportFile extends Nimiq.Observable {
...
+        new ProgressIndicator($exportFileIntroPage.querySelector('.progress-indicator'), 3, 1);
+        new ProgressIndicator($unlockFilePage.querySelector('.progress-indicator'), 3, 2);
+        new ProgressIndicator($setPasswordPage.querySelector('.progress-indicator'), 3, 2);
+        new ProgressIndicator($downloadFilePage.querySelector('.progress-indicator'), 3, 3);
+        /* eslint-enable no-new */
+
+        $fileButton.addEventListener('click', async () => {
+            if (this._request.keyInfo.encrypted) {
+                if (this._key && this._password) {
+                    await this._passwordSubmitted('');
```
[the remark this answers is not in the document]
nibhar: removed the side effect from passwordSubmitted to be more explicit about it.

src/request/export/ExportFile.js
nibhar: `this._request.keyInfo.encrypted` is readonly. Added a setter.

src/request/export/ExportWords.js
nibhar: The words do not need the password. The file does as it stands. I looked at Import words code for it where it still has the comment about having the key encrypted twice. I did not change this behaviour in this PR, but we can look into it. I wanted both events to fire the same parameters, but I can change it.

src/components/ProgressIndicator.css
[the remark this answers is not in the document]
sisou: Ok.

src/request/export/Export.js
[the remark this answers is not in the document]
nibhar: added parameter to it. Also added a button to export file's intro page to go to words directly.

src/request/export/ExportFile.js
```diff
@@ -95,7 +97,11 @@ class ExportFile extends Nimiq.Observable {
         });
 
         this._passwordBox.on(PassphraseBox.Events.SUBMIT, async password => {
-            await this._passwordSubmitted(password);
+            if (this._key && this._password) {
+                await this._passwordSubmitted(this._password);
+            } else {
+                await this._passwordSubmitted(password);
```
sisou: Does this make sense to separate those now? If the passphrase is entered, why would you use another password then? Even when the password is right there as a parameter?

nibhar: Because it is not. The PassphraseBox can `.hideInput()`, but the button then still fires `PassphraseBox.Events.Submit`. However, there is no input.

sisou: Ah. So does it maybe make sense to do the check the other way around (check if `password` and only then do the check for _key and _password, including a comment of when this situation might happen?

nibhar: `undefined` is in general a valid password. So doing it the other way around is not a good idea I think, even though it would work in this case.
I will add a comment about it.

src/request/export/ExportWords.js
```diff
@@ -21,104 +22,123 @@ class ExportWords extends Nimiq.Observable {
...
+            }
+            this.fire(ExportWords.Events.KEY_CHANGED, key, password);
         }
 
-        this.setKey(key);
-        this.fire(ExportWords.Events.KEY_CHANGED, {
-            key,
-            isProtected: this._request.keyInfo.encrypted,
-        });
+        this.setKey(key, password);
```
Bettelstab: For the `if (this._key)` case, this is redundant, isn't it?

src/translations/index.js
```diff
@@ -239,12 +269,40 @@ const TRANSLATIONS = {
         'validate-words-text': 'Bitte wähle das richtige Wort aus deiner Liste von Wiederherstellungswörtern aus.',
         'validate-words-back': 'Zurück zu den Wörtern',
         'validate-words-skip': 'Überprüfung erstmal überspringen',
+        'validate-words-1-hint': 'Was ist das 1te Wort?',
```
Bettelstab: 1., 2. etc.

src/translations/index.js
```diff
@@ -239,12 +269,40 @@ const TRANSLATIONS = {
...
+        'validate-words-19-hint': 'Was ist das 19te Wort?',
+        'validate-words-20-hint': 'Was ist das 20te Wort?',
+        'validate-words-21-hint': 'Was ist das 21te Wort?',
+        'validate-words-22-hint': 'Was ist das 22te Wort?',
+        'validate-words-23-hint': 'Was ist das 23te Wort?',
+        'validate-words-24-hint': 'Was ist das 34te Wort?',
 
         'export-file-heading': 'Schlüsseldatei herunterladen',
+        'export-file-intro-heading': 'Sichere deinen Account',
+        'export-words-intro-heading': 'Es gibt keine Passwort Wiederherstellung',
```
Bettelstab: Passwortwiederherstellung :p

src/translations/index.js
```diff
@@ -239,12 +269,40 @@ const TRANSLATIONS = {
...
         'export-button-words': 'Wiederherstellungswörter anzeigen',
         'export-button-file': 'Schlüsseldatei herunterladen',
         'export-finish': 'Export abschließen',
         'export-more-options-heading': 'Zusätzliche Export Optionen',
```
Bettelstab: Exportoptionen :)

src/request/export/ExportWords.js
nibhar: It is. Moved it into the else block of said conditional.

src/translations/index.js
nibhar: no longer used, so I removed it.

nibhar: I think we need to check on the translation file in terms of which strings are even still in use.

src/request/export/ExportFile.js
danimoh: Apparently no comment was added :laughing:

nibhar: Apparently I forgot about it. Will add it in the current feedback branch.

PR #230, opened 2019-02-23
Switch QR Scanner

danimoh:
- change QR scanner library in QrScannerWorker.js to jsQR
- remove unused code from QrScanner.js and refactor a little bit
- change file importer to use object urls instead of FileLoader
- reuse rendered image in file importer for QR scanning
- preload qr scanner worker in import file request

Note that I have prove read the QrScanner library but feel free to also do so.


src/components/FileImporter.js
```diff
@@ -71,45 +71,41 @@ class FileImporter extends Nimiq.Observable {
...
+            AnimationUtils.animate('shake', this.$el);
+            this.$errorMessage.textContent = 'Could not read Login File.';
+            return;
+        }
+
+        if (this._displayFile) {
+            // Show image in UI
+            this.$el.appendChild(image);
+            image.classList.add('pop-down');
+        }
```
sisou: Just a note about the new CSS classes behavior (this is not relevant in the way the importer is used currently, as `displayFile` is set to false, so the image is never displayed, but still):

When imorting a second image, while the first one is still displayed in the page (when `displayFile` is set to `true`), the second image is loaded into the page (`image.src = url`) first, and _then_ decoded by the qr-reader. Depending on how long this takes, the image is displayed - with full opacity and non-blurred - while the qr-reader is working. Only when the qr-reader returns, is the image animated with the pop-down animation (although it was already displayed in the animation's end state).

Before, the opacity and scale was at least set to the initial animation values while the qr-ready was working.

I fear that these changes will lead to UI-jumping from end state, to animation, to end state.

src/components/FileImporter.css
```diff
@@ -55,14 +55,19 @@
     left: -1rem;
     width: calc(100% + 2rem);
     height: auto;
-    transform: scale(1.05);
-    opacity: 0.2;
-    transition: transform .3s, opacity .3s;
+    display: none;
```
Bettelstab: Finally you're listening to me 👍

src/components/FileImporter.js
danimoh: This should not happen as the image is set to `display: none` if it doesn't have the `pop-down` class.

sisou: Oh, you're right. I must have overlooked that. Disregard my comment :D

danimoh:
Note: The contrast for yellow and turquoise login files is too low for the used grayscale weights and thus the scanner is having issues detecting the QR code for those colors.
I'll implement one of the following approaches to fix that:
- Search for a set of greyscale weights that work for all color versions.
- Use two different sets of grayscale weights and check both.
- Add a bias towards black in the binarizer.

danimoh:
It should work for all colors now.

PR #232, opened 2019-02-26
Type all resolve results

### LINKED ISSUE #227: Type resolve in all handlers with their expected result type

sisou:
This is a **Draft PR**, because we are waiting for #229.

Types for the `change-passphrase` request are already included in #226.

Resolves #227.

### DIFF
--- a/src/request/sign-message/SignMessage.js
+++ b/src/request/sign-message/SignMessage.js
@@ -8,11 +8,16 @@
 /* global Errors */
 /* global TopLevelApi */
 
+/**
+ * @callback SignMessage.resolve
+ * @param {KeyguardRequest.SignMessageResult} result
+ */
+
 class SignMessage {
     /**
      * @param {Parsed<KeyguardRequest.SignMessageRequest>} request
-     * @param {Function} resolve
-     * @param {Function} reject
+     * @param {SignMessage.resolve} resolve
+     * @param {reject} reject
      */
     constructor(request, resolve, reject) {
         /** @type {HTMLDivElement} */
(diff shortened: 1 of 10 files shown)


Bettelstab:
Don't forget to add the same for sign-message after #41 is merged.

sisou:
SignMessage added, now only waiting for #229.

src/request/sign-transaction/SignTransaction.js
```diff
@@ -148,7 +153,9 @@ class SignTransaction {
 
         const publicKey = key.derivePublicKey(request.keyPath);
         const signature = key.sign(request.keyPath, request.transaction.serializeContent());
-        const result = /** @type {SignTransactionResult} */ {
+
+        /** @type {KeyguardRequest.SignTransactionResult} */
```
Bettelstab: And that cast?

src/request/sign-message/SignMessage.js
```diff
@@ -98,11 +103,12 @@ class SignMessage {
 
         const signingResult = key.signMessage(request.keyPath, request.message);
 
-        const result = /** @type {KeyguardRequest.SignMessageResult} */ ({
+        /** @type {KeyguardRequest.SignMessageResult} */
```
Bettelstab: Do we need this cast?

src/request/sign-transaction/SignTransaction.js
sisou: That cast is not _necessary_ anymore, but it helps in identifying a missing property (because the error is then on the object, and not on the `resolve()` call.

src/request/sign-message/SignMessage.js
sisou: For DevX as described above.

src/request/sign-transaction/SignTransaction.js
Bettelstab: Ok

PR #236, opened 2019-02-27
ExportRequest and Result types

nibhar:
Also includes disable password skip for exportFile

### DIFF
--- a/src/request/export/Export.js
+++ b/src/request/export/Export.js
@@ -36,10 +36,10 @@ class Export { // eslint-disable-line no-unused-vars
     }
 
     run() {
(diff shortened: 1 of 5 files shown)


src/request/export/Export.js
```diff
@@ -36,7 +36,7 @@ class Export { // eslint-disable-line no-unused-vars
     }
 
     run() {
-        if (this._request.keyInfo.type === Nimiq.Secret.Type.ENTROPY) {
+        if (!this._request.skipFile || this._request.keyInfo.type === Nimiq.Secret.Type.ENTROPY) {
```
danimoh: Should be `&&`?

PR #243, opened 2019-03-12
Add core-web dependency in client

Bettelstab:
We need to add @nimiq/core-web as dependency (for typing only). I noticed this when I tried to replace network-client by nano-api in accounts, which removed the former present @nimiq/core-web dependency there.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nimiq/keyguard-client",
-  "version": "0.2.3",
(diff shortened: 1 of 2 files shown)


sisou:
Why did you need to increase the version number again in a separate commit?

Bettelstab:
> Why did you need to increase the version number again in a separate commit?

For the statistics, obviously.
No, I just forgot it before, but thought it would be nice to have the right version number committed, with the package available on NPM.
Squashed it now.

PR #244, opened 2019-03-12
Client: Update Typescript and fix state type

Bettelstab:
In Client:
- Update typescript and resolve issues
- Type state as ObjectType|null, like in update of https://github.com/nimiq/rpc


client/src/KeyguardClient.ts
```diff
@@ -206,25 +207,34 @@ export class KeyguardClient {
         return publicResult;
     }
 
-    private _onReject(error: any, id: number, state: any) {
-        const command = state.__command;
-        if (!command) {
-            throw new Error('Invalid state after RPC request');
+    private parseState(state?: ObjectType|null) {
```
danimoh: should be called `_parseState` as it's private

client/src/KeyguardClient.ts
```diff
@@ -206,25 +207,34 @@ export class KeyguardClient {
...
+        if (state) {
+            const command = state.__command;
+            if (command) {
+                delete state.__command;
```
danimoh: Side effect ok?

client/src/Observable.ts
```diff
@@ -65,7 +65,11 @@ export default class Observable {
             let callback;
             if (type === Observable.WILDCARD) {
                 callback = function(this: Observable) {
-                    this.fire.apply(this, arguments);
+                    const args = arguments;
+                    if (typeof args === 'object' && args && typeof args[0] === 'string') {
+                        const argsArray = [ ...arguments ];
+                        this.fire.apply(this, [argsArray[0], ...argsArray.slice(1)]);
+                    }
```
danimoh: Would something like `this.fire.apply(this, arguments as [string, ...any]);` have been possible?
(I don't know whether typescript even supports something like that though).

Otherwise `this.fire.apply(this, [arguments[0] as string, ...[...arguments].slice(1)]);`?

sisou:
```
this.fire.apply(this, [...arguments] as [string, ...any[]]);
```
works.

client/src/KeyguardClient.ts
sisou: Was done the same way before also: https://github.com/nimiq/keyguard-next/pull/244/files#diff-f7e2d477e26eefd5332b8236f5588184L214

PR #252, opened 2019-04-01
add TemplateTags and tag many uses of template strings with number of…

Bettelstab:
Adds template tags to template strings indicating how many variables were used.
Usage: Either e.g.
```
TemplateTags.hasVars(1)`${'foo'}bar`
```
or
```
TemplateTags.noVars`foobar`
```
If tag and actual number of variables don't match, both typescript compiler and the `hasVars` method (at runtime) will reject.
I didn't add those tags to obvious one-liners.

Tests failing with ```"An error was thrown in afterAll\nUncaught ReferenceError: TemplateTags is not defined"```, I haven't found out yet where that comes from.

### DIFF
--- a/src/components/AddressInfo.js
+++ b/src/components/AddressInfo.js
@@ -71,7 +71,7 @@ class AddressInfo {
 }
 
 /* eslint-disable max-len */
-AddressInfo.HEXAGON_CLIP_PATH = TemplateTags.noVariables`
+AddressInfo.HEXAGON_CLIP_PATH = TemplateTags.noVars`
     <svg width="0" height="0" viewBox="0 0 146 146">
         <defs>
             <clipPath id="nimiq-rounded-hexagon" clipPathUnits="objectBoundingBox">
--- a/src/components/AddressInfo.js
+++ b/src/components/AddressInfo.js
@@ -1,4 +1,5 @@
 /* global Identicon */
+/* global TemplateTags */
 
 class AddressInfo {
     /**
@@ -70,7 +71,7 @@ class AddressInfo {
 }
 
 /* eslint-disable max-len */
-AddressInfo.HEXAGON_CLIP_PATH = `
+AddressInfo.HEXAGON_CLIP_PATH = TemplateTags.noVariables`
     <svg width="0" height="0" viewBox="0 0 146 146">
         <defs>
             <clipPath id="nimiq-rounded-hexagon" clipPathUnits="objectBoundingBox">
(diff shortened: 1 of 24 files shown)


sisou:
I'd propose to shorten the tag name to `TemplateTags.hasVars`.

sisou:
`noVars` could also be called `clean`.

Bettelstab:
> I'd propose to shorten the tag name to `TemplateTags.hasVars()`/`TemplateTags.noVars`.

Yep, we could do that.

> `noVars` could also be called `clean`.

I think `noVars` makes it clearer.

sisou:
I think we don't need sanitization in template tags. Looking at the occurances, there's is *no* relevant location outside the iqons that uses an external variable.

For e.g. https://github.com/nimiq/keyguard-next/pull/252/files#diff-a05500321c657b0aa2840c803f76fe09R28 I think putting a `parseInt()` around the number should be fine.

Only issue is thus with the Iqons. But there the only check is to see if it's valid XML anyway, which can be done, but not in template tags.

Bettelstab:
Let's not overdo it, please :)

danimoh:
But then we don't really gain much additional security from `hasVars(x)` if still anything can be injected.
Note that stuff like https://github.com/nimiq/keyguard-next/pull/252/files#diff-80c8a7449a4c2be32d6b326fdaaab39eR80 can also benefit and could be rewritten as:
```js
$el.innerHTML = TemplateSanitization('no-html', 'no-html') `
    <a class="cancel">
        <svg class="nq-icon">
            <use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-close"/>
        </svg>
    </a>
    <div class="prompt nq-text-s" data-i18n="passwordbox-enter-password">Enter your password</div>
    <div password-input></div>
    <button class="submit" data-i18n="${options.buttonI18nTag}">${I18n.translatePhrase(options.buttonI18nTag)}</button>
`;
```

sisou:
The reason for the button tags to be as they are is so the automated language checker tool can find the tags and their translations in the DOM.

danimoh:
@sisou but why wouldn't that work anymore?

sisou:
Because this
```js
`<button class="submit" data-i18n="${options.buttonI18nTag}">${I18n.translatePhrase(options.buttonI18nTag)}</button>`
```
does not fulfill this regexp
```js
const textContentRegEx = /data-i18n="(.*?)".*?>\s*(.*?)\s*</gs;
```
(https://github.com/nimiq/keyguard-next/blob/master/tools/translationValidator.js#L18)

PR #253, opened 2019-04-02
Add devnet config option, use constant for CONFIG.NETWORK

sisou:


sisou:
It actually doesn't work... :(
Somehow I cannot use the constant in the config object creation. See failing tests.

Bettelstab:
In karma.conf.js add before line 21: ```'src/lib/Constants.js'```

PR #255, opened 2019-04-06
Use full key hash as key id

sisou:
As discussed, we use the full 32-byte key hash as it's ID.

@Bettelstab, are there other simplifications we can do again, which you may have removed when changing it to auto-increment?


tests/lib/IframeApi.spec.js
```diff
@@ -62,7 +62,7 @@ describe('IframeApi', () => {
         expect(CookieJar.eat).not.toHaveBeenCalled();
         expect(AccountStore.instance.list).not.toHaveBeenCalled();
         expect(KeyStore.instance.list).toHaveBeenCalled();
-        expect(listedKeyObjects).toEqual(Dummy.keyInfoObjects());
+        expect(listedKeyObjects).toEqual(Dummy.keyInfoObjects().reverse());
```
Bettelstab: why `reverse()` ?

src/lib/Key.js
```diff
@@ -7,18 +7,18 @@ class Key {
...
      * @param {boolean} [hasPin]
-     * @param {number?} [id]
      */
-    constructor(secret, hasPin = false, id = null) {
+    constructor(secret, hasPin = false) {
         this._secret = secret;
         this._hasPin = hasPin;
-        this._id = id;
+        /** @type {string?} */
+        this._id = null;
```
Bettelstab: why not just set the hash right here? performance?

tests/lib/IframeApi.spec.js
Bettelstab: Ok, both options are fine, but a comment would be nice here :)

sisou: I have reversed the DUMMY data objects in a separate branch https://github.com/nimiq/keyguard-next/commit/bffceb3dc3a17e4a0581595fcaf94862ce0e601a to not clutter this PR.

My suggestion is to create a separate PR for that after this has been merged. Do you agree, or should I add it to this PR?

Bettelstab: Agree :)

client/src/KeyguardClient.ts
```diff
@@ -190,18 +152,16 @@ export class KeyguardClient {
...
     }
 
     private async _iframeRequest<T1 extends IFrameRequest, T2 extends IFrameResult>(
         command: KeyguardCommand,
         request?: T1,
     ): Promise<T2> {
-        const args = request ? [ KeyguardClient.publicToInternal(request) ] : [];
-        const internalResult = await this._iframeBehavior.request(this._endpoint, command, args);
-        return KeyguardClient.internalToPublic(internalResult) as T2;
+        const args = request ? [request] : [];
```
nibhar: could be inlined I think. Depeding on what `[ undefined ]` returns you could even just put `[ request ]` in the next line.

client/src/KeyguardClient.ts
```diff
@@ -190,18 +152,16 @@ export class KeyguardClient {
...
     ) {
-        const internalRequest = KeyguardClient.publicToInternal(request);
-        this._redirectBehavior.request(this._endpoint, command, [ internalRequest ]);
+        this._redirectBehavior.request(this._endpoint, command, [ request ]);
         // return value of redirect call is received in _onResolve()
     }
 
     private async _iframeRequest<T1 extends IFrameRequest, T2 extends IFrameResult>(
         command: KeyguardCommand,
         request?: T1,
```
nibhar: Unrelated, but I just noticed that most of these fit in 1 line.

PR #256, opened 2019-04-08
Rework message signing

sisou:
- Prohibit control characters (except tabs and line breaks) in signMessage messages
- Do not return signed data from signMessage
- Use Ethereum model for message signing (length bytes, hashing before signing)
- Export message prefix in client

To use with https://github.com/nimiq/accounts/pull/199.


sisou:
The new test for `RequestParser` uses `expect(...).toThrow()`.

But I'm sill waiting for Philipp's exploit data.

Bettelstab:
> The new test for `RequestParser` uses `expect(...).toThrow()`.
> 
> But I'm also still waiting for Philipp's exploit data.

Ah... cool.

src/lib/Utf8Tools.js
```diff
@@ -94,24 +94,25 @@ class Utf8Tools { // eslint-disable-line no-unused-vars
...
-            if (first <= 0x7F) ++i; // Is valid single-byte (ASCII)
+            const controlCharsWhitelist = [
+                0x09, /* horizontal tab (\t) */
+                0x0A, /* line feed (\n) */
+                0x0D, /* carriage return (\r) */
+            ];
+
+            if (first <= 0x7F) { // Possible one-byte
+                if (controlCharsWhitelist.indexOf(first) > -1) ++i;
+                else if (first >= 0x20 /* space */ && first <= 0x7E /* tilde */) ++i; // Only allow non-control chars
```
nibhar: Are we sure that there are no other 'bad' characters in the multibyte range? Have they already been filtered out by the previous implementation?

src/request/remove-key/index.html
```diff
@@ -22,6 +22,7 @@
     <script defer bundle-common src="../../lib/KeyInfo.js"></script>
     <script defer bundle-common src="../../lib/Errors.js"></script>
     <script defer bundle-common src="../../lib/ErrorConstants.js"></script>
+    <script defer bundle-common src="../../lib/SignMessageConstants.js"></script>
```
nibhar: why is this needed for this request, and the others. I only comment here, but question is for all requests except for `sign-message`.

src/lib/RequestParser.js
```diff
@@ -169,11 +169,15 @@ class RequestParser { // eslint-disable-line no-unused-vars
 
     /**
      * @param {any} message
-     * @returns {Uint8Array}
+     * @returns {string | Uint8Array}
```
nibhar: I see that it is displayed in hex in Uint8Array case and as string otherwise. Is that a good enough criteria for the determination of whether it is data or displayable text?

tests/lib/RequestParser.spec.js
```diff
@@ -146,7 +146,8 @@ describe('RequestParser', () => {
         expect(() => requestParser.parseMessage(undefined)).toThrow();
         expect(() => requestParser.parseMessage(5)).toThrow();
         expect(() => requestParser.parseMessage({})).toThrow();
-        expect(requestParser.parseMessage('123456')).toEqual(new Uint8Array([49, 50, 51, 52, 53, 54]));
+        expect(() => requestParser.parseMessage(Utf8Tools.utf8ByteArrayToString(new Uint8Array([72, 101, 108, 108, 111, 0, 32, 87, 111, 114, 108, 100, 33])))).toThrow(); // 'Hello World!' with a zero byte in the middle
```
nibhar:
```suggestion
        expect(() => requestParser.parseMessage(Utf8Tools.utf8ByteArrayToString(new Uint8Array([72, 101, 108, 108, 111, 0, 32, 87, 111, 114, 108, 100, 33])))).toThrow(); // 'Hello World!' with a zero byte in the middle as string
        expect(() => requestParser.parseMessage(new Uint8Array([72, 101, 108, 108, 111, 0, 32, 87, 111, 114, 108, 100, 33]))).toThrow(); // 'Hello World!' with a zero byte in the middle as Uint8Array
```

src/request/remove-key/index.html
nibhar: ah nevermind. Key.js needs it so yes, it is needed.

PR #259, opened 2019-04-09
JS config files with safe defaults

### LINKED ISSUE #258: Safe default config

Bettelstab:
Use .js files for configuration and prevent development config from being used in production. I removed the CDN setting. Resolves #258  and #257


tools/build.sh
```diff
@@ -97,16 +85,15 @@ for DIR in src/request/*/ ; do
...
+    LIST_JS="$(grep '<script' $DIR/index.html | grep -v 'bundle-' | grep -v 'config' | grep -v -E 'web-offline.js' | cut -d\" -f2)"
 
     # concat them
     for url in $LIST_JS; do
         cat $DIR/$url >> dist/request/$REQUEST/$JS_BUNDLE
     done
 
-    replace_config_variable "CONFIG.ALLOWED_ORIGIN" "KEYGUARD_ALLOWED_ORIGIN" dist/request/$REQUEST/$JS_BUNDLE
-    replace_config_variable "CONFIG.NETWORK" "KEYGUARD_NETWORK" dist/request/$REQUEST/$JS_BUNDLE
+    replace_core_lib_url dist/request/${REQUEST}/$JS_BUNDLE
```
sisou: The core-lib (previous CDN config) does not occur in JS files, only in HTML. This line should be removed.

tools/build.sh
```diff
@@ -173,21 +160,21 @@ LIST_CSS_TOPLEVEL="../../../node_modules/@nimiq/style/nimiq-style.min.css ../../
...
+# put config first
+cat src/config/config.$BUILD.js >> dist/request/$JS_COMMON_BUNDLE
 # (since all urls are relative to request directories, we simply use the create request directory as the base)
 for url in $LIST_JS_COMMON; do
     cat src/request/create/$url >> dist/request/$JS_COMMON_BUNDLE
 done
 
-replace_config_variable "CONFIG.ALLOWED_ORIGIN" "KEYGUARD_ALLOWED_ORIGIN" dist/request/$JS_COMMON_BUNDLE
-replace_config_variable "CONFIG.NETWORK" "KEYGUARD_NETWORK" dist/request/$JS_COMMON_BUNDLE
+replace_core_lib_url dist/request/$JS_COMMON_BUNDLE
```
sisou: Same here. CDN does not appear in JS files, please remove.

tools/build.sh
```diff
@@ -173,21 +160,21 @@ LIST_CSS_TOPLEVEL="../../../node_modules/@nimiq/style/nimiq-style.min.css ../../
...
 replace_icon_sprite_url dist/request/$JS_COMMON_BUNDLE
 
 for url in $LIST_JS_TOPLEVEL; do
     cat src/request/create/$url >> dist/request/$JS_TOPLEVEL_BUNDLE
 done
 
-replace_config_variable "CONFIG.ALLOWED_ORIGIN" "KEYGUARD_ALLOWED_ORIGIN" dist/request/$JS_TOPLEVEL_BUNDLE
-replace_config_variable "CONFIG.NETWORK" "KEYGUARD_NETWORK" dist/request/$JS_TOPLEVEL_BUNDLE
+replace_core_lib_url dist/request/$JS_TOPLEVEL_BUNDLE
```
sisou: And here.

tools/functions.js
```diff
@@ -88,7 +88,7 @@ function findDependencies(startFile, class2Path, deps) {
 
         deps.push(dep);
 
-        if (dep === 'Nimiq') return;
+        if (dep === 'Nimiq' || dep === 'CONFIG') return;
```
sisou: Why is this necessary? the CONFIG uses the `Constants` global, so I think it's fair to continue building the dependency graph for config as well.

tools/build.sh
```diff
@@ -173,21 +160,21 @@ LIST_CSS_TOPLEVEL="../../../node_modules/@nimiq/style/nimiq-style.min.css ../../
 
 # generate bundle files
 output "📦  Generating bundle files"
+# put config first
+cat src/config/config.$BUILD.js >> dist/request/$JS_COMMON_BUNDLE
```
sisou: The config depends on a constant from `Constants.js`. Have you tested, if this throws a `Constants is undefined` error for the built files?

Bettelstab: > It only resolves #258 to a degree, as the default config for local is still `'*'`. I am aware that it's necessary for IP-based testing, but it does not _really_ fix the issue.

I don't think the issue was local config. Might be problematic for an offline build, but when we add an option for that we might as well restrict the default local config to be not compatible with that.

Bettelstab:
> It only resolves #258 to a degree, as the default config for local is still `'*'`. I am aware that it's necessary for IP-based testing, but it does not _really_ fix the issue.

I don't think the issue was local config. Might be problematic for an offline build, but when we add an option for that we might as well restrict the default local config to be not compatible with that.

tools/build.sh
Bettelstab: ...I tested first and replaced the values by constants later.
Ok. this needs a fix.

PR #260, opened 2019-04-10
Mini fixes

sisou:
- Update favicon and link to it from requests' index.html
- Update wording according to feedback document


sisou:
Force-push was due to a copy-paste error adding `favicon.png` instead of `.ico`. Ups.

PR #262, opened 2019-04-11
add length checks for transaction data and labels

nibhar:
Also increases wait time for open db connections in tests in an attempt to have travis fail less often.

### DIFF
--- a/src/lib/RequestParser.js
+++ b/src/lib/RequestParser.js
@@ -70,7 +70,7 @@ class RequestParser { // eslint-disable-line no-unused-vars
             if (!allowEmpty) throw new Errors.InvalidRequestError('Label must not be empty');
             return undefined;
         }
-        if (Utf8Tools.stringToUtf8ByteArray(label).length > 63) {
+        if (Utf8Tools.stringToUtf8ByteArray(label).byteLength > 63) {
             throw new Errors.InvalidRequestError('Label must not exceed 63 bytes');
         }
         return label;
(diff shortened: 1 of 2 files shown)


src/lib/RequestParser.js
```diff
@@ -139,7 +142,21 @@ class RequestParser { // eslint-disable-line no-unused-vars
...
+        if (flags === Nimiq.Transaction.Flag.NONE && data.length > 64) {
+            throw new Errors.InvalidRequestError('Data must not exceed 64 bytes');
+        }
+        if (flags === Nimiq.Transaction.Flag.CONTRACT_CREATION
+                && data.length !== 78 // HTLC
+                && data.length !== 24 // Vesting
+                && data.length !== 36 // Vesting
+                && data.length !== 44) { // Vesting
+            throw new Errors.InvalidRequestError(
+                'Contract creation data must equal 78 bytes for HTLC and 24, 36, or 44 bytes for vesting contracts',
```
sisou: I think this error message can be simplified to `Invalid data length`.

src/lib/RequestParser.js
```diff
@@ -70,6 +70,9 @@ class RequestParser { // eslint-disable-line no-unused-vars
             if (!allowEmpty) throw new Errors.InvalidRequestError('Label must not be empty');
             return undefined;
         }
+        if (label.length > 63) {
```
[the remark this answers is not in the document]
nibhar: You are absolutely right

src/lib/RequestParser.js
nibhar: It looses information if I do. I don't think simplification is necessarily good, if it is acieved by removing useful information.

sisou: Ok, but then it is still _technically_ incorrectd, because you write that the _data_ must equal that, but what you mean is the data _length_.

nibhar: I'd add 'length' then, rather than shortening it.

sisou:
```suggestion
                'Contract creation data must be 78 bytes for HTLC and 24, 36, or 44 bytes for vesting contracts',
```

src/lib/RequestParser.js
```diff
@@ -70,6 +70,9 @@ class RequestParser { // eslint-disable-line no-unused-vars
             if (!allowEmpty) throw new Errors.InvalidRequestError('Label must not be empty');
             return undefined;
         }
+        if (Utf8Tools.stringToUtf8ByteArray(label).length > 63) {
```
sisou:
```suggestion
        if (Utf8Tools.stringToUtf8ByteArray(label).byteLength > 63) {
```

nibhar: changed it in `parseTransaction`, too. Is there a real difference, though?

src/lib/RequestParser.js
```diff
@@ -139,7 +142,24 @@ class RequestParser { // eslint-disable-line no-unused-vars
         }
 
         const flags = object.flags || Nimiq.Transaction.Flag.NONE;
-        const data = object.data || new Uint8Array(0);
+
+        let data = new Uint8Array(0);
+        if (object.data) {
+            data = typeof object.data === 'string' ? Utf8Tools.stringToUtf8ByteArray(object.data) : object.data;
+        }
```
sisou: This block can be 'simplified' to

```js
const data = typeof object.data === 'string'
    ? Utf8Tools.stringToUtf8ByteArray(object.data)
    : object.data || new Uint8Array(0);
```

src/lib/RequestParser.js
```diff
@@ -139,7 +142,24 @@ class RequestParser { // eslint-disable-line no-unused-vars
...
+        if (flags === Nimiq.Transaction.Flag.NONE && data.byteLength > 64) {
+            throw new Errors.InvalidRequestError('Data must not exceed 64 bytes');
+        }
+        if (flags === Nimiq.Transaction.Flag.CONTRACT_CREATION
+                && data.byteLength !== 78 // HTLC
+                && data.byteLength !== 24 // Vesting
+                && data.byteLength !== 36 // Vesting
+                && data.byteLength !== 44) { // Vesting
+            throw new Errors.InvalidRequestError(
+                'Contract creation data must be 78 bytes for HTLC and 24, 36, or 44 bytes for vesting contracts',
```
Bettelstab: "bytes long", but yeah ^^

PR #263, opened 2019-04-11
export and remove-key update

nibhar:
This PR updates the export and remove-key requests according to feedback.
It includes:

- Login File Success Overlay
- updated loading animation. Initial loading now covers the entire page except for header until loading is finished. Subsequent loading states will be displayed either in the Password(-Setter)-Box, or in the button that triggered the loading.
- A blurred Login File Dummy, which is displayed in the background of the login-file-unlock as well as the login-file-success screen.
- Updated return value for export. Words are only returned as successfully exported, when validation was successfully done.
- Updated export request, which includes an option to either skipWords or skipFile.
- removed go to recovery words link from initial export file page.
- minor update to remove-key
- improved back-arrow functionality. The password box will show (in export-file) as long as the key-password combination wasn't set prior from outside. If it does not show, the progress indicators represent that change, too. (This is still not 100% perfect, as in remove-key you still can: export file complete 3 steps, export words, export file again now 2 steps, then use back button until you reach first time you did export file and it will show the password box, but wrong indicators.)
- removes the back-arrows for initial pages of requests.
- some updated wording


src/request/export/Export.js
```diff
@@ -34,14 +34,46 @@ class Export { // eslint-disable-line no-unused-vars
...
+            /** @type {HTMLElement} */
+            (/** @type {HTMLElement} */ (document.getElementById(ExportFile.Pages.LOGIN_FILE_INTRO))
+                .querySelector('.page-header-back-button')).classList.add('display-none');
+        }
+
+        /** @type {HTMLDivElement} */
+        this._fileSuccessPage = (document.getElementById(Export.Pages.LOGIN_FILE_SUCCESS));
+
+        /** @type {HTMLButtonElement} */
+        const closeRequestButton = (this._fileSuccessPage.querySelector('button'));
```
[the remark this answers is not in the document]
sisou: Also, instead of adding your own click handler for this button, you could make it an `a` and give it the class `.page-header-back-button` and have it handled automatically by the global back button handler.

src/request/export/ExportFile.js
```diff
@@ -228,18 +229,26 @@ class ExportFile extends Nimiq.Observable {
...
         this._key = key;
+        this._password = password;
         if (password) {
-            this._passwordBox.hideInput(true);
-            this._password = password;
+            /* eslint-disable no-new */
+            new ProgressIndicator(this.$exportFileIntroPage.querySelector('.progress-indicator'), 2, 1);
+            new ProgressIndicator(this.$downloadFilePage.querySelector('.progress-indicator'), 2, 2);
+            /* eslint-enable no-new */
+            this._keyAlreadySet = true;
```
sisou: Seems to me the `_keyAlreadySet` variable rather stores if the _password_ was already set, not the key. So how is this variable different from `!!this._password`?

src/request/export/index.html
```diff
@@ -167,7 +178,7 @@ <h1 data-i18n="recovery-words-title" class="nq-h1">Write these 24 words on a pie
                 </div>
 
                 <div class="page-footer">
-                    <button class="to-validate-words nq-button light-blue" data-i18n="continue">Continue</button>
+                    <button class="to-validate-words nq-button light-blue" data-i18n="recovery-words-validate">validate words</button>
```
sisou: Capitalization? I know it's uppercased in the button, but grammatically?
```suggestion
                    <button class="to-validate-words nq-button light-blue" data-i18n="recovery-words-validate">Validate words</button>
```

src/request/export/index.html
```diff
@@ -188,44 +199,42 @@ <h1 data-i18n="export-file-intro-heading" class="nq-h1">Save your account</h1>
...
                         </li>
                         <li>
                             <svg class="nq-icon">
                                 <use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-skull"/>
                             </svg>
-                            <span>
-                                It is part of your browser’s storage and<br/>
-                                might be deleted by accident.
+                            <span data-i18n="export-file-intro-accident">
+                                As a part of your browser’s data it might be deleted by accident.
```
sisou: Non-standard apostrophe.
```suggestion
                                As a part of your browser's data it might be deleted by accident.
```

src/request/export/index.html
```diff
@@ -268,6 +277,44 @@ <h1 data-i18n="export-file-heading" class="nq-h1">Download Login File</h1>
...
+                            </svg>
+                            <span data-i18n="export-file-success-save-file">Save Login File</span>
+                        </li>
+                        <li class="nq-grey">
+                            <span class="nq-icon bullet-point"></span>
+                            <span data-i18n="export-file-success-create-backup">Create backup</span>
+                        </li>
+                    </ul>
+                    <h1 class="nq-h1" data-i18n="export-file-success-heading">Take 5 minutes for a backup</h1>
+                    <p class="nq-text" data-i18n="export-file-success-words-intro">There is no ‘forgot password’ option. Write down 24 words to create a secure backup.</p>
```
sisou: I don't know which apostrophe is correct.

```suggestion
                    <p class="nq-text" data-i18n="export-file-success-words-intro">There is no 'forgot password' option. Write down 24 words to create a secure backup.</p>
```

src/components/DownloadLoginFile.js
```diff
@@ -88,6 +88,14 @@ class DownloadLoginFile extends Nimiq.Observable {
         this.file = new LoginFile(Nimiq.BufferUtils.toBase64(encryptedEntropy), color);
     }
 
+    /**
+     * @param {Nimiq.Address} firstAddress
+     */
+    createDummyFile(firstAddress) {
+        const color = IqonHash.getBackgroundColorIndex(firstAddress.toUserFriendlyAddress());
+        this.file = new LoginFile(Nimiq.BufferUtils.toBase64(new Uint8Array(0)), color);
```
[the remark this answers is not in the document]
nibhar: potentially yes. I have left it out for now, but we absolutely can.

src/request/export/Export.js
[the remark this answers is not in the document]
nibhar: good point. will add the `.page-header-back-button`

src/request/export/ExportFile.js
nibhar: It is used to indicate that the passowrd and key were set from outside of ExportFile. this._password is not only set in this method, but in other places inside of ExportFile, too.

sisou: Ok, but in the way it is currently used to skip the password input, how is it different from `!!this._password`?

nibhar: the password should only be skipped, when the indicators also show it as a 2-step process. That only happens, when the key was decrypted elsewhere (recovery words in remove-key). !!this._password is not sufficient for that as that would lead to the password being skipped when going back to the first page after entering your password in export file, while the indicators still show it as a 3-step process.

nibhar: I'll change it though, making `_keyAlreadySet` obsolete and changing how `_password` is used. Makes more sense I think.

nibhar: Should be much cleaner now

nibhar:
1 and 2 are done in a different PR after this one.
3 I don't exactly know how it happens, still investigating.
4 I have fixed.

PR #264, opened 2019-04-11
Add address details

sisou:
Add a pop-over after selecting an identicon that shows the address.

### DIFF
--- a/src/request/derive-address/index.html
+++ b/src/request/derive-address/index.html
@@ -113,6 +113,9 @@
             </div>
             <div id="effect-container">
                 <div class="page-header nq-card-header">
+                    <a tabindex="0" class="page-header-back-button display-none">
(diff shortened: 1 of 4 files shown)


src/request/derive-address/index.html
```diff
@@ -99,12 +101,24 @@ <h1 data-i18n="derive-address-heading-password" class="nq-h1">Unlock your accoun
...
+                </button>
+                <div id="details" class="addressInfo detailed-view">
+                    <div class="identicon"></div>
+                    <div class="address"></div>
+                </div>
+                <button class="nq-button light-blue confirm" data-i18n="derive-address-add-to-account-button">Add to account</button>
             </div>
+            <div id="effect-container">
+                <div class="page-header nq-card-header">
+                    <h1 data-i18n="derive-address-heading-choose-identicon" class="nq-h1">Choose a new address</h1>
```
nibhar: Missing a back button here

sisou: Only when key is encrypted. But yes.

PR #266, opened 2019-04-15
PasswordBox Update and minor styling changes

### LINKED ISSUE #197: Remove key: Animate label input to turn into logout button

### LINKED ISSUE #250: Refocus password setter for repeat
When confirming the first password with the button, focus is removed from the password input and users need to manually focus the password input again for the password-repeat.

### LINKED ISSUE #271: Reloading wordsOnly export does not init
When starting an `export` request with the `wordsOnly` flag, and then reloading the popup (F5), the request does not get started correctly. Instead, I see a blurred Login File.

I have not tried the `fileOnly` or no-flag request yet.

nibhar:
This PR updates the password box. It includes an updating progress bar for password setter box as well. Also included is a smaller update to remove-key styling.

resolves #250 
resolves #197 
resolves #271

### DIFF
--- a/src/components/FlippableHandler.js
+++ b/src/components/FlippableHandler.js
@@ -8,6 +8,16 @@ class FlippableHandler {
         if (!FlippableHandler.flippableHandlerInitialised) {
             /** @type {HTMLElement} */
             const $rotationContainer = (document.getElementById('rotation-container'));
+            if (window.location.hash) {
+                const $page = document.querySelector(window.location.hash);
+                if ($page) {
+                    $rotationContainer.classList.add('disable-transition');
+                    window.setTimeout(() => {
+                        $rotationContainer.classList.toggle('flipped', $page.classList.contains(classname));
+                        window.setTimeout(() => $rotationContainer.classList.remove('disable-transition'), 10);
+                    }, 0);
+                }
+            }
             window.addEventListener('hashchange', event => {
                 const newHash = new URL(event.newURL).hash;
                 const oldHash = new URL(event.oldURL).hash;
--- a/src/components/PasswordSetterBox.js
+++ b/src/components/PasswordSetterBox.js
@@ -1,8 +1,10 @@
+/* global Constants */
 /* global Nimiq */
 /* global I18n */
 /* global PasswordInput */
 /* global AnimationUtils */
 /* global PasswordStrength */
+/* global TopLevelApi */
 /* global TemplateTags */
 
 class PasswordSetterBox extends Nimiq.Observable {
@@ -58,7 +60,7 @@ class PasswordSetterBox extends Nimiq.Observable {
 
             <div password-input></div>
 
-            <button class="submit" data-i18n="passwordbox-continue">Continue</button>
+            <button class="submit" data-i18n="passwordbox-repeat">Repeat password</button>
 
             ${options.hideSkip ? '' : TemplateTags.noVars`
                 <a tabindex="0" class="password-skip nq-text-s">
@@ -107,18 +109,9 @@ class PasswordSetterBox extends Nimiq.Observable {
         this.$el.classList.remove('repeat');
     }
 
-    /**
-     * @returns {Promise<void>}
-     */
     async onPasswordIneligible() {
-        // We have to shake both possible too-weak notices
         const $hintTooShort = /** @type {HTMLElement} */ (this.$el.querySelector('.password-strength.strength-short'));
-        const $hintWeak = /** @type {HTMLElement} */ (this.$el.querySelector('.password-strength.strength-weak'));
-
-        await Promise.all([
-            AnimationUtils.animate('shake', $hintTooShort),
-            AnimationUtils.animate('shake', $hintWeak),
-        ]);
+        await AnimationUtils.animate('shake', $hintTooShort);
     }
 
     /**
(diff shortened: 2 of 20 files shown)


nibhar:
> No way we deploy this without a way to start over with pasword input.

you can just use the back arrow?

src/components/PasswordSetterBox.js
```diff
@@ -166,25 +168,22 @@ class PasswordSetterBox extends Nimiq.Observable {
...
             this.$el.classList.add('repeat');
             this.fire(PasswordSetterBox.Events.ENTERED);
+            this._passwordInput.focus();
             return;
         }
         if (this._password !== this._passwordInput.text) {
-            this.reset(true);
-            this.fire(PasswordSetterBox.Events.NOT_EQUAL);
-            return;
+            await AnimationUtils.animate('shake', this._passwordInput.$el);
```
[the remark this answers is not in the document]
julianbauer: how do you mean @sisou ?

Bettelstab:
> > No way we deploy this without a way to start over with pasword input.
> 
> you can just use the back arrow?

Nope, I stil have to input the same password then.

src/components/PasswordSetterBox.js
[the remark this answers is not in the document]
Bettelstab: @julianbauer We need a possibility for users which entered the wrong password the *first* time to somehow proceed.

Bettelstab:
> > > No way we deploy this without a way to start over with pasword input.
> > 
> > 
> > you can just use the back arrow?
> 
> Nope, I stil have to input the same password then. At least it was like this in my tests.

Clarification: Just figured out with @sisou that the back button in the interface *does* work, but browser back doesn't.

src/components/PasswordSetterBox.js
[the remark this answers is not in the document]
nibhar: When a user presses enter after repeating the password wrong, the password box does not reset anymore. The user stays in the repeat step and the password he did enter so far stays. Do you want it reset, or not is the question.

src/request/TopLevelApi.js
```diff
@@ -156,6 +157,14 @@ class TopLevelApi extends RequestParser {
                 handler.run();
 
                 TopLevelApi.setLoading(false);
+
+                if (TopLevelApi.getDocumentWidth() > Constants.MIN_WIDTH_FOR_AUTOFOCUS) {
+                    const $page = document.querySelector(window.location.hash);
+                    if ($page && $page.querySelector('.password-box')) {
```
sisou: Would it be possible to do this instead?
```suggestion
                    if (document.querySelector('.page:target .password-box')) {
```

src/request/TopLevelApi.js
```diff
@@ -156,6 +157,14 @@ class TopLevelApi extends RequestParser {
                 handler.run();
 
                 TopLevelApi.setLoading(false);
+
+                if (TopLevelApi.getDocumentWidth() > Constants.MIN_WIDTH_FOR_AUTOFOCUS) {
```
sisou: You could extract this into it's own function `TopLevelApi.focus()` and then also call it from all other places where `passwordBox.focus()` is currently called.

src/components/FlippableHandler.js
```diff
@@ -8,6 +8,16 @@ class FlippableHandler {
         if (!FlippableHandler.flippableHandlerInitialised) {
             /** @type {HTMLElement} */
             const $rotationContainer = (document.getElementById('rotation-container'));
+            if (window.location.hash) {
```
sisou: This is actually not relevant for PasswordBox, but is a fix for the recent export PR.

src/components/PasswordSetterBox.js
```diff
@@ -54,11 +54,13 @@ class PasswordSetterBox extends Nimiq.Observable {
             <div class="password-strength strength-good   nq-text-s" data-i18n="passwordbox-password-strength-good"  >Ok, that is an average password</div>
```
sisou: I cannot add a comment to the line above this (too far away from the diff), but the text `That password is too weak` does no longer hold true, because the password is still allowed. Should be changed to `That password is weak` or similar.

sisou:
There is also no a visual glitch when going to "export words" during Logout. While the page is spinning around, the blurred Login File is shown on the white side.

src/components/PasswordSetterBox.js
```diff
@@ -148,15 +177,14 @@ class PasswordSetterBox extends Nimiq.Observable {
...
+        if (this._passwordInput.text.length < PasswordInput.DEFAULT_MIN_LENGTH) return false;
+        return true;
     }
 
     /**
      * @param {Event} event
      */
-    _onSubmit(event) {
+    async _onSubmit(event) {
         event.preventDefault();
```
sisou: Maybe this function should also not be executed during password repeat (just like the button is hidden, ENTER should then also not work). So there could be `if (this._password) return` here.

I'm not sure it should be though, so this is not a change request. More a discussion starter.

src/components/PasswordSetterBox.js
```diff
@@ -195,6 +201,8 @@ class PasswordSetterBox extends Nimiq.Observable {
             this._passwordInput.reset();
             this.$el.classList.add('repeat');
             this.fire(PasswordSetterBox.Events.ENTERED);
+            window.history.replaceState({ repeat: false }, 'Keyguard');
+            window.history.pushState({ repeat: true }, 'Keyguard');
```
sisou: Using the history, it is much better now. Only issue I have is the name of the state property. I think we should namespace it more specificly, something like `isPasswordRepeat`

src/components/PasswordSetterBox.js
```diff
@@ -36,6 +36,12 @@ class PasswordSetterBox extends Nimiq.Observable {
         }
 
         this._onInputChangeValidity(false);
+
+        window.onpopstate = /** @param {PopStateEvent} ev */ ev => {
+            if (ev.state && !ev.state.repeat) {
```
sisou: Additionally, I think the reset event should only fire if the property is explicitly `false` and not also for `undefined`.

sisou:
I have just tried out the changes, and the history method does not actually work for me. Nothing happens after the first click on the back button while in repeat-password state. Neither on export(file) nor on change password (both cases for a key that is unencrypted).

Furthermore, the password repeat input shakes and tells me "the password is too short" while repeating and stopping to input after e.g. 3 characters (here the rule is missing that prevents validation while the input is shorter than the first password).

src/components/PasswordSetterBox.js
```diff
@@ -107,32 +116,59 @@ class PasswordSetterBox extends Nimiq.Observable {
...
+                );
+                return;
+            }
+            if (this._passwordInput.text.length > 0) {
+                if (this._passwordInput.text.length < this._password.length) {
+                    this._repeatPasswordTimout = window.setTimeout(
+                        async () => {
+                            this.$el.classList.remove('repeat-long');
+                            this.$el.classList.add('repeat-short');
+                            await AnimationUtils.animate('shake', this._passwordInput.$el);
```
sisou: Please remove this shake here. It does not make any sense, as the password entry is not finished yet.

nibhar:
The visibility indicator for the shake animation I'd like to postpone until it is clear whether or not the shake animation for the input stays, or if the box will be shaken.

PR #269, opened 2019-04-15
UI style adaptions

sisou:
Implements various style adaptions from https://github.com/nimiq/accounts/issues/182.


src/common.css
```diff
@@ -226,6 +226,15 @@ body.loading .page:target .password-box > .loading-spinner {
     z-index: 3;
```
nibhar: I can't mark it here, but the rule in Line 53 should state:
```
.page-header-back-button {
    width: 3rem;
    height: 3.6rem;
    position: absolute;
    left: 4rem;
    cursor: pointer;
    opacity: 0.4;
    transition: opacity .3s ease, left .3s ease;
}
```
to adjust for the `line-height: 1.2;`.

src/request/export/Export.css
```diff
@@ -42,25 +42,21 @@ ul.nq-list li > .nq-icon {
...
 
-.page#recovery-words .nq-text {
-    text-align: center;
-    color: var(--nimiq-orange); /* Can be removed when CSS specificity is fixed in nimiq-styles */
-    margin-bottom: 0;
-}
-
 .page#recovery-words .recovery-words {
-    max-height: 100%;
+    height: 39rem;
```
nibhar: this seems too large of a number to me. Looks better with 
```suggestion
    height: 35rem;
```
combined with the following margins and paddings in my opinion. 
Not really a requested change, just my personal preference.

src/request/export/Export.css
```diff
@@ -42,25 +42,21 @@ ul.nq-list li > .nq-icon {
...
     margin: 0;
     position: relative;
-    margin: 0 -1.5rem -3rem;
-    padding: 3rem 0;
-    -webkit-mask-image: linear-gradient(0deg , rgba(255,255,255,0), rgba(255,255,255, 1) 20%, rgba(255,255,255,1) 80%, rgba(255,255,255,0));
-    mask-image: linear-gradient(0deg , rgba(255,255,255,0), rgba(255,255,255, 1) 20%, rgba(255,255,255,1) 80%, rgba(255,255,255,0));
+    margin: -4rem -4rem -5rem;
```
nibhar:
```suggestion
    margin: -2rem -4rem -5rem;
```

src/request/export/Export.css
```diff
@@ -42,25 +42,21 @@ ul.nq-list li > .nq-icon {
...
+    padding: 4rem 3rem;
```
nibhar:
```suggestion
    padding: 2rem 3rem;
```

src/request/remove-key/remove-key.css
```diff
@@ -70,15 +71,19 @@
...
 .page#remove-key .page-body {
     display: flex;
     flex-flow: row wrap;
     justify-content: center;
 }
 
 .page#remove-key .page-footer button {
-    margin-top: 0;
-    width: calc(100% - 7rem);
+    margin: 3.5rem auto 3rem;
```
nibhar: remind me to change this in the passwordBox update transition ;)

src/request/export/Export.css
```diff
@@ -42,25 +42,21 @@ ul.nq-list li > .nq-icon {
     padding-bottom: 0;
```
nibhar:
```suggestion
    padding-bottom: 0;
    justify-content: space-evenly;
```

src/request/export/Export.css
```diff
@@ -238,10 +232,6 @@ ul.nq-list li > .nq-icon {
     display: none;
 }
 
-.page#validate-words .validate-words .word-list {
-    margin: 0;
-}
-
 .page-footer {
     padding: 1rem;
 }
```
nibhar: The following rules about `.skip-words` need either to add `.skip`, or replaced by .skip everywhere, since `ValidateWords` does use `.skip`, which does not have a `:hover` rule defined. 

Potentially this can even go to `common.css`.

src/request/export/Export.css
sisou: When reducing the padding, the gradient would overlap the words, which was explicitly requested not to happen by Julian. We would need to reduce the gradient to 2rem, which is too strong I think.

src/components/ProgressIndicator.css
```diff
@@ -12,11 +12,17 @@
...
+    opacity: 0.1;
 }
 
-.nq-card-header .progress-indicator ~ * {
+.nq-blue-bg .progress-indicator .indicator {
+    background: white;
+    opacity: 0.25;
+}
+
+.nq-card-header .progress-indicator + * {
```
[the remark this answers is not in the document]
sisou: Then I need to make a second rule for the arrow, because we now also have `.nq-notice`s in the header, which are also affected by a sibling selector.

src/request/export/Export.css
sisou: Handled in #277.

sisou:
@Bettelstab without a reproducable set of steps to create this situations, it's really not that helpful. I have tried for 15 minutes now to get what you are seeing, but I cannot.

Bettelstab:
http://drive.nim.team/s/BY4sRwXn6kLcH3R
up to date.
That's not a blocker I would say, but weird.

sisou:
The word import style issues seems to be happening when cutting a word from it's input, yes? E.g. with CTRL+X. This does not happen in my Chrome (Fedora), so this might be an issue specific to MacOS not sending an `input` event or similar after cutting. I will ignore this in this PR, however.

sisou:
@Bettelstab input field style problem is tracked in #279.

PR #270, opened 2019-04-15
Client v0.4.1

sisou:
Prepares client for KEY_NOT_FOUND import notice:

image

### DIFF
--- a/src/request/import/ImportApi.js
+++ b/src/request/import/ImportApi.js
@@ -14,8 +14,8 @@ class ImportApi extends TopLevelApi {
 
         const parsedRequest = {};
         parsedRequest.appName = this.parseAppName(request.appName);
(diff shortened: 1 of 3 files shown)


src/request/import/ImportApi.js
```diff
@@ -14,8 +14,8 @@ class ImportApi extends TopLevelApi {
 
         const parsedRequest = {};
         parsedRequest.appName = this.parseAppName(request.appName);
-        parsedRequest.defaultKeyPath = this.parsePath(request.defaultKeyPath, 'defaultKeyPath');
```
Bettelstab: This one we just don't use I guess?

sisou: Exactly. It was used back when we had the wallet type selector (selection between legacy and bip39 in Keyguard).

Bettelstab: ok :)

PR #275, opened 2019-04-17
Root redirect

Bettelstab:
Redirects to a configurable URL when user accidentally visits root. Default: testnet safe

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -155,10 +155,8 @@ for DIR in src/request/*/ ; do
     replace_icon_sprite_url dist/request/${REQUEST}/index.html
 done
 
-# create bundle for redirect page
-cat src/lib/Constants.js >> dist/redirect.js
-cat src/config/config.$BUILD.js >> dist/redirect.js
-cat src/redirect.js >> dist/redirect.js
+# copy root redirect script
+cp src/redirect.js dist
 
 # replace scripts in redirect page and output result in dist
     awk '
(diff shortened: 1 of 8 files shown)


sisou:
Please rebase onto master to get rid of the already merged changes in the changeset from #274.

tools/build.sh
```diff
@@ -155,6 +155,29 @@ for DIR in src/request/*/ ; do
...
+            skip_script = 0
+            skip_link = 0
+        }
+        /<script/ {
+            # Replace first script tag
+            if (!skip_script) {
+                skip_script = 1
+                # Preserve whitespace / intendation. Note: 1 is first array index in awk
+                split($0, space, "<")
+                print space[1] "<script defer src=\"/request/'${JS_COMMON_BUNDLE}'\"></script>"
```
sisou: What do we need the common bundle for in the redirect page? All it needs is the constants and the config, both of which are `manual` and then only the redirect script.

Bettelstab: That's right. I just wanted to avoid to create another bundle file (constants + config).

sisou: Ah, right, they are included inbthat bundle nevertheless. Ok, let's gooooo!

Bettelstab: Which one do you prefer now? Just did it :D

Bettelstab: Maybe it's cleaner that way and the additional code size is not really noteworthy.

sisou: Yeah, I think using the common bundle is fine. It's used in all other pages anyway.

However, I just noticed that you could also add an index.html to `/request`, to prevent that dead-end. That index could redirect to root, which then redirects as configured.

Bettelstab: Turns out I can use the same index.html for both root and /request.

PR #277, opened 2019-04-25
Skip buttons

nibhar:


src/common.css
```diff
@@ -286,3 +286,41 @@ body.loading .page:target .password-box > .loading-spinner {
...
+    opacity: 0.7;
+    text-decoration: none;
+}
+
+.skip:hover .nq-icon,
+.skip:focus .nq-icon {
+    transform: translate3D(-0.25rem, 0, 0);
+}
+
+:not(.password-box) > :not(.password-box) + .skip,
```
sisou: Double `:not(.password-box)`?

src/request/export/Export.css
```diff
@@ -43,7 +43,7 @@ ul.nq-list li > .nq-icon {
 }
 
 .page#recovery-words .recovery-words {
-    height: 39rem;
+    height: 37rem;
```
sisou: Does this have anything to do with the skip buttons? Otherwise I disagree with it.

PR #280, opened 2019-04-29
Subresource integrity

sisou:
This PR adds `integrity` attributes to all `<script>` and most `<link`> in the build-output.

Subresource integrity is described here: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

To do this, I had to split the per-request handling in the build script into bundle-contents-collection and index.html-writing, as the bundles need to be complete to generate their hash, before writing them into the respective script and link tags in the index.html files.

I have left out the Google Fonts CSS links, as those may change unannounced. The integrity for our own CSS bundles is still validated by the browser. For the script tags, we will additionally require their integrity checks via the CSP.

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -51,6 +51,11 @@ replace_core_lib_url() {
     sed -i -e "s/$OLD_PATH/$NEW_PATH/g" $1
 }
 
+# generate a base64 file integrity hash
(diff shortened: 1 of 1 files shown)


nibhar:
We should enable the corresponding csp once this is merged.

PR #282, opened 2019-04-30
Skip Buttons

nibhar:
Skip buttons now only appear on:

- change-password's file download page to skip the download of the file
- login's file download page after word import to skip the download of the file
- complete export's file-completed page to skip recovery words

Styles for the skip button are moved to common.css.

### DIFF
--- a/src/request/export/ExportFile.js
+++ b/src/request/export/ExportFile.js
@@ -66,11 +66,7 @@ class ExportFile extends Nimiq.Observable {
                 hideCancel: true,
             },
         );
-        this._passwordSetterBox = new PasswordSetterBox(
-            $passwordSetterBox, {
-                hideSkip: true,
-            },
-        );
+        this._passwordSetterBox = new PasswordSetterBox($passwordSetterBox);
         this._loginFileIcon = new LoginFileIcon($loginFileIcon);
         this._downloadLoginFile = new DownloadLoginFile($downloadLoginFile);
         this._downloadLoginFile.createDummyFile(this._request.keyInfo.defaultAddress);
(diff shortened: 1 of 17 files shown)


src/request/export/ExportFile.js
```diff
@@ -66,11 +66,7 @@ class ExportFile extends Nimiq.Observable {
                 hideCancel: true,
             },
         );
-        this._passwordSetterBox = new PasswordSetterBox(
-            $passwordSetterBox, {
-                hideSkip: true,
-            },
-        );
+        this._passwordSetterBox = new PasswordSetterBox($passwordSetterBox);
```
sisou: There should be no need for a password setter during export anymore, when keys cannot be unencrypted anyway. But we might want to keep it in anyway, to be sure?

src/request/export/ExportWords.js
```diff
@@ -72,21 +68,12 @@ class ExportWords extends Nimiq.Observable {
         /* eslint-enable no-new */
 
         // events
-        $noRecoverySkip.addEventListener('click', event => {
-            event.preventDefault();
-            this._resolve({ success: false });
-        });
-        $recoveryWordsSkip.addEventListener('click', event => {
```
sisou: Do we really want to remove the skip for the word validation? I liked that one.

sisou: As discussed, to not confuse users, word validation will be non-skippable.

src/request/export/ExportFile.js
nibhar: As discussed previously, I did not remove the functionality. It is just no longer reachable.

PR #285, opened 2019-05-06
Update color names

sisou:
According to recent decision.


sisou:
@nibhar ah yes, we are using the names directly for the CSS classes, which are not updated. So we either need a mapping from color name to CSS class, or update the class names in @nimiq/style.

nibhar:
I would prefer to update the names for consistency. Is somewhat breaking though.

sisou:
It's a breaking change: https://github.com/nimiq/nimiq-style/commit/cf822baca5480693237759e4865037181ebe4727

We can merge that and update everything after launch, when we have time.

PR #287, opened 2019-05-06
Wording update

nibhar:


src/components/DownloadLoginFile.js
```diff
@@ -68,7 +68,7 @@ class DownloadLoginFile extends Nimiq.Observable {
                 <span data-i18n="download-loginfile-download">Download Login File</span>
             </button>
             <span class="nq-label tap-and-hold" data-i18n="download-loginfile-tap-and-hold">Tap and hold image to download</span>
-            <button class="nq-button light-blue continue" data-i18n="download-loginfile-continue">Continue</button>
+            <button class="nq-button light-blue continue" data-i18n="download-loginfile-continue">Download</button>
```
sisou: This is not correct. This button is there to continue to the next step, after the long-tap download of the image file is expected to be completed. This button is a fallback for iOS devices, which do not support the `download` attribute on links and need a manual button to _continue_.

Thus this button does _not_ download the image.

src/components/PasswordSetterBox.js
```diff
@@ -56,18 +56,18 @@ class PasswordSetterBox extends Nimiq.Observable {
 
         /* eslint-disable max-len */
         $el.innerHTML = TemplateTags.hasVars(1)`
-            <div class="password-strength strength-short  nq-text-s" data-i18n="passwordbox-password-strength-short" >Enter at least 8 characters</div>
-            <div class="password-strength strength-weak   nq-text-s" data-i18n="passwordbox-password-strength-weak"  >That is a weak password</div>
-            <div class="password-strength strength-good   nq-text-s" data-i18n="passwordbox-password-strength-good"  >Ok, that is an average password</div>
-            <div class="password-strength strength-strong nq-text-s" data-i18n="passwordbox-password-strength-strong">Great, that is a strong password</div>
-            <div class="password-strength strength-secure nq-text-s" data-i18n="passwordbox-password-strength-secure">Super, that is a secure password</div>
+            <div class="password-strength strength-short  nq-text-s" data-i18n="passwordbox-password-strength-short" >Enter min. 8 characters</div>
```
sisou: Suggestion to change this to
```
Enter 8 characters or more
```
@julianbauer

src/components/PasswordSetterBox.js
```diff
@@ -56,18 +56,18 @@ class PasswordSetterBox extends Nimiq.Observable {
...
+            <div class="password-strength strength-weak   nq-text-s" data-i18n="passwordbox-password-strength-weak"  >Sufficient password</div>
```
sisou: Suggestion to change this to
```
Weak password
```
to make clear that it is _weak_, and not actually sufficient in the interpreted meaning of the word.
@julianbauer

src/components/PasswordSetterBox.js
```diff
@@ -56,18 +56,18 @@ class PasswordSetterBox extends Nimiq.Observable {
...
+            <div class="password-strength strength-strong nq-text-s" data-i18n="passwordbox-password-strength-strong">Strong password</div>
+            <div class="password-strength strength-secure nq-text-s" data-i18n="passwordbox-password-strength-secure">Secure password</div>
             <div class="repeat-long nq-text-s" data-i18n="passwordbox-repeat-password-long">No match, please try again</div>
             <div class="repeat-short nq-text-s" data-i18n="passwordbox-repeat-password-short">Password is too short</div>
             <div class="repeat-password nq-text-s" data-i18n="passwordbox-repeat-password">Repeat your password</div>
 
             <div password-input></div>
 
-            <button class="submit" data-i18n="passwordbox-repeat">Repeat password</button>
+            <button class="submit" data-i18n="passwordbox-confirm">Confirm</button>
```
sisou: This is the button text during the first password input, yes? I think this should still say "repeat password", as that is the next step. Seeing "confirm", I would not expect another input, but that I _confirm_ the password with this button (but instead I have to re-enter it).

I expect the word "confirm" to _end_ the input.

@julianbauer

src/request/create/index.html
```diff
@@ -114,7 +114,7 @@ <h1 class="nq-h1 show-on-repetition" data-i18n="create-heading-repeat-password">
                     <svg class="nq-icon"><use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-hexagon"/></svg>
                     <div class="identicon"></div>
                 </div>
-                <h2 class="nq-h2 nq-blue" data-i18n="create-wallet-desc">This is your new account with its first address.</h2>
+                <h2 class="nq-h2 nq-blue" data-i18n="create-wallet-desc">This is your Account with your first Address in it.</h2>
```
sisou: @julianbauer 
Elion explicitly suggested to not use `in it` here:

>when creating new account current wording is "This is your account with your first address in it"
>"in it" is odd and not really smooth english this way, also would emphasize that it is new.
>solution: update wording to "This is a new account with its first address." or "Your new account with its first address"

I actually changed it to that, and now this PR is changing it back.

src/request/derive-address/index.html
```diff
@@ -79,12 +79,12 @@
     <div id="app">
         <div id="unlock" class="page nq-card">
             <div class="page-header nq-card-header">
-                <h1 data-i18n="derive-address-heading-password" class="nq-h1">Unlock your account</h1>
+                <h1 data-i18n="derive-address-heading-password" class="nq-h1">Unlock account</h1>
```
sisou: I think `account` should be capitalized here for consistency:
```suggestion
                <h1 data-i18n="derive-address-heading-password" class="nq-h1">Unlock Account</h1>
```

src/request/derive-address/index.html
```diff
@@ -109,14 +109,14 @@ <h1 data-i18n="derive-address-heading-password" class="nq-h1">Unlock your accoun
...
-                <button class="nq-button light-blue confirm" data-i18n="derive-address-add-to-account-button">Add to account</button>
+                <button class="nq-button light-blue confirm" data-i18n="derive-address-add-to-account-button">Add to Account</button>
             </div>
             <div id="effect-container">
                 <div class="page-header nq-card-header">
                     <a tabindex="0" class="page-header-back-button display-none">
                         <svg class="nq-icon"><use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-arrow-left"/></svg>
                     </a>
-                    <h1 data-i18n="derive-address-heading-choose-identicon" class="nq-h1">Choose a new address</h1>
+                    <h1 data-i18n="derive-address-heading-choose-identicon" class="nq-h1">Choose a new Avatar</h1>
```
sisou: @julianbauer 
If we remove `a` and `an` everywhere, why not here?
```suggestion
                    <h1 data-i18n="derive-address-heading-choose-identicon" class="nq-h1">Choose new Avatar</h1>
```

src/request/export/index.html
```diff
@@ -114,7 +114,7 @@ <h1 data-i18n="export-words-intro-heading" class="nq-h1">There is no password re
                                 <use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-paper-edit"/>
                             </svg>
                             <span data-i18n="recovery-words-intro-offline">
-                                Keep your Words offline, don’t enter them anywhere but in the Nimiq Safe.
+                                Keep your words offline, enter them nowhere but on keyguard.nimiq.com
```
sisou: @julianbauer 
Suggestion to make the domain bold and add a period at the end:
```suggestion
                                Keep your words offline, enter them nowhere but on <strong>keyguard.nimiq.com</strong>.
```

src/request/export/index.html
```diff
@@ -168,15 +168,15 @@ <h1 data-i18n="create-heading-validate-backup" class="nq-h1">Validate your backu
...
+                    <p class="nq-notice warning" data-i18n="recovery-words-text">Anyone with these words can access your Account! Keep them safe.</p>
                 </div>
 
                 <div class="page-body nq-card-body">
                     <div class="recovery-words"></div>
                 </div>
 
                 <div class="page-footer">
-                    <button class="to-validate-words nq-button light-blue" data-i18n="recovery-words-validate">Validate words</button>
+                    <button class="to-validate-words nq-button light-blue" data-i18n="recovery-words-validate">Confirm backup</button>
```
sisou: `Validate backup`? @julianbauer

src/request/export/index.html
```diff
@@ -213,19 +213,19 @@ <h1 data-i18n="export-file-intro-heading" class="nq-h1">Save your account</h1>
                                 <use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-skull"/>
                             </svg>
                             <span data-i18n="export-file-intro-accident">
-                                As a part of your browser's data it might be deleted by accident.
+                                It might be deleted automatically or by accident.
```
sisou: I think we should clarify here, that it is not _us_ who is deleting it automatically, but the operating system and browser.

src/request/export/index.html
```diff
@@ -213,19 +213,19 @@ <h1 data-i18n="export-file-intro-heading" class="nq-h1">Save your account</h1>
...
                             </span>
                         </li>
                         <li>
                             <svg class="nq-icon">
                                 <use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-login"/>
                             </svg>
                             <span data-i18n="export-file-intro-download-file">
-                                Download a Login File to stay save and to login to other browsers and devices.
+                                Download the Login File to save your Account.
```
sisou: @julianbauer 
```suggestion
                                Download the Login File to secure access to your Account.
```

src/request/import/index.html
```diff
@@ -155,8 +155,8 @@ <h1 data-i18n="import-upload-login-file" class="nq-h1">Upload your Login File</h
                     <button data-i18n="import-file-button-words" class="nq-button-s" id="goto-words">
                         Login with Recovery Words
                     </button>
-                    <a class="nq-text-s nq-blue" id="goto-create" href="#">
-                        <span data-i18n="import-create-account">Create new account</span>
+                    <a class="skip nq-link" id="goto-create" href="#">
```
sisou: This line should go into the skip button PR and be removed here.

src/components/DownloadLoginFile.js
sisou: Furthermore, the title for the Login File download pages is changed from `Download Login File` to `Save Account`, so this button should not say `Download` anymore.

src/request/remove-key/index.html
```diff
@@ -153,10 +153,8 @@ <h1 data-i18n="remove-key-heading" class="nq-h1">Don't lose access</h1>
                         <svg class="nq-icon"><use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-arrow-left"/></svg>
                     </a>
                     <h1 data-i18n="export-words-intro-heading" class="nq-h1">There is no password recovery!</h1>
-                    <p class="nq-notice warning" data-i18n="recovery-words-intro-text">
-                        The Recovery Words are the ONLY way to
-                        restore your account in case you lose your
-                        Login File or password.
+                    <p class="nq-notice warning" data-i18n="remove-key-intro-text">
+                        If you log out without saving your Account, you will irreversible lose access to it!
```
sisou: Grammar
```suggestion
                        If you log out without saving your Account, you will irreversibly lose access to it!
```

src/request/derive-address/index.html
nibhar: missed that one.

src/request/import/index.html
sisou: @nibhar I added this skip button to its PR (https://github.com/nimiq/keyguard-next/pull/282/commits/c25b1f30ff9994b4e61cc6fa422b746c010b6b9c), you can remove this change from this PR.

src/components/PasswordSetterBox.js
Tammoovernice: Okay, let's go with this

src/components/PasswordSetterBox.js
Tammoovernice: Weak could be interpreted as not sufficient, while it is. But as the potential 'error' happening here is a user having to secure their PW further, let's go with your suggestion.

src/components/PasswordSetterBox.js
Tammoovernice: The difference here is what this button means – does it describe the action that a user performs by clicking it or does it describe where we sent the user after he/she clicked it. Still, you have a very valid point that this might lead to a false expectation that we disappoint a bit, so let's stay with the 'Repeat Password'.

src/request/create/index.html
Tammoovernice: 'new' is obsolete info here, let's go with 'This is your Account with your first Address.'

src/request/export/index.html
Tammoovernice: I don't think that makes a difference here. Obviously, we are not harmfully deleting user data on purpose. At the same time, we have built this system that comes with these dangers and deficiencies, so we are exposing a user to this risk. At the same time, with the launch of the new KG, there is the possibility that it is, in fact, us that might be deleting data. Let's stick to this clear message: It might be deleted, doesn't matter by whom.

src/components/DownloadLoginFile.js
sisou: Again, @Tammoovernice, _this_ is _not_ the download button. The download button is above in line 68.

This button here is a button that appears only on iOS devices, where image _cannot_ be downloaded with a tap (because iOS does not support that), but _has to be_ downloaded with a long-tap. Because of all the things that can go wrong or the user aborting the save-file-dialog, we do not automatically continue to the next screen like we do for a regular click/tap on all other devices. This button is there to manually continue after downloading the image.

PR #292, opened 2019-05-06
Support for export of pre-migration legacy accounts

sisou:
To be able to export the Recovery Words of legacy accounts before running the migration.

### DIFF
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -78,6 +78,7 @@ export type SimpleRequest = BasicRequest & {
 export type ExportRequest = SimpleRequest & {
     fileOnly?: boolean,
     wordsOnly?: boolean,
+    useLegacyStore?: boolean,
 };
 
 export type ExportResult = {
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -78,7 +78,6 @@ export type SimpleRequest = BasicRequest & {
 export type ExportRequest = SimpleRequest & {
     fileOnly?: boolean,
     wordsOnly?: boolean,
-    useLegacyStore?: boolean,
 };
 
 export type ExportResult = {
--- a/src/lib/AccountStore.js
+++ b/src/lib/AccountStore.js
@@ -1,3 +1,8 @@
+/* global Nimiq */
+/* global Key */
+/* global KeyInfo */
+/* global KeyStore */
+
 /**
  * DEPRECATED
  * This class is only used for retrieving keys and accounts from the old KeyStore.
@@ -114,6 +119,44 @@ class AccountStore {
             };
             openCursorRequest.onerror = () => reject(openCursorRequest.error);
         });
+    }
+
+    /**
+     * @param {string} id
+     * @param {Uint8Array} password
+     * @returns {Promise<Key?>}
+     */
+    async get(id, password) {
+        /** @type {KeyRecord?} */
+        const keyRecord = await this._get(id);
+        if (!keyRecord) return null;
+
+        const secret = await Nimiq.Secret.fromEncrypted(new Nimiq.SerialBuffer(keyRecord.secret), password);
+        return new Key(secret, keyRecord.hasPin);
+    }
+
+    /**
+     * @param {string} id
+     * @returns {Promise<KeyInfo?>}
+     */
+    async getInfo(id) {
+        const keyRecord = await this._get(id);
+        if (!keyRecord) return null;
+        return KeyInfo.fromObject(keyRecord, KeyStore.isEncrypted(keyRecord), keyRecord.defaultAddress);
+    }
+
+    /**
+     * @param {string} id
+     * @returns {Promise<KeyRecord?>}
+     * @private
+     */
+    async _get(id) {
+        const db = await this.connect();
+        if (!db) return null;
+        const transaction = db.transaction([AccountStore.ACCOUNT_DATABASE], 'readonly');
+        const request = transaction.objectStore(AccountStore.ACCOUNT_DATABASE).get(id);
+        const accountRecord = await KeyStore._requestToPromise(request, transaction);
+        return KeyStore.accountRecords2KeyRecords([accountRecord])[0];
     }
 
     async close() {
--- a/src/lib/AccountStore.js
+++ b/src/lib/AccountStore.js
@@ -156,6 +156,7 @@ class AccountStore {
         const transaction = db.transaction([AccountStore.ACCOUNT_DATABASE], 'readonly');
         const request = transaction.objectStore(AccountStore.ACCOUNT_DATABASE).get(id);
         const accountRecord = await KeyStore._requestToPromise(request, transaction);
+        if (!accountRecord) return null;
         return KeyStore.accountRecords2KeyRecords([accountRecord])[0];
     }
 
--- a/src/lib/AccountStore.js
+++ b/src/lib/AccountStore.js
@@ -142,7 +142,9 @@ class AccountStore {
     async getInfo(id) {
         const keyRecord = await this._get(id);
         if (!keyRecord) return null;
-        return KeyInfo.fromObject(keyRecord, KeyStore.isEncrypted(keyRecord), keyRecord.defaultAddress);
+        const keyInfo = KeyInfo.fromObject(keyRecord, KeyStore.isEncrypted(keyRecord), keyRecord.defaultAddress);
+        keyInfo.useLegacyStore = true;
+        return keyInfo;
     }
 
     /**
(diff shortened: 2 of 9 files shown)


src/lib/RequestParser.js
```diff
@@ -78,13 +79,16 @@ class RequestParser { // eslint-disable-line no-unused-vars
 
     /**
      * @param {any} keyId
+     * @param {boolean} [useLegacyStore = false]
      * @returns {Promise<KeyInfo>}
      */
-    async parseKeyId(keyId) {
+    async parseKeyId(keyId, useLegacyStore = false) {
```
nibhar: Could this potentially be determined by the used `keyId`? As far as I can tell the `AccountStore` used the `userFriendlyAddress` as key while the `KeyStore` uses `id`.

client/src/PublicRequest.ts
```diff
@@ -78,6 +78,7 @@ export type SimpleRequest = BasicRequest & {
 export type ExportRequest = SimpleRequest & {
     fileOnly?: boolean,
     wordsOnly?: boolean,
+    useLegacyStore?: boolean,
```
nibhar: in case automatically detecting from which store to read this would be unnecessary.

src/request/export/ExportWords.js
```diff
@@ -19,7 +20,7 @@ class ExportWords extends Nimiq.Observable {
      * if a complete page is missing it will be created.
      * However these pages wil be the default pages which usually don't match the applications requirements.
      * Refer to the corresponsing _build(Privcy | RecoveryWords | ValidateWords) to see the general Structure.
-     * @param {Parsed<KeyguardRequest.SimpleRequest>} request
+     * @param {Parsed<KeyguardRequest.ExportRequest>} request
```
danimoh: Is an extra type `ExportRequest` now still required here or is that just a leftover from when the `useLegacyStore` flag had to be provided by the caller?

src/request/export/ExportWords.js
```diff
@@ -112,7 +113,23 @@ class ExportWords extends Nimiq.Observable {
...
+                /**
+                 * Because Typescript does not detect the ExportRequest to be different from
+                 * a SimpleRequest (because both extra properties of the ExportRequest are
+                 * optional), we cannot add a new condition for the ExportRequest to the
+                 * Parsed<> transform in types/Keyguard.d.ts.
+                 *
+                 * Because the keyInfo on the parsed request also does not contain any indication
+                 * whether the key is a legacy account or a regular key, we have otherwise no
+                 * idea which IndexedDB store to use.
+                 */
```
danimoh: Is this comment still accurate? The `useLegacyStore` flag seems to always be set to either true or false and is thus not optional.

tests/DummyData.spec.js
```diff
@@ -122,6 +122,18 @@ const deprecatedAccount2KeyInfoObjects = () => [{
...
+    KeyInfo.fromObject(
+        deprecatedAccount2KeyInfoObjects()[0],
+        true,
+        deprecatedAccount2KeyInfoObjects()[0].legacyAccount.address,
+    ),
+];
+
+const deprecatedAccount2Keys = () => [
+    new Key(secrets[1], false)
+];
```
danimoh: Why are these methods? They could be simply consts, e.g.:
```ts
const deprecatedAccountAsKeyInfo = KeyInfo.fromObject(...)
```
as the tests do not seem to manipulate the data.

src/lib/AccountStore.js
```diff
@@ -116,6 +121,45 @@ class AccountStore {
...
+     * @private
+     */
+    async _get(id) {
+        const db = await this.connect();
+        if (!db) return null;
+        const transaction = db.transaction([AccountStore.ACCOUNT_DATABASE], 'readonly');
+        const request = transaction.objectStore(AccountStore.ACCOUNT_DATABASE).get(id);
+        const accountRecord = await KeyStore._requestToPromise(request, transaction);
+        if (!accountRecord) return null;
+        return KeyStore.accountRecords2KeyRecords([accountRecord])[0];
```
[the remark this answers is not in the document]
sisou: AccountStore is a legacy class and should hopefully be removed in the future. I think there is nothing wrong with calling static helper functions on the KeyStore, instead of a higher-order class.

src/request/export/ExportWords.js
sisou: It would be good to have an extra type, yes, for the reasons in the comment below.

src/request/export/ExportWords.js
sisou: The problem is that the `useLegacyStore` flag only exists on the _parsed_ version of the request, not the incoming external request.

But because the `Parsed<>` transformer does not recognize the ExportRequest as another type other than SimpleRequest, I cannot add a special handling for it, which would not also apply to all other SimpleRequests.

tests/DummyData.spec.js
sisou: That is true. They are functions because my initial tries with the tests had dummy data, which was dependent on other function data.

demos/Demo.js
```diff
@@ -36,6 +37,7 @@ class Demo {
 
         /**
          * @type {AccountRecord}
+         * @description Password: 1234567890
```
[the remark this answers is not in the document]
sisou: Nobody looks at the code :D
This is for our convenience. There a 30 other vesting contracts in the testnet genesis block that we can switch to if somebody empties it.

src/request/export/ExportWords.js
sisou: The ExportRequest also has the `fileOnly` and `wordsOnly` flags, so it was actually a mistake that it was a SimpleRequest here before.

tests/DummyData.spec.js
sisou: Well, actually only the `deprecatedAccount2Keys` can be a `const`, because `deprecatedAccount2KeyInfos` depends on `deprecatedAccount2KeyInfoObjects()`, which depends on `hashes()`, which cannot be calculated before the Nimiq lib is loaded, which is only _after_ the DummyData file has been run.

sisou: Turns out, even the `deprecatedAccount2Keys` has to be a function, because `new Key()` internally uses `Nimiq` for default address derivation.

src/request/export/ExportWords.js
Bettelstab: If you have a look at https://github.com/nimiq/keyguard-next/blob/668e6944f1ada756b8f9d37d1f9a6069f1aa0351/src/request/remove-key/RemoveKey.js#L28 you'll notice we use it as the same type  indeed. The additional parameter would have to be parsed there as well. Maybe also at other places.
(Side note: would be cleaner to explicitly create a new request object there with an imported constructor, and probably to include a `kind` field in each request.)
My proposal for a quick fix without `ts-ignore`: https://github.com/nimiq/keyguard-next/tree/max/useLegacyStore-type-fix

sisou: I now resolved it by adding a `KeyInfo.useLegacyStore` flag, which is `false` by default and can be set/get by setters/getters. This way there is no type trickery or type inconsistency required and the only request using it is the export(words).

src/lib/AccountStore.js
```diff
@@ -116,6 +121,45 @@ class AccountStore {
...
+     * @param {string} id
+     * @returns {Promise<KeyRecord?>}
+     * @private
+     */
+    async _get(id) {
+        const db = await this.connect();
+        if (!db) return null;
+        const transaction = db.transaction([AccountStore.ACCOUNT_DATABASE], 'readonly');
+        const request = transaction.objectStore(AccountStore.ACCOUNT_DATABASE).get(id);
+        const accountRecord = await KeyStore._requestToPromise(request, transaction);
```
Bettelstab: `_requestToPromise` is not private anymore then.

src/lib/AccountStore.js
nibhar: could this not be:
```
const key = KeyStore.accountRecords2KeyRecords([accountRecord])[0];
key.useLegacyStore(true);
return key;
```
then you do not need it in RequestParser and this seems the logical place to me.

src/lib/RequestParser.js
```diff
@@ -84,10 +85,15 @@ class RequestParser { // eslint-disable-line no-unused-vars
...
-        const keyInfo = await KeyStore.instance.getInfo(keyId);
+
+        const keyInfo = this.isLegacyKeyId(keyId)
+            ? await AccountStore.instance.getInfo(keyId)
+            : await KeyStore.instance.getInfo(keyId);
+
         if (!keyInfo) {
             throw new Errors.KeyNotFoundError();
         }
+
```
nibhar: magnificent ;)

PR #294, opened 2019-05-07
Enable building on Macs

### LINKED ISSUE #293: sha256sum is not available on mac
the build.sh fails as it requires sha256sum. This is not available on mac.
Mac has `shasum -a 256` replacement

valentinvieriu:
Making sha256 check work on Mac.
Here is a proposal to resolve #293


sisou:
Superseded by #303.

PR #296, opened 2019-05-07
Update Identicon.html

### LINKED ISSUE #295: Identicon demo is not working
The Identicon demo is missing some dependencies

valentinvieriu:
Adding the missing dependencies.

resolve #295

### DIFF
--- a/demos/Identicon.html
+++ b/demos/Identicon.html
@@ -13,6 +13,8 @@
         margin: 0 1em;
     }
 </style>
+<script src="../src/lib/TemplateTags.js"></script>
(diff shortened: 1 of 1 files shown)


sisou:
The `TemplateTags` might be missing in other demos, too.

valentinvieriu:
> The `TemplateTags` might be missing in other demos, too.

I'm having a hard time running the demos using some simple http serve like ` python -m SimpleHTTPServer 9001` the service worker is acting funny.

I'll check more of them, but I think we might need a  small docker with nginx so that we can have https too

sisou:
This or is included in #304.

PR #297, opened 2019-05-07
Multiple Account Login

nibhar:
For the purpose of having multiple Accounts returned to the AccountsManager, the Legacy Accounts also need to be added to the storage, so they can be deleted if they are not needed.

Also adds test words for the 4 different cases of ambiguous Recovery Words.

### DIFF
--- a/src/request/import/ImportWords.js
+++ b/src/request/import/ImportWords.js
@@ -273,8 +273,7 @@ class ImportWords {
                 };
                 this._keyResults.push(result);
 
-                const secretString = Nimiq.BufferUtils.toBase64(key.secret.serialize());
-                sessionStorage.setItem(ImportApi.SESSION_STORAGE_KEY_PREFIX + result.keyId, secretString);
+                sessionStorage.setItem(ImportApi.SESSION_STORAGE_KEY_PREFIX + result.keyId, 'Private key');
             } else {
                 TopLevelApi.setLoading(false);
             }
--- a/src/request/import/ImportWords.js
+++ b/src/request/import/ImportWords.js
@@ -138,6 +138,7 @@ class ImportWords {
         // @ts-ignore (Property 'test' does not exist on type 'Window'.)
         window.test = (n = 0) => {
             const testPasswords = [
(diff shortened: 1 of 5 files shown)


valentinvieriu:
Can we have the option to delete Ledger accounts too?

Bettelstab:
@valentinvieriu This PR is not about removing accounts from keyguard. It's about handling a special case of login by 24 words where we cannot tell if the imported account is a legacy one or new bip39 one.
Keyguard and ledger do not interact, that is handled in the accounts repo. Logging out ledger accounts is possible: https://github.com/nimiq/accounts/pull/217 (It means removing data about the ledger account from your browser database, not from the ledger itself.)

valentinvieriu:
I've kind of realized that after adding the comment. Makes sense!
nice to see  nimiq/accounts#217 being merged. Do you know when can be tested?

src/request/import/ImportWords.js
```diff
@@ -241,6 +272,9 @@ class ImportWords {
                     wordsExported: true,
                 };
                 this._keyResults.push(result);
+
+                const secretString = Nimiq.BufferUtils.toBase64(key.secret.serialize());
+                sessionStorage.setItem(ImportApi.SESSION_STORAGE_KEY_PREFIX + result.keyId, secretString);
```
sisou: The IFrameApi does not check the contents of the stored sessionStorage item, only if it exists. Thus the secret does not need to be written into it, it is enough to write anything into to item, like `'This is not the key you are looking for.'`.

src/request/import/ImportWords.js
```diff
@@ -241,6 +272,8 @@ class ImportWords {
                     wordsExported: true,
                 };
                 this._keyResults.push(result);
+
+                sessionStorage.setItem(ImportApi.SESSION_STORAGE_KEY_PREFIX + result.keyId, '');
```
danimoh: I think using the empty string here is wrong, as it will be evaluated as falsy in `releaseKey` in `IframeApi`.
When changing it, I think force pushing would be good.

PR #299, opened 2019-05-07
Added the option to use nginx with Http2 to run the demos

### LINKED ISSUE #298: Have an https server to run the demos
Let's have a Docker based approach to be able to run the demos and test the website. We can use https and Http2 to simulate the real life situation of the keyguard

valentinvieriu:
Proposal to add a Based approach to run nginx with https and http2 enabled
resolve #298


Bettelstab:
Thank you for your initiative. While we think this stuff is nice material for a tutorial or possibly an own repo (which e.g. might include keyguard as submodule), we prefer to not include it here.

PR #302, opened 2019-05-07
Handle non-Login-File QR codes in imported files

### LINKED ISSUE #300: Handle non-decodable "Login File" QR-data
When `Nimiq.BufferUtils.fromBase64(decoded);` fails to decode, because the data my not be valid base64, the error is currently not handled but the whole request is rejected.

There needs to be handling for non-decodable strings and also checks if the data is actually a valid Nimiq secret.

sisou:
Sentry logged an error where the base64 decoding failed in BufferUtils.

Thus we found that we currently do not handle the case, where the imported file contains a valid QR code, but that QR code does not actually encode a Nimiq secret in the NIM-activation or ImageWallet format.

Fixes #300.

### DIFF
--- a/src/components/FileImporter.js
+++ b/src/components/FileImporter.js
@@ -4,6 +4,7 @@
 /* global I18n */
 /* global LoginFile */
 /* global TemplateTags */
+/* global KeyStore */
 
 class FileImporter extends Nimiq.Observable {
     /**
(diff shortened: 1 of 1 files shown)


src/components/FileImporter.js
```diff
@@ -90,6 +91,23 @@ class FileImporter extends Nimiq.Observable {
...
+        try {
+            // Make sure it is base64
+            const bytes = Nimiq.BufferUtils.fromBase64(decoded.substr(0, 2) === '#2' ? decoded.substr(2) : decoded);
+
+            // Make sure the data size is correct
+            if (bytes.byteLength !== KeyStore.ENCRYPTED_SECRET_SIZE
+                && bytes.byteLength !== KeyStore.ENCRYPTED_SECRET_SIZE_V2) {
+                throw new Error('Invalid encrypted secret size');
+            }
+        } catch (error) {
```
danimoh: Why via try ... catch?
What other exception are you expecting other than the one you're throwing yourself?

sisou: The one from the base64 decoder.

danimoh: Could have amended that comment to the previous commit, but ok

sisou: I'm squashing it on merge, ok? ;)

PR #303, opened 2019-05-07
Allow to the build script to detect a working version of sha256sum

cb5890:
Allow to the build script to detect a working version of `sha256sum` or at least fail with a productive error message.

Tested on the latest versions of MacOS and Fedora.

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -7,6 +7,21 @@ output() {
     echo "$(tput bold) $1$(tput sgr0)"
 }
 
+# Check for a working implementation of sha256sum
+SHA256SUM=""
+which sha256sum > /dev/null && SHA256SUM="sha256sum"
+# For (at least) MacPorts on MacOS, the GNU coreutils have been prefixed with 'g'
+# to be easily seperateable from the BSD ones
+[ -z "${SHA256SUM}" ] && which gsha256sum > /dev/null && SHA256SUM="gsha256sum"
+# Yes, that leaves the possibility that the current version of shasum doesn't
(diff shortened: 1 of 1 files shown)


tools/build.sh
```diff
@@ -7,6 +7,21 @@ output() {
...
+which sha256sum > /dev/null && SHA256SUM="sha256sum"
+# For (at least) MacPorts on MacOS, the GNU coreutils have been prefixed with 'g'
+# to be easily seperateable from the BSD ones
+[ -z "${SHA256SUM}" ] && which gsha256sum > /dev/null && SHA256SUM="gsha256sum"
+# Yes, that leaves the possibility that the current version of shasum doesn't
+# support the SHA256 algorithm. But I don't think that's rather common by now...
+[ -z "${SHA256SUM}" ] && which shasum > /dev/null && SHA256SUM="shasum -a 256"
+
+if [ -z "${SHA256SUM}" ]; then
+  output "Please install sha256sum or shasum first."
```
terorie: For the worst cases, there is also `openssl dgst -sha256`

sisou: @terorie can that also be used for checking files against known hashes?

terorie: Meh, I didn't expect `openssl` to have a different usage.
This is one possibility: `[ $(openssl dgst -sha256 -binary $file | xxd -p -c80) == $hash ]`.

terorie: From my experience, openssl has always been available on mac, but shasum is a better alternative

PR #312, opened 2019-05-11
Catch service worker installation error

Bettelstab:


src/common.js
```diff
@@ -19,6 +19,8 @@ if ('serviceWorker' in navigator) {
         scope: '../../',
     }).then(reg => {
         console.debug(`Service worker has been registered for scope: ${reg.scope}`);
+    }).catch(error => {
+        console.error(`Service worker installation failed: ${error}`);
```
danimoh: I think rethrowing would be good.

src/common.js
```diff
@@ -12,14 +13,19 @@
...
-if ('serviceWorker' in navigator) {
-    navigator.serviceWorker.register('../../ServiceWorker.js', {
-        scope: '../../',
-    }).then(reg => {
-        console.debug(`Service worker has been registered for scope: ${reg.scope}`);
-    });
+if (BrowserDetection.isIOS() || BrowserDetection.isSafari()) {
+    // Register service worker to strip cookie from requests
+    // This file is always called from a ./request/*/ folder, hence the paths.
+    if ('serviceWorker' in navigator) {
```
danimoh: This condition can be included in the outer `if`.

PR #322, opened 2019-05-12
FlippableHandler on reload

### LINKED ISSUE #319: Page disappears when reloading on 'flipped' side
When reloading the window on the "flipped" side (e.g. on the recovery words input page in the import request), the requests restarts on the initial page (`#import-file`), but the `.flipped` class gets still applied, hiding the initial page.

nibhar:
Adding a timeout should resolve #319

### DIFF
--- a/src/components/FlippableHandler.js
+++ b/src/components/FlippableHandler.js
@@ -30,7 +30,9 @@ class FlippableHandler {
                         window.setTimeout(() => $newEl.classList.remove('display-flex'), 600);
                         $oldEl.classList.add('display-flex');
                         window.setTimeout(() => $oldEl.classList.remove('display-flex'), 300);
-                        $rotationContainer.classList.toggle('flipped', $newEl.classList.contains(classname));
+                        window.setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/components/FlippableHandler.js
```diff
@@ -30,7 +30,9 @@ class FlippableHandler {
                         window.setTimeout(() => $newEl.classList.remove('display-flex'), 600);
                         $oldEl.classList.add('display-flex');
                         window.setTimeout(() => $oldEl.classList.remove('display-flex'), 300);
-                        $rotationContainer.classList.toggle('flipped', $newEl.classList.contains(classname));
+                        window.setTimeout(() => {
+                            $rotationContainer.classList.toggle('flipped', $newEl.classList.contains(classname));
+                        }, 0);
```
danimoh: As always: Comments are great.
And this could be written in one line, i.e. without opening a `{ }` block.

nibhar: It does not fit in one line if I strip the `{ }`. I choose to have it in this form as it reads better in my opinion. 
As all the toggles have the same setTimeout I don't think a comment was warranted.

PR #327, opened 2019-05-12
Curd/build improvements

cb5890:
* Make the build script less dependant on external tools. 
    => Except for git, it now works out of the box on a minimal Fedora.
* Fix build inaccuracy and (potential) errors on MacOS.
    => The build script should be able to detect and avoid differences between GNU and BSD tools to archive platform independence at least to some degree.

The build output before and after the modification remains identical on Fedora (most likely on Linux in general). After the modification, the output is now identical on MacOS too.

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -51,9 +51,16 @@ replace_font_url() {
     sed -i -r -e "s/$OLD_PATH/$NEW_PATH/g" $1
 }
 
+# Replace xxd -r -p with a nice bash function
+function hex_to_binary() {
+    for (( i=0; i<${#1}; i+=2 )); do
+        printf "\x${1:$i:2}"
+    done
+}
+
 # generate a base64 file integrity hash
 make_file_hash() {
-    echo $(${SHA256SUM} "$1" | awk '{print $1}' | xxd -r -p | base64)
+    hex_to_binary "$(${SHA256SUM} "$1" | cut -d' ' -f1)" | base64
 }
 
 # Before writing any files, verify integrity of Nimiq lib
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -9,13 +9,13 @@ output() {
 
 # Check for a working implementation of sha256sum
 SHA256SUM=""
-which sha256sum > /dev/null && SHA256SUM="sha256sum"
+command -v sha256sum > /dev/null && SHA256SUM="sha256sum"
 # For (at least) MacPorts on MacOS, the GNU coreutils have been prefixed with 'g'
 # to be easily seperateable from the BSD ones
-[ -z "${SHA256SUM}" ] && which gsha256sum > /dev/null && SHA256SUM="gsha256sum"
+[ -z "${SHA256SUM}" ] && command -v gsha256sum > /dev/null && SHA256SUM="gsha256sum"
 # Yes, that leaves the possibility that the current version of shasum doesn't
 # support the SHA256 algorithm. But I don't think that's rather common by now...
-[ -z "${SHA256SUM}" ] && which shasum > /dev/null && SHA256SUM="shasum -a 256"
+[ -z "${SHA256SUM}" ] && command -v shasum > /dev/null && SHA256SUM="shasum -a 256"
 
 if [ -z "${SHA256SUM}" ]; then
   output "Please install sha256sum or shasum first."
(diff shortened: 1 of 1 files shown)


sisou:
After testing, the new output still contains everything, but the order in the `common` and `toplevel` bundles are slightly different.

cb5890:
Yep, right. On the final test, I've already used `LC_ALL=C` in the Fedora container together with the old build script. Otherwise, I couldn't easily compare the files with MacOS. The order will stay the same for now on. 

You can see for yourself with `export LC_ALL=C` before running with the script from commit ` c3936d471badae269d633490ec4ecebd0303059a` ;-)

tools/build.sh
```diff
@@ -40,20 +60,28 @@ replace_icon_sprite_url() {
...
     NEW_PATH="\/assets\/fonts"
 
-    sed -i -r -e "s/$OLD_PATH/$NEW_PATH/g" $1
+    inplace_sed -E "s/$OLD_PATH/$NEW_PATH/g" $1
+}
+
+# Replace xxd -r -p with a nice bash function
+function hex_to_binary() {
+  for (( i=0; i<${#1}; i+=2 ))
+  do
```
sisou:
```suggestion
  for (( i=0; i<${#1}; i+=2 )); do
```

tools/build.sh
```diff
@@ -40,20 +60,28 @@ replace_icon_sprite_url() {
...
+      printf "\x${1:$i:2}"
+  done
 }
 
 # generate a base64 file integrity hash
 make_file_hash() {
-    echo $(${SHA256SUM} "$1" | awk '{print $1}' | xxd -r -p | base64)
+  hex_to_binary "$(${SHA256SUM} "$1" | cut -d' ' -f1)" | base64
```
sisou: Would it be possible to make the output non-implicit? E.g. make it so that `echo` is still required here to return the value from `make_file_hash`?
Also, indentation is off by a space here.

tools/build.sh
```diff
@@ -3,25 +3,45 @@
...
-[ -z "${SHA256SUM}" ] && which shasum > /dev/null && SHA256SUM="shasum -a 256"
+[ -z "${SHA256SUM}" ] && command -v shasum > /dev/null && SHA256SUM="shasum -a 256"
 
 if [ -z "${SHA256SUM}" ]; then
   output "Please install sha256sum or shasum first."
   exit 1
 fi
 
+# Detecting whether we're using GNU or BSD sed. There is a small but significant
+# idiosyncracy regarding the inplace option (-i).
```
sisou: `idiosyncracy` has negative connotations. I'd rather use `quirk` or simply `difference` to stay neutral here.

tools/build.sh
cb5890: But why? Is there a reason? :) There isn't a difference. That's simply inefficient and unnecessary... (And frankly speaking, it also looks really painful for a BASH programmer's eye.)

tools/build.sh
cb5890: Mh, actually, that was the intention and I don't think there is any reason to even think about that here.. That's the common stance regarding the subtle differences between GNU and BSD/UNIX tools. I wouldn't use the phrase "pain in the ass" though  ;-)

tools/build.sh
cb5890: Yep, you're right. I also noticed that, but didn't thought it's important enough to deserve dedicated attention given the other potential weak spots of the build script... :)

tools/build.sh
cb5890: Indentation is fixed. Also for the `hex_to_binary` function.

tools/build.sh
jeffesquivels: Hi guys, just my 2 cents as someone who has had to deal with the different versions of the same tools in different platforms: `idiosyncrasies` is indeed often used to describe these differences, take [this stackoverflow answer](https://unix.stackexchange.com/questions/401905/bsd-sed-vs-gnu-sed-and-i?rq=1), for example:

`Macs use the BSD version of utilities such as sed and date, which have their own idiosyncrasies.`

Having said that, I will be the first to admit that we may be [bikeshedding](https://www.urbandictionary.com/define.php?term=bikeshedding) a bit too much here :-).

PR #332, opened 2019-05-13
Remove width from .skip

### LINKED ISSUE #328: Skip-button shrinks too small on mobile
On thin screens, the 40% width available to the `.skip` link become too small, so that longer skip text breaks line:

image

Not limiting the width, but instead adding a left and right padding would be a better solution, I believe.

nibhar:
resolves #328

### DIFF
--- a/src/common.css
+++ b/src/common.css
@@ -204,6 +204,8 @@
 
 .page-footer {
     font-size: 2rem;
+    display: flex;
+    flex-direction: column;
 }
 
 .page:target {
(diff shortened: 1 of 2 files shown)


src/common.css
```diff
@@ -204,6 +204,8 @@
 
 .page-footer {
     font-size: 2rem;
+    display: flex;
+    flex-direction: column;
```
sisou: Can remove these styles from `Import.css` then

PR #333, opened 2019-05-13
Adapt RpcServer to include the id in the fragment instead of search.

nibhar:
In conjunction with nimiq/rpc#21

### DIFF
--- a/client/src/KeyguardClient.ts
+++ b/client/src/KeyguardClient.ts
@@ -59,10 +59,7 @@ export class KeyguardClient {
         this._redirectBehavior = new RedirectRequestBehavior(returnURL, localState, handleHistoryBack);
         this._iframeBehavior = new IFrameRequestBehavior();
 
-        // If this is a page-reload, allow location.origin as RPC origin
-        const allowedOrigin = document.referrer && new URL(document.referrer).origin === window.location.origin
-            ? window.location.origin
-            : RequestBehavior.getAllowedOrigin(this._endpoint);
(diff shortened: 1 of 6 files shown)


client/src/KeyguardClient.ts
```diff
@@ -59,10 +59,7 @@ export class KeyguardClient {
         this._redirectBehavior = new RedirectRequestBehavior(returnURL, localState, handleHistoryBack);
         this._iframeBehavior = new IFrameRequestBehavior();
 
-        // If this is a page-reload, allow location.origin as RPC origin
-        const allowedOrigin = document.referrer && new URL(document.referrer).origin === window.location.origin
-            ? window.location.origin
-            : RequestBehavior.getAllowedOrigin(this._endpoint);
+        const allowedOrigin = RequestBehavior.getAllowedOrigin(this._endpoint);
```
Bettelstab: Cool, why is it not needed anymore?

sisou: It was needed before because the request was always parsed from the URL before, even on reload. The url parser includes the check for the referrer. Thus when the page was reloaded or loaded by resubmitting the url, the request had to be loaded from url and the referrer test had to pass, otherwise it would have been an invalid request and rejected.

Now, the request is only parsed from the url on the very first page load, always passing the referrer check (if it is indeed a valid request). On every subsequent page load, either reload, url-submitting or other, the request is always loaded from the sessionstore. That loader does _not_ include a referrer check, because we don't have to. The request was found to be valid and stored the first time and will only be answered in that way. So no referrer check necessary again.

PR #334, opened 2019-05-13
Enable the back arrow in requests coming from Onboarding (Import, Create)

nibhar:


src/request/create/Create.js
```diff
@@ -36,9 +36,6 @@ class Create {
         /** @type {HTMLFormElement} */
         this.$walletIdentifier = (document.querySelector('.wallet-identifier'));
 
-        /** @type {HTMLLinkElement} */
-        this.$setPasswordBackButton = (this.$setPasswordPage.querySelector('a.page-header-back-button'));
```
sisou: Unused?

src/request/create/index.html
```diff
@@ -80,9 +80,6 @@
         <div id="choose-identicon" class="page nq-card">
             <div class="page-header nq-card-header">
                 <div class="progress-indicator"></div>
-                <a tabindex="0" class="page-header-back-button">
-                    <svg class="nq-icon"><use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-arrow-left"/></svg>
-                </a>
                 <h1 data-i18n="create-heading-choose-identicon" class="nq-h1">Choose an Avatar</h1>
```
sisou: But what if the request comes from the onboarding menu? Then a back button would be valid.

src/request/create/Create.js
nibhar: yes

src/request/create/index.html
nibhar: That is true. We do not have a back arrow for login requests, too.

nibhar:
Might as well come from LOGIN in case of CREATE or vice versa. So a more generally applicable name was chosen. 
Feel free to recommend a renaming.

sisou:
Hm, you are right. Ok, lets go with as it is now.

PR #337, opened 2019-05-14
Display transaction value excluding fee

### LINKED ISSUE #329: Should displayed tx value include fee?
I find this strange: I send 1 NIM. Now it says in big numbers "1.01242". I would expect it to rather be "1 NIM \n plus 0.01242 NIM fee".

screenshot_from_2019-05-12_17-06-34

sisou:
Fixes #329.

### DIFF
--- a/src/request/sign-transaction/SignTransaction.js
+++ b/src/request/sign-transaction/SignTransaction.js
@@ -71,11 +71,7 @@ class SignTransaction {
         const $data = (this.$el.querySelector('#data'));
 
         // Set value and fee.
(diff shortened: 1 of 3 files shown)


sisou:
The fee will mostly be very small (thus not as important) and just add noise to the amount (thus visually bad). I also like it better when the amount that I input into the value input when sending a transaction is displayed there (especially for newbies who don't really know how fees work and then would suddenly see a different value there than they entered), or when the value displayed here is exactly the amount that I saw in the shop or app before clicking on checkout.

PR #342, opened 2019-05-16
Mobile Layout

sisou:
This adds CSS styles for a simple mobile layout where the page is sticking to the bottom of the screen and there are no margins left and right. The global close button is at the top right next to the Nimiq logo. 

Sadly there are still scroll bars on mobile, or rather a couple-pixel wide gap both on the right and bottom of the screen, which I can't explain nor get rid of. If anybody has ideas, please try!

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   "devDependencies": {
     "@nimiq/core-web": "1.4.3",
     "@nimiq/rpc": "^0.3.0",
-    "@nimiq/style": "^0.7.0",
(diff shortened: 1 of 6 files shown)


sisou:
As far as I know, when the flippable container is present the `#app` gets a `perspective: 250rem` style, which when removed fixes the scrollbars on mobile, but also kills the perspective (who would have thought :D).

In requests without it (like create), the scrollbars are still there though.

PR #344, opened 2019-05-16
PasswordBox shake instead of PasswordInput shake

nibhar:


sisou:
I would switch the behavior or the password setter to shake/red.

Instead of shaking when too short and doing nothing when too long, I would make it do nothing when too short but shake when too long.

Reason being that when I have a long password, I may take a break in input because I need to look up the next part, and my password-entering may not yet be finished. Then it makes no sense to shake the field as if it were a wrong input. It's not wrong, it's just not complete.

On the other hand when I repeat a longer password than before, it is definitely wrong.

PR #348, opened 2019-05-17
Replace loading-spinner sprites by the real SVG

sisou:
Because Safari cannot target SVG sprites' shadow-DOM with CSS selectors, prohibiting the loader animation.

With this change, the loading spinner code is only added 2 times more than it was before (in the PasswordBox and PasswordSetterBox respectively). There are no other extra occurances, so the cost to making the animatin work in Safari/iOS is very low.

### DIFF
--- a/src/components/PasswordBox.js
+++ b/src/components/PasswordBox.js
@@ -82,7 +82,7 @@ class PasswordBox extends Nimiq.Observable {
             <div password-input></div>
             ${buttonVersions[options.buttonI18nTag]}
             <!-- Loading spinner SVG -->
-            <svg height="48" width="54" color="inherit" class="loading-spinner"><g id="loading-spinner">
+            <svg height="48" width="54" color="inherit" class="loading-spinner"><g>
                 <path class="big-hex" d="M51.9,21.9L41.3,3.6c-0.8-1.3-2.2-2.1-3.7-2.1H16.4c-1.5,0-2.9,0.8-3.7,2.1L2.1,21.9c-0.8,1.3-0.8,2.9,0,4.2 l10.6,18.3c0.8,1.3,2.2,2.1,3.7,2.1h21.3c1.5,0,2.9-0.8,3.7-2.1l10.6-18.3C52.7,24.8,52.7,23.2,51.9,21.9z" stroke="currentColor" stroke-width="3" fill="none" stroke-linecap="round" opacity="0.4" stroke-dasharray="92.5 60"/>
                 <path class="small-hex" d="M51.9,21.9L41.3,3.6c-0.8-1.3-2.2-2.1-3.7-2.1H16.4c-1.5,0-2.9,0.8-3.7,2.1L2.1,21.9c-0.8,1.3-0.8,2.9,0,4.2 l10.6,18.3c0.8,1.3,2.2,2.1,3.7,2.1h21.3c1.5,0,2.9-0.8,3.7-2.1l10.6-18.3C52.7,24.8,52.7,23.2,51.9,21.9z" stroke="currentColor" stroke-width="3" fill="none" stroke-linecap="round" stroke-dasharray="47.5 105"/>
             </g></svg>
--- a/src/components/PasswordBox.js
+++ b/src/components/PasswordBox.js
@@ -71,6 +71,7 @@ class PasswordBox extends Nimiq.Observable {
 
         if (!buttonVersions[options.buttonI18nTag]) throw new Error('PasswordBox button i18n tag not defined');
 
+        /* eslint-disable max-len */
         $el.innerHTML = TemplateTags.hasVars(1)`
             <a class="cancel">
                 <svg class="nq-icon">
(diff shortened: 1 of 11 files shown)


src/components/PasswordBox.js
```diff
@@ -80,10 +81,13 @@ class PasswordBox extends Nimiq.Observable {
...
             <div password-input></div>
             ${buttonVersions[options.buttonI18nTag]}
-            <svg height="48" width="54" color="inherit" class="loading-spinner">
-                <use xlink:href="#loading-spinner" />
-            </svg>
+            <!-- Loading spinner SVG -->
+            <svg height="48" width="54" color="inherit" class="loading-spinner"><g id="loading-spinner">
+                <path class="big-hex" d="M51.9,21.9L41.3,3.6c-0.8-1.3-2.2-2.1-3.7-2.1H16.4c-1.5,0-2.9,0.8-3.7,2.1L2.1,21.9c-0.8,1.3-0.8,2.9,0,4.2 l10.6,18.3c0.8,1.3,2.2,2.1,3.7,2.1h21.3c1.5,0,2.9-0.8,3.7-2.1l10.6-18.3C52.7,24.8,52.7,23.2,51.9,21.9z" stroke="currentColor" stroke-width="3" fill="none" stroke-linecap="round" opacity="0.4" stroke-dasharray="92.5 60"/>
+                <path class="small-hex" d="M51.9,21.9L41.3,3.6c-0.8-1.3-2.2-2.1-3.7-2.1H16.4c-1.5,0-2.9,0.8-3.7,2.1L2.1,21.9c-0.8,1.3-0.8,2.9,0,4.2 l10.6,18.3c0.8,1.3,2.2,2.1,3.7,2.1h21.3c1.5,0,2.9-0.8,3.7-2.1l10.6-18.3C52.7,24.8,52.7,23.2,51.9,21.9z" stroke="currentColor" stroke-width="3" fill="none" stroke-linecap="round" stroke-dasharray="47.5 105"/>
+            </g></svg>
```
danimoh: Suggestion: Could also clone the svg from the page to not have to repeat it here and save some bytes.

src/components/PasswordBox.js
```diff
@@ -80,10 +81,13 @@ class PasswordBox extends Nimiq.Observable {
             <div class="prompt nq-text-s" data-i18n="passwordbox-enter-password">Enter your password</div>
             <div password-input></div>
             ${buttonVersions[options.buttonI18nTag]}
-            <svg height="48" width="54" color="inherit" class="loading-spinner">
-                <use xlink:href="#loading-spinner" />
-            </svg>
+            <!-- Loading spinner SVG -->
+            <svg height="48" width="54" color="inherit" class="loading-spinner"><g id="loading-spinner">
```
danimoh: Remove double id `loading-spinner`

src/components/PasswordSetterBox.js
```diff
@@ -65,9 +65,11 @@ class PasswordSetterBox extends Nimiq.Observable {
...
             <button class="submit" data-i18n="passwordbox-repeat">Repeat password</button>
 
-            <svg height="48" width="54" color="inherit" class="loading-spinner">
-                <use xlink:href="#loading-spinner" />
-            </svg>
+            <!-- Loading spinner SVG -->
+            <svg height="48" width="54" color="inherit" class="loading-spinner"><g id="loading-spinner">
+                <path class="big-hex" d="M51.9,21.9L41.3,3.6c-0.8-1.3-2.2-2.1-3.7-2.1H16.4c-1.5,0-2.9,0.8-3.7,2.1L2.1,21.9c-0.8,1.3-0.8,2.9,0,4.2 l10.6,18.3c0.8,1.3,2.2,2.1,3.7,2.1h21.3c1.5,0,2.9-0.8,3.7-2.1l10.6-18.3C52.7,24.8,52.7,23.2,51.9,21.9z" stroke="currentColor" stroke-width="3" fill="none" stroke-linecap="round" opacity="0.4" stroke-dasharray="92.5 60"/>
+                <path class="small-hex" d="M51.9,21.9L41.3,3.6c-0.8-1.3-2.2-2.1-3.7-2.1H16.4c-1.5,0-2.9,0.8-3.7,2.1L2.1,21.9c-0.8,1.3-0.8,2.9,0,4.2 l10.6,18.3c0.8,1.3,2.2,2.1,3.7,2.1h21.3c1.5,0,2.9-0.8,3.7-2.1l10.6-18.3C52.7,24.8,52.7,23.2,51.9,21.9z" stroke="currentColor" stroke-width="3" fill="none" stroke-linecap="round" stroke-dasharray="47.5 105"/>
+            </g></svg>
```
danimoh: Same as in `PasswordBox.js`

src/components/PasswordBox.js
sisou: When you say clone, you mean look for the page element and clone-append it here during runtime?

src/components/PasswordBox.js
sisou: What do you mean? In this line one is a class and one is an id. Or do you mean that side-wide it's a double ID because of the one on the page already?
It still works as it is, however, because browsers are lenient to such 'errors'.

danimoh: Yes, I meant the duplicate id on the page.
I know that browser can handle such errors.
Still, id's should be unique.

src/components/PasswordBox.js
danimoh: Yes, clone the node and append it.
But really whatever you see fit.

src/components/PasswordBox.js
nibhar: I agree with @danimoh. 
I think the id is obsolete anyways since the spinner is no longer referenced by a use tag.

src/components/PasswordBox.js
sisou: Not doing that to keep it simple, and because it only repeats 2 times in the whole code base, so not too bad.

PR #351, opened 2019-05-18
Export words button

nibhar:
The button is hidden until scrolled down. In its place an explanatory text is displayed instead.


src/request/export/Export.css
```diff
@@ -51,6 +51,33 @@ ul.nq-list li > .nq-icon {
...
+.page#recovery-words .page-footer.scrolled-down p {
+    z-index: -1;
+    opacity: 0;
+}
+
+.page#recovery-words .page-footer button {
+    z-index: -1;
+    position: relative;
+    opacity: 0;
+    transition: opacity .4s ease;
```
sisou: You need to include the existing transitions from @nimiq/style in this `transition` property, otherwise e.g. the transform has no transition anymore.

PR #352, opened 2019-05-19
Return derived addresses from derive-address request directly

sisou:
Without identicon-selection UI, as that is now happening in the Hub.

Since there are changes in the public types, this PR creates a new Keyguard Client version.


src/request/derive-address/DeriveAddress.js
```diff
@@ -117,51 +56,49 @@ class DeriveAddress {
...
-        // last space is necessary for the rendering to work properly with white-space: pre-wrap.
-        this.$address.textContent = `${address} `;
-
-        this.$chooseIdenticonPage.classList.add('account-details-open');
-    }
+    _generateAddresses(masterKey, pathsToDerive) {
+        /** @type { KeyguardRequest.DerivedAddress[] } */
+        const derivedAddresses = [];
+
+        for (let i = 0; i < pathsToDerive.length; i++) {
```
nibhar: could be 
```suggestion
        for (const path of pathsToDerive) {
```
and then replace the `pathsToDerive[i]` by `path`.

To me personally that reads better.

client/src/PublicRequest.ts
```diff
@@ -143,7 +143,7 @@ export type LegacyKeyInfoObject = KeyInfoObject & {
 
 // Specific Results
 
-export type DeriveAddressResult = {
+export type DerivedAddress = {
```
danimoh: Can `DeriveAddressesResult` be rewritten to use this?

src/request/derive-address/DeriveAddress.js
```diff
@@ -22,86 +20,27 @@ class DeriveAddress {
...
-            /** @type {KeyguardRequest.DeriveAddressResult} */
-            const result = {
-                address: this._selectedAddress.address.serialize(),
-                keyPath: this._selectedAddress.keyPath,
-            };
-
-            this._resolve(result);
+        this._passwordBox.on(PasswordBox.Events.SUBMIT, /** @param {string|undefined} password */ password => {
+            this._onPasswordEntered(password);
         });
```
danimoh: No need for the `{ }` code block.

sisou: When I omit the parantheses, eslint expects a @returns type annotation, because the result of the `_onPasswordEntered` method is then returned.

sisou: The alternative is to omit the whole JSDoc block, then ESLint is fine. But then `password` is `any`.

client/src/PublicRequest.ts
```diff
@@ -173,14 +171,14 @@ export type Result = RedirectResult | IFrameResult;
 
 export type ResultType<T extends RedirectRequest> =
     T extends Is<T, SignMessageRequest> | Is<T, SignTransactionRequest> ? SignatureResult :
-    T extends Is<T, DeriveAddressRequest> ? Is<T, DeriveAddressResult> :
+    T extends Is<T, DeriveAddressRequest> ? Is<T, DerivedAddress[]> :
```
Bettelstab:
```suggestion
    T extends Is<T, DeriveAddressRequest> ? DerivedAddress[] :
```
That was my fault, but good opportunity to sneak in and fix it.

PR #354, opened 2019-05-19
Various style improvements

sisou:


src/components/DownloadLoginFile.js
```diff
@@ -104,12 +107,12 @@ class DownloadLoginFile extends Nimiq.Observable {
...
             // Setup native download
-            this.$el.href = href;
-            this.$el.download = filename;
+            this.$downloadButton.href = href;
+            this.$downloadButton.download = filename;
         } else {
             // Setup fallback download
             // Hack to make image downloadable on iOS via long tap.
-            this.$el.href = 'javascript:void(0);'; // eslint-disable-line no-script-url
+            // this.$downloadButton.href = 'javascript:void(0);'; // eslint-disable-line no-script-url
```
nibhar: no longer needed?

sisou: It was. Haven't had time or the iPhone to test it before. Fixed it.

src/components/DownloadLoginFile.js
```diff
@@ -33,23 +39,23 @@ class DownloadLoginFile extends Nimiq.Observable {
...
      */
     static _createElement($el) {
-        $el = $el || document.createElement('a');
+        $el = $el || document.createElement('div');
         $el.classList.add('download-loginfile');
 
         /* eslint-disable max-len */
         $el.innerHTML = TemplateTags.noVars`
-            <img class="loginfile" src=""></img>
+            <a class="loginfile-link"><img class="loginfile" src=""></img></a>
```
danimoh: What is that anchor needed for?
I think iOS download should work without the anchor too.

src/components/DownloadLoginFile.js
```diff
@@ -131,31 +137,14 @@ class DownloadLoginFile extends Nimiq.Observable {
     _onMouseDown(event) {
         /** @type {HTMLElement} */
         const target = (event.target);
-        // A click on the Continue button is already covered by a 'click' handler.
-        if (target.matches('.continue')) return;
+        // Clicks on the continue or download buttons are already covered by a 'click' handler.
+        if (target.matches('.continue') || target.matches('.download-button')) return;
```
danimoh: Still needed now that the mousedown handler is only bound to the login file?

src/components/DownloadLoginFile.js
sisou: The element itself (`this.$el`) is no longer a link, but for iOS long-touch download to work, the image needs to be wrapped in a link with a `href` set. That's what this anchor is for.

src/components/DownloadLoginFile.js
sisou: Probably not.

PR #355, opened 2019-05-19
Resolve when cancelling words-export after file was exported

sisou:
When the user cancels the export-words action after the Login File was downloaded, this new hook will now _resolve_ the request instead of cancelling it.

### DIFF
--- a/src/request/TopLevelApi.js
+++ b/src/request/TopLevelApi.js
@@ -32,10 +32,9 @@
  *          // The return value is not used.
  *      }
  *
- *      async onBeforeCancel(handler) {
- *          // This optional method receives the instantiated handler and returns if the handling
- *          // of the user's click on the global-cancel button should be halted.
- *          // Returning boolean<true> stops the handling.
+ *      async onGlobalClose(handler) {
+ *          // Handle the user's click on the global-close button.
+ *          // This method receives the instantiated handler.
  *      }
  *  }
  *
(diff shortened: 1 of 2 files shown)


sisou:
Yes, that would also work. I decided to keep with the _hooks_ paradigm for optional overwrites.

Because what we currently have in the `TopLevelApi` are two methods which are meant to be overwritten by extenders: the `parseRequest` method, which *must* be overloaded, and the `onBeforeRun`, which is an optional overload. Since the interception of the global-close button is also optional, I chose to go with the established pattern of a `onBeforeCancel` hook.

nibhar:
call it `onGlobalClose(handler)` then? I think my proposal is a simpler, cleaner solution.

I would argue, that `onBeforeRun()` could be done in a similar fashion, as it is only ever used once to set the text of the global close button.

sisou:
It's exactly the same solution, just instead of checking for the return value in `TopLevelApi`, you have to remember call `super.onGlobalClose()` in the child.

nibhar:
Gets rid of `onBeforeCancel` though, which basically is `globalClose()`. Just like `onBeforeRun` basically is `setGlobalCloseButtonText`.

nibhar:
Also in both cases the functions exist to alter the default behaviour. They do not exist to perform some action and then return to the default behaviour. 

I think in case of altering the behaviour of the base class my approach is preferable. In case we would want to augment it a hook is the better choice.

sisou:
Your argument that it gets rid of `onBeforeCancel` is void, because in its place you are adding the `onGlobalClose` method. Thus it's exactly the same thing and the same amount of code change.

Aber der Klügere gibt nach.

PR #358, opened 2019-05-21
Use patched version of @nimiq/core-web to force JS worker on iOS

sisou:


src/common.js
```diff
@@ -16,11 +16,9 @@ let __nimiqLoaded = null;
 if ((BrowserDetection.isIOS() || BrowserDetection.isSafari()) && 'serviceWorker' in navigator) {
     // Register service worker to strip cookie from requests.
     // This file is always called from a ./request/*/ folder, hence the paths.
-    navigator.serviceWorker.register('../../ServiceWorker.js', {
-        scope: '../../',
-    }).then(reg => {
+    navigator.serviceWorker.register('/ServiceWorker.js').then(reg => {
```
nibhar: Screen Shot 2019-05-21 at 13 10 52
I don't think that works as intended.

nibhar:
```suggestion
    navigator.serviceWorker.register('/src/ServiceWorker.js').then(reg => {
```

sisou: ah, it fails locally... :(

sisou: Ok, I'll remove the registration adaptions

src/common.js
```diff
@@ -35,7 +35,7 @@ async function loadNimiq() {
     // eslint-disable-next-line no-return-assign
     return __nimiqLoaded || (__nimiqLoaded = new Promise(async resolve => {
         // Load web assembly encryption library into browser (if supported)
-        await Nimiq.WasmHelper.doImport();
+        await Nimiq.WasmHelper.doImport(BrowserDetection.isIOS());
```
Bettelstab:
```suggestion
        await Nimiq.WasmHelper.doImport(BrowserDetection.isIOS() && !window.opener);
```

PR #367, opened 2019-06-17
Make file hashes dependent on file content, not git hash

sisou:
Currently, whenever the Keyguard is built from a new git commit, the file names for all JS and CSS files change (because they are taken from the git commit hash).

To reduce the need for browsers to re-download unchanged files just because their file names changed, this PR makes the file name hash dependent on the file contents. This prevents that a file name is changed when the file itself did not change. This will prevent (1) unnecessary git change sets and (2) unnecessary file re-downloads by browsers.

This PR was enabled by the restructuring of the build script in #280, which split the collection and linking of JS/CSS bundles into two steps. This is required here, because the file hash can only be calculated once the bundle is complete.

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -63,6 +63,17 @@ replace_icon_sprite_url() {
     inplace_sed "s/$OLD_PATH/$NEW_PATH/g" $1
 }
 
+# add file hash to file name and return new file name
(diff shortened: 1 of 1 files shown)


danimoh (approved):
I think this looks ok

PR #368, opened 2019-07-12
Add tab width dropdown selector

MatthewDLudwig:
- Only appears when message is a string containing a tab.
 - Options are 1, 2, 4, and 8 defaulting to 8.
 - Choice is saved in localStorage and recalled when signing in the future.

### DIFF
--- a/src/request/sign-message/SignMessage.css
+++ b/src/request/sign-message/SignMessage.css
@@ -114,7 +114,24 @@
     outline: none;
     width: 100%;
     padding: 1.5rem;
-    flex-grow: 1;
+    flex-grow: 4;
     resize: none;
     font-family: 'Fira Mono', monospace;
+}
+
+#tabdiv {
+    display: flex;
+    flex-direction: row:
+    flex-grow: 1
+}
+
+#tabtext {
+    margin-left: 1rem;
+    flex-grow: 2;
+}
+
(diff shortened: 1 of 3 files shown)


MatthewDLudwig:
I had forgotten to run `yarn pr` prior to my PR.  Seems 3 errors occur which I assume is what's causing Travis to fail.  I've fixed one with my most recent commit but the remaining 2 relate to `tabSize` not being part of `CSSStyleDeclaration`.  My guess is that it's not in CSSStyleDeclaration due to still being ["experimental"](https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size) and not [fully cross browser compatible](https://caniuse.com/#search=tab-size).  

Are we willing to work around this (and if so, what is the workaround) or should I close the PR?

sisou:
Thanks a lot, this is great! :+1:

@julianbauer it currently looks like this. Do you have input for how the selector can be better integrated into the UI? I was thinking to position inside the top right corner of the text area.

Screenshot_20190712_081707

sisou:
Julian designed this layout:

Frame 5

sisou:
Replaced by #391, which includes the commits from this PR.

danimoh:
Your changes have been merged.
Thanks for the great contribution :)

PR #373, opened 2019-09-12
Checkout adaptions

danimoh:
Adaptions for new hub multi currency checkout features.

### DIFF
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -93,13 +93,23 @@ export type ExportResult = {
     wordsExported: boolean,
 };
 
+export type CurrencyInfo = {
+    code: string,
+    digits: number,
+};
+
 export type SignTransactionRequest = SimpleRequest & TransactionInfo & {
     keyPath: string,
     layout?: SignTransactionRequestLayout,
     recipientLabel?: string,
     senderLabel?: string,
+    // TODO these should only get added for 'checkout' layout
     shopOrigin?: string,
     shopLogoUrl?: string,
+    time?: number,
+    expires?: number,
+    fiatCurrency?: CurrencyInfo,
+    fiatAmount?: number,
 };
 
 export type SignMessageRequest = SimpleRequest & {
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -206,6 +206,8 @@ export type KeyguardError = {
         GOTO_CREATE: 'GOTO_CREATE',
         // used to signal a user initiated cancelation of the request
         CANCELED: 'CANCELED',
+        // used to signal that the request expired
+        EXPIRED: 'EXPIRED',
         // used to signal that a given keyId no longer exist in KG, to be treated by caller.
         KEY_NOT_FOUND: 'keyId not found',
         // network name does not exist
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -93,19 +93,24 @@ export type ExportResult = {
     wordsExported: boolean,
 };
 
-export type SignTransactionRequest = SimpleRequest & TransactionInfo & {
+export type SignTransactionRequestSimple = SimpleRequest & TransactionInfo & {
+    layout?: Exclude<SignTransactionRequestLayout, 'checkout'>,
     keyPath: string,
-    layout?: SignTransactionRequestLayout,
     recipientLabel?: string,
     senderLabel?: string,
-    // TODO these should only get added for 'checkout' layout
-    shopOrigin?: string,
+};
+
+export type SignTransactionRequestCheckout = Omit<SignTransactionRequestSimple, 'layout'> & {
+    layout: 'checkout'
+    shopOrigin: string,
     shopLogoUrl?: string,
     time?: number,
     expires?: number,
     fiatCurrency?: string,
     fiatAmount?: number,
 };
+
+export type SignTransactionRequest = SignTransactionRequestSimple | SignTransactionRequestCheckout;
 
 export type SignMessageRequest = SimpleRequest & {
     keyPath: string,
@@ -175,18 +180,21 @@ export type Result = RedirectResult | IFrameResult;
 // Derived Result types
 
 export type ResultType<T extends RedirectRequest> =
-    T extends Is<T, SignMessageRequest> | Is<T, SignTransactionRequest> ? SignatureResult :
+    T extends Is<T, SignMessageRequest> | Is<T, SignTransactionRequestCheckout> | Is<T, SignTransactionRequestSimple>
+        ? SignatureResult :
     T extends Is<T, DeriveAddressRequest> ? DerivedAddress[] :
     T extends Is<T, CreateRequest> | Is<T, ImportRequest> ? KeyResult :
     T extends Is<T, ExportRequest> ? ExportResult :
-    T extends Is<T, RemoveKeyRequest> | Is<T, SimpleRequest> ? SimpleResult : never;
+    T extends Is<T, RemoveKeyRequest> | Is<T, SimpleRequest> ? SimpleResult :
+    never;
 
 export type ResultByCommand<T extends KeyguardCommand> =
     T extends KeyguardCommand.SIGN_MESSAGE | KeyguardCommand.SIGN_TRANSACTION ? SignatureResult :
     T extends KeyguardCommand.DERIVE_ADDRESS ? DerivedAddress[] :
     T extends KeyguardCommand.CREATE | KeyguardCommand.IMPORT ? KeyResult :
     T extends KeyguardCommand.EXPORT ? ExportResult :
-    T extends KeyguardCommand.REMOVE ? SimpleResult : never;
+    T extends KeyguardCommand.REMOVE ? SimpleResult :
+    never;
 
 // Error constants
 
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -93,11 +93,6 @@ export type ExportResult = {
     wordsExported: boolean,
 };
 
-export type CurrencyInfo = {
-    code: string,
-    digits: number,
-};
-
 export type SignTransactionRequest = SimpleRequest & TransactionInfo & {
     keyPath: string,
     layout?: SignTransactionRequestLayout,
@@ -108,7 +103,7 @@ export type SignTransactionRequest = SimpleRequest & TransactionInfo & {
     shopLogoUrl?: string,
     time?: number,
     expires?: number,
-    fiatCurrency?: CurrencyInfo,
+    fiatCurrency?: string,
     fiatAmount?: number,
 };
 
--- a/src/components/PaymentInfoLine.js
+++ b/src/components/PaymentInfoLine.js
@@ -3,7 +3,7 @@
 /* global TemplateTags */
 /* global NumberFormatting */
 
-class PaymentInfoLine extends Nimiq.Observable {
+class PaymentInfoLine { // eslint-disable-line no-unused-vars
     /**
      * @param {object}           paymentInfo
      * @param {string}           paymentInfo.recipient
@@ -17,7 +17,6 @@ class PaymentInfoLine extends Nimiq.Observable {
      * @param {HTMLElement}      [$el]
      */
     constructor(paymentInfo, $el) {
-        super();
         this.$el = PaymentInfoLine._createElement($el);
 
         /** @type HTMLElement */
@@ -39,14 +38,7 @@ class PaymentInfoLine extends Nimiq.Observable {
             imageUrl: paymentInfo.imageUrl || null,
             accountLabel: null,
         });
-        recipientInfo.renderTo(/** @type HTMLElement */ (this.$el.querySelector('.recipient-info')));
-
-        /** @type HTMLElement */
-        const $recipient = (this.$el.querySelector('.recipient'));
-        $recipient.addEventListener('click', event => {
-            event.preventDefault(); // to avoid CSP warnings about the javascript:void(0) href
-            this.fire(PaymentInfoLine.Events.RECIPIENT_CLICKED);
-        });
+        recipientInfo.renderTo(/** @type HTMLElement */ (this.$el.querySelector('.recipient')));
     }
 
     /**
@@ -68,20 +60,9 @@ class PaymentInfoLine extends Nimiq.Observable {
                     <use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-arrow-right-small"/>
                 </svg>
             </div>
-            <a href="javascript:void(0)" class="recipient">
-                <div class="recipient-info"></div>
-                <div class="info-circle-container">
-                    <svg class="nq-icon">
-                        <use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-info-circle"/>
-                    </svg>
-                </div>
-            </a>
+            <div class="recipient"></div>
         `;
 
         return $el;
     }
 }
-
-PaymentInfoLine.Events = {
-    RECIPIENT_CLICKED: 'recipient-clicked',
-};
--- a/src/components/PaymentInfoLine.js
+++ b/src/components/PaymentInfoLine.js
@@ -4,18 +4,21 @@
 /* global TemplateTags */
 /* global NumberFormatting */
 
+/** @typedef {{
+ *      recipient: string,
+ *      label?: string,
+ *      imageUrl?: URL,
+ *      lunaAmount: number,
+ *      fiatAmount?: number,
+ *      fiatCurrency?: string,
+ *      time?: number,
+ *      expires?: number
+ *  }} PaymentInfo */
+
 class PaymentInfoLine { // eslint-disable-line no-unused-vars
     /**
-     * @param {object}           paymentInfo
-     * @param {string}           paymentInfo.recipient
-     * @param {string|undefined} [paymentInfo.label]
-     * @param {URL|undefined}    [paymentInfo.imageUrl]
-     * @param {number}           paymentInfo.lunaAmount
-     * @param {number|undefined} [paymentInfo.fiatAmount]
-     * @param {string|undefined} [paymentInfo.fiatCurrency]
-     * @param {number|undefined} [paymentInfo.time]
-     * @param {number|undefined} [paymentInfo.expires]
-     * @param {HTMLElement}      [$el]
+     * @param {PaymentInfo} paymentInfo
+     * @param {HTMLElement} [$el]
      */
     constructor(paymentInfo, $el) {
         this.$el = PaymentInfoLine._createElement($el);
@@ -27,7 +30,7 @@ class PaymentInfoLine { // eslint-disable-line no-unused-vars
 
         /** @type HTMLElement */
         const $fiatAmount = (this.$el.querySelector('.fiat-amount'));
-        if (paymentInfo.fiatAmount !== undefined && paymentInfo.fiatCurrency !== undefined) {
+        if (paymentInfo.fiatAmount && paymentInfo.fiatCurrency) {
             $fiatAmount.textContent = NumberFormatting.formatCurrency(paymentInfo.fiatAmount, paymentInfo.fiatCurrency);
         } else {
             $fiatAmount.remove();
@@ -43,7 +46,7 @@ class PaymentInfoLine { // eslint-disable-line no-unused-vars
 
         /** @type HTMLElement */
         const $timer = (this.$el.querySelector('.timer'));
-        if (paymentInfo.time !== undefined && paymentInfo.expires !== undefined) {
+        if (paymentInfo.time && paymentInfo.expires) {
             new Timer(paymentInfo.time, paymentInfo.expires, $timer); // eslint-disable-line no-new
         } else {
             $timer.remove();
--- a/src/components/PaymentInfoLine.js
+++ b/src/components/PaymentInfoLine.js
@@ -1,5 +1,6 @@
 /* global Nimiq */
 /* global AddressInfo */
+/* global Timer */
 /* global TemplateTags */
 /* global NumberFormatting */
 
@@ -39,6 +40,14 @@ class PaymentInfoLine { // eslint-disable-line no-unused-vars
             accountLabel: null,
         });
         recipientInfo.renderTo(/** @type HTMLElement */ (this.$el.querySelector('.recipient')));
+
+        /** @type HTMLAnchorElement */
+        const $timer = (this.$el.querySelector('.timer'));
+        if (paymentInfo.time !== undefined && paymentInfo.expires !== undefined) {
+            new Timer(paymentInfo.time, paymentInfo.expires, $timer); // eslint-disable-line no-new
+        } else {
+            $timer.remove();
+        }
     }
 
     /**
@@ -61,6 +70,7 @@ class PaymentInfoLine { // eslint-disable-line no-unused-vars
                 </svg>
             </div>
             <div class="recipient"></div>
+            <a class="timer"></a>
         `;
 
         return $el;
--- a/src/components/PaymentInfoLine.js
+++ b/src/components/PaymentInfoLine.js
@@ -5,16 +5,16 @@
 
 class PaymentInfoLine extends Nimiq.Observable {
     /**
-     * @param {object}                                 paymentInfo
-     * @param {string}                                 paymentInfo.recipient
-     * @param {string|undefined}                       [paymentInfo.label]
-     * @param {URL|undefined}                          [paymentInfo.imageUrl]
-     * @param {number}                                 paymentInfo.lunaAmount
-     * @param {number|undefined}                       [paymentInfo.fiatAmount]
-     * @param {KeyguardRequest.CurrencyInfo|undefined} [paymentInfo.fiatCurrency]
-     * @param {number|undefined}                       [paymentInfo.time]
-     * @param {number|undefined}                       [paymentInfo.expires]
-     * @param {HTMLElement}                            [$el]
+     * @param {object}           paymentInfo
+     * @param {string}           paymentInfo.recipient
+     * @param {string|undefined} [paymentInfo.label]
+     * @param {URL|undefined}    [paymentInfo.imageUrl]
+     * @param {number}           paymentInfo.lunaAmount
+     * @param {number|undefined} [paymentInfo.fiatAmount]
+     * @param {string|undefined} [paymentInfo.fiatCurrency]
+     * @param {number|undefined} [paymentInfo.time]
+     * @param {number|undefined} [paymentInfo.expires]
+     * @param {HTMLElement}      [$el]
      */
     constructor(paymentInfo, $el) {
         super();
@@ -28,8 +28,7 @@ class PaymentInfoLine extends Nimiq.Observable {
         /** @type HTMLElement */
         const $fiatAmount = (this.$el.querySelector('.fiat-amount'));
         if (paymentInfo.fiatAmount !== undefined && paymentInfo.fiatCurrency !== undefined) {
-            const fiatAmount = NumberFormatting.formatNumber(paymentInfo.fiatAmount, paymentInfo.fiatCurrency.digits);
-            $fiatAmount.textContent = `${fiatAmount} ${paymentInfo.fiatCurrency.code}`;
+            $fiatAmount.textContent = NumberFormatting.formatCurrency(paymentInfo.fiatAmount, paymentInfo.fiatCurrency);
         } else {
             $fiatAmount.remove();
         }
--- a/src/components/PaymentInfoLine.js
+++ b/src/components/PaymentInfoLine.js
@@ -1,0 +1,88 @@
+/* global Nimiq */
+/* global AddressInfo */
+/* global TemplateTags */
+/* global NumberFormatting */
+
+class PaymentInfoLine extends Nimiq.Observable {
+    /**
+     * @param {object}                                 paymentInfo
+     * @param {string}                                 paymentInfo.recipient
+     * @param {string|undefined}                       [paymentInfo.label]
+     * @param {URL|undefined}                          [paymentInfo.imageUrl]
+     * @param {number}                                 paymentInfo.lunaAmount
+     * @param {number|undefined}                       [paymentInfo.fiatAmount]
+     * @param {KeyguardRequest.CurrencyInfo|undefined} [paymentInfo.fiatCurrency]
+     * @param {number|undefined}                       [paymentInfo.time]
+     * @param {number|undefined}                       [paymentInfo.expires]
+     * @param {HTMLElement}                            [$el]
+     */
+    constructor(paymentInfo, $el) {
+        super();
+        this.$el = PaymentInfoLine._createElement($el);
+
+        /** @type HTMLElement */
+        const $nimAmount = (this.$el.querySelector('.nim-amount'));
+        const nimAmount = NumberFormatting.formatNumber(Nimiq.Policy.lunasToCoins(paymentInfo.lunaAmount));
+        $nimAmount.textContent = `${nimAmount} NIM`;
+
+        /** @type HTMLElement */
+        const $fiatAmount = (this.$el.querySelector('.fiat-amount'));
+        if (paymentInfo.fiatAmount !== undefined && paymentInfo.fiatCurrency !== undefined) {
+            const fiatAmount = NumberFormatting.formatNumber(paymentInfo.fiatAmount, paymentInfo.fiatCurrency.digits);
+            $fiatAmount.textContent = `${fiatAmount} ${paymentInfo.fiatCurrency.code}`;
+        } else {
+            $fiatAmount.remove();
+        }
+
+        const recipientInfo = new AddressInfo({
+            userFriendlyAddress: paymentInfo.recipient,
+            label: paymentInfo.label || null,
+            imageUrl: paymentInfo.imageUrl || null,
+            accountLabel: null,
+        });
+        recipientInfo.renderTo(/** @type HTMLElement */ (this.$el.querySelector('.recipient-info')));
+
+        /** @type HTMLElement */
+        const $recipient = (this.$el.querySelector('.recipient'));
+        $recipient.addEventListener('click', event => {
+            event.preventDefault(); // to avoid CSP warnings about the javascript:void(0) href
+            this.fire(PaymentInfoLine.Events.RECIPIENT_CLICKED);
+        });
+    }
+
+    /**
+     * @private
+     * @param {HTMLElement} [$el]
+     * @returns {HTMLElement}
+     */
+    static _createElement($el) {
+        $el = $el || document.createElement('div');
+        $el.classList.add('payment-info-line');
+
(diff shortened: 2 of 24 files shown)


src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -29,6 +29,22 @@ class SignTransactionApi extends TopLevelApi {
             if (parsedRequest.shopLogoUrl && parsedRequest.shopLogoUrl.origin !== parsedRequest.shopOrigin) {
                 throw new Errors.InvalidRequestError('origin of shopLogoUrl must be same as referrer');
             }
+
+            parsedRequest.fiatAmount = this.parseNonNegativeFiniteNumber(request.fiatAmount);
+            parsedRequest.fiatCurrency = this.parseCurrencyInfo(request.fiatCurrency);
+            if ((parsedRequest.fiatAmount === undefined) !== (parsedRequest.fiatCurrency === undefined)) {
+                throw new Errors.InvalidRequestError('fiatAmount and fiatCurrency must be both defined or undefined.');
+            }
```
danimoh, line 39: I could add a check here that `fiatAmount` does not have more decimal digits than allowed by the the currency info.
Not sure though whether that's desirable?

client/src/PublicRequest.ts
```diff
@@ -93,13 +93,23 @@ export type ExportResult = {
...
+    code: string,
+    digits: number,
+};
+
 export type SignTransactionRequest = SimpleRequest & TransactionInfo & {
     keyPath: string,
     layout?: SignTransactionRequestLayout,
     recipientLabel?: string,
     senderLabel?: string,
+    // TODO these should only get added for 'checkout' layout
```
nibhar: could be done with something like this:
```typescript
export type SignTransactionRequest = SimpleRequest & TransactionInfo & {
    keyPath: string,
    recipientLabel?: string,
    senderLabel?: string,
} & ({
    layout: 'checkout',
    shopOrigin: string,
    shopLogoUrl?: string,
    time?: number,
    expires?: number,
    fiatCurrency?: CurrencyInfo,
    fiatAmount?: number,
} | { layout?: 'cashlink' | 'standard' });
```

However I think I would prefer to split that type in its components and then expose a union. That would result in a line or two more but would maybe be better readable. 
As this is not longer making use of `SignTransactionRequestLayout` it could potentially be removed, too.

src/components/Timer.js
```diff
@@ -0,0 +1,297 @@
...
+    get _timeLeft() {
+        return Math.max(0, Math.min(this._totalTime, this._endTime - Date.now()));
+    }
+
+    /**
+     * @private
+     * @returns {number}
+     */
+    get _progress() {
+        return 1 - this._timeLeft / this._totalTime;
```
nibhar, line 128: `this._totalTime` could be zero. Is the Timer behaving alright if `this._progress = Number.NEGATIVE_INFINITY`?
If additionally `this._timeLeft = 0` it results in  `this._progress = Number.NaN`, too. 
I realise that this is not really something that should happen, but it is not checked anywhere as far as I can tell if the start and end times do actually make sense.

src/lib/RequestParser.js
```diff
@@ -256,6 +256,22 @@ class RequestParser { // eslint-disable-line no-unused-vars
...
     }
 
+    /**
+     * Checks that a given value is a non-negative finite number.
+     * @param {any} value
+     * @param {boolean} [allowUndefined=true]
+     * @returns {number | undefined}
+     */
+    parseNonNegativeFiniteNumber(value, allowUndefined = true) {
+        if (value === undefined && allowUndefined) {
```
nibhar, line 266: I would swap the conditions
```suggestion
        if (allowUndefined && value === undefined) {
```
as the second condition is only relevant iff the first one is fulfilled.

src/request/sign-transaction/SignTransaction.js
```diff
@@ -102,6 +121,10 @@ class SignTransaction {
                 this._onConfirm(request, resolve, reject, password);
             },
         );
+
+        if (request.expires !== undefined) {
+            setTimeout(() => reject(new Errors.RequestExpired()), request.expires - Date.now());
```
nibhar, line 129: This could instead listen for `Timer.Events.EXPIRED`?

src/request/sign-transaction/SignTransactionApi.js
nibhar: I would say it should not throw in case it has more digits than it is supposed to have. However just rounding it to the least significant digit here is maybe not a bad idea.

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -39,7 +55,7 @@ class SignTransactionApi extends TopLevelApi {
     /**
      * Checks that the given layout is valid
      * @param {any} layout
```
nibhar, line 57: Not in this PR, but this should probably be 
```suggestion
      * @param {unknown} layout
```

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -52,6 +68,42 @@ class SignTransactionApi extends TopLevelApi {
         return layout;
     }
 
+    /**
+     * Parses that a currency info is valid.
+     * @param {any} currencyInfo
```
nibhar: Or maybe not. Did we not have those as `{unknown}`? If not there maybe should be a future PR to make them `{unknown}` across the board.

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -52,6 +68,42 @@ class SignTransactionApi extends TopLevelApi {
...
+     * @returns {KeyguardRequest.CurrencyInfo | undefined}
+     */
+    parseCurrencyInfo(currencyInfo) {
+        if (currencyInfo === undefined) {
+            return undefined;
+        }
+
+        // parse currency code
+        if (typeof currencyInfo.code !== 'string'
```
nibhar: So this will throw for i.e `typeof currencyInfo === 'number'`, but the Error will only show `Invalid currency code undefined`, which I would say is not ideal.

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -66,9 +118,9 @@ class SignTransactionApi extends TopLevelApi {
     }
 }
 
-/** @type {{[layout: string]: string}} */
+/** @enum {KeyguardRequest.SignTransactionRequestLayout} */
 SignTransactionApi.Layouts = {
-    STANDARD: 'standard',
-    CHECKOUT: 'checkout',
-    CASHLINK: 'cashlink',
+    STANDARD: /** @type {'standard'} */ ('standard'),
```
nibhar, line 108: Why is the annotation required?

nibhar (changes requested):

src/lib/RequestParser.js
sisou: The other way around the second one (`allowUndefined`) is also only relevant when the first one is fulfilled :wink:

sisou: IMO it is better as it is.

src/request/sign-transaction/SignTransactionApi.js
sisou: There was an effort by Max once to make them unknown, but the Typescript version had too many problem and thus was never merged.

src/lib/RequestParser.js
nibhar: I like it better the other way around, beginning with the boolean that toggles this functionality.

src/request/sign-transaction/SignTransactionApi.js
nibhar: I tested it just now and most of them actually work with `{unknown}`.

sisou: Yes, they work. I'm just saying this change was part of a bigger PR which did not get merged because of other problems.

src/request/sign-transaction/SignTransactionApi.js
danimoh: The fiat decimal count is now not part of the API anymore. Rendering and rounding of fiat currency strings is now handled by the browser and this comment does not apply anymore.

client/src/PublicRequest.ts
danimoh: I experimented with this too.
However, I was using an enum for `layout` which made it appear as this was not working while actually the problem was that ts couldn't match that ts enum to SignTransactionApi.Layouts on the javascipt side.
I eventually figured this out and managed to map the js enum to the ts enum via
```js
/**
 * @enum {KeyguardRequest.SignTransactionRequestLayout}
 * @readonly
 */
SignTransactionApi.Layouts = {
    STANDARD: /** @type {KeyguardRequest.SignTransactionRequestLayout.STANDARD} */ ('standard'),
    CHECKOUT: /** @type {KeyguardRequest.SignTransactionRequestLayout.CHECKOUT} */ ('checkout'),
    CASHLINK: /** @type {KeyguardRequest.SignTransactionRequestLayout.CASHLINK} */ ('cashlink'),
};
```

However, to not change the existing type for `KeyguardRequest.SignTransactionRequestLayout` I later discarded the change to an enum.
But I kept the separation of the `SignTransactionRequest` into two sub types.
For this also extra handling had to be added to `Parsed` in `Keyguard.d.ts` as we discussed together.

src/components/Timer.js
danimoh: `SignTransactionApi` rejects requests with `time >= expires`, thus `_totalTime` will never be 0.
Of course you could argue that `Timer` is a more general component and shouldn't rely on how it is actually used in `SignTransaction`.
However, the timer behaves ok even when `_totalTime === 0`. Note that when `_totalTime === 0` also `_timeLeft === 0` holds true, therefore `_progress` will be `NaN` and the browser will render a full circle.
Also other than in the implementation in `vue-components`, the timer is not reactive on the `startTime` and `endTime`. These values are just set initially on creation. Thus, a Timer instance with `_totalTime === 0` will never serve a useful purpose.

src/lib/RequestParser.js
danimoh: I think this doesn't really matter, but I also prefer how it is currently.

src/request/sign-transaction/SignTransaction.js
danimoh: It could but I would need to forward that through the `PaymentInfoLine` so that I think that the current approach is nicer.

src/request/sign-transaction/SignTransactionApi.js
danimoh: Why would that be better than `any`?
Isn't `any` appropriate here?
The method can handle `any`thing and returns the parsed layout if the input could actually be parsed as a layout.

src/request/sign-transaction/SignTransactionApi.js
danimoh: It is an enum over type `KeyguardRequest.SignTransactionRequestLayout` which includes only the constants 'standard', 'checkout' and 'cashlink'. Therefore, the strings need to be casted to these constant values.

src/request/sign-transaction/SignTransactionApi.js
danimoh: This does now not apply anymore as `fiatCurrency` changed from being a currency info object to a simple ISO string.

src/request/sign-transaction/SignTransactionApi.js
nibhar: `unknown` is less permissive than `any`. 
As an example
```ts
(x: any) => { console.log(x.test); }
```
is valid for the typechecker, while
```ts
(x: unknown) => { console.log(x.test); }
```
is not.

Therefore I think it would be better for the parser to assume `unknown` than `any`, generally speaking.

src/components/Timer.js
nibhar: I agree, and just wanted to point out, that the timer does not test for its inputs to be valid.

client/src/PublicRequest.ts
```diff
@@ -93,15 +93,25 @@ export type ExportResult = {
     wordsExported: boolean,
 };
 
-export type SignTransactionRequest = SimpleRequest & TransactionInfo & {
+export type SignTransactionRequestSimple = SimpleRequest & TransactionInfo & {
```
nibhar, line 96: I don't know if I like `Simple` here as it still is somewhat of a complex request in comparison.

Aside from that, nice work on the types.

src/components/PaymentInfoLine.js
```diff
@@ -0,0 +1,78 @@
...
+/* global AddressInfo */
+/* global Timer */
+/* global TemplateTags */
+/* global NumberFormatting */
+
+class PaymentInfoLine { // eslint-disable-line no-unused-vars
+    /**
+     * @param {object}           paymentInfo
+     * @param {string}           paymentInfo.recipient
+     * @param {string|undefined} [paymentInfo.label]
```
nibhar: I think the `| undefined` are not required, as the brackets in `[paymentInfo.label]` already take care of that. Same for the other optional fields.

src/components/Timer.js
```diff
@@ -0,0 +1,297 @@
...
+        /** @private
+         *  @type {boolean} */
+        this._detailsShown = false;
+        // While the radius r of the circle and the values stroke-dasharray, stroke-dashoffset and stroke-width that
+        // depend on the radius can be transitioned via css, the behavior on value update during an ongoing transition
+        // is not consistent (e.g. time update while animating on user hover or quick hover and unhover). Therefore we
+        // animate via JS.
+        /** @private
+         *  @type {Tweenable} */
+        this._radius = new Tweenable(8);
```
nibhar: That `8` could be a variable, or a constant. The `12` later on (`_showDetails()`) could be `1.5 * this.timerRadius`, or `1.5 * Timer.RADIUS` respectively.

src/components/PaymentInfoLine.js
```diff
@@ -0,0 +1,78 @@
...
+     * @param {URL|undefined}    [paymentInfo.imageUrl]
+     * @param {number}           paymentInfo.lunaAmount
+     * @param {number|undefined} [paymentInfo.fiatAmount]
+     * @param {string|undefined} [paymentInfo.fiatCurrency]
+     * @param {number|undefined} [paymentInfo.time]
+     * @param {number|undefined} [paymentInfo.expires]
```
sisou: That's cool that you can do object prop types like this! :open_mouth: 
However, wouldn't it be better to define the `PaymentInfo` interface as a type and simply use the type here?

src/components/Timer.js
```diff
@@ -0,0 +1,297 @@
...
+        return $el;
+    }
+
+    /**
+     * @private
+     * @param {number} millis
+     * @param {boolean} [includeUnit=true]
+     * @returns {string|number}
+     */
+    static _toSimplifiedTime(millis, includeUnit = true) {
```
nibhar: This function is private and only ever used on `this.timeLeft`.
Is there a reason to have it static instead of
```js
_toSimplifiedTime(includeUnit = true) {...}
```
Additionally I think `_toSimplifiedTime` is maybe not descriptive for what this does. Maybe something along the lines of `_toSmallestPossibleTimeUnit()`?

It might also make sense to have the return type as `string` only, as it is either cast to `string` or `toString`ed anyways.

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -23,14 +24,29 @@ class SignTransactionApi extends TopLevelApi {
         parsedRequest.recipientLabel = this.parseLabel(request.recipientLabel);
         parsedRequest.transaction = this.parseTransaction(request);
         parsedRequest.layout = this.parseLayout(request.layout);
-        if (parsedRequest.layout === SignTransactionApi.Layouts.CHECKOUT) {
+        if (parsedRequest.layout === SignTransactionApi.Layouts.CHECKOUT
+            && request.layout === SignTransactionApi.Layouts.CHECKOUT) {
```
nibhar, line 28: Why do we need to check the layout for both the parsed and unparsed request?

nibhar (approved):
Approved with some remarks.

src/components/PaymentInfoLine.js
```diff
@@ -0,0 +1,78 @@
...
+/* global TemplateTags */
+/* global NumberFormatting */
+
+class PaymentInfoLine { // eslint-disable-line no-unused-vars
+    /**
+     * @param {object}           paymentInfo
+     * @param {string}           paymentInfo.recipient
+     * @param {string|undefined} [paymentInfo.label]
+     * @param {URL|undefined}    [paymentInfo.imageUrl]
+     * @param {number}           paymentInfo.lunaAmount
```
sisou: Does it really have to be called `lunaAmount`? I feel the default for `amount` should just be luna by definition.

src/lib/NumberFormatting.js
```diff
@@ -0,0 +1,45 @@
+class NumberFormatting { // eslint-disable-line no-unused-vars
+    /**
+     * @param {number} value
+     * @param {number} [maxDecimals]
+     * @param {number} [minDecimals]
+     * @returns {string}
+     */
+    static formatNumber(value, maxDecimals = 5, minDecimals = Math.min(0, maxDecimals)) {
```
sisou: Why is the `Math.min` required for the `minDecimals`? Why is it not enough to just put `0` as the default? Or is it supposed to be `Math.max(0, minDecimals)` to not have negative decimals?

src/lib/NumberFormatting.js
```diff
@@ -0,0 +1,45 @@
...
+        return result.replace(regexp, '$1\u202F');
+    }
+
+    /**
+     * @param {number} value
+     * @param {string} currency
+     * @param {string} [locale = navigator.language]
+     * @returns {string}
+     */
+    static formatCurrency(value, currency, locale = navigator.language) {
```
sisou, line 29:
```suggestion
    static formatFiat(value, currency, locale = navigator.language) {
```

src/components/PaymentInfoLine.js
```diff
@@ -0,0 +1,78 @@
...
+        this.$el = PaymentInfoLine._createElement($el);
+
+        /** @type HTMLElement */
+        const $nimAmount = (this.$el.querySelector('.nim-amount'));
+        const nimAmount = NumberFormatting.formatNumber(Nimiq.Policy.lunasToCoins(paymentInfo.lunaAmount));
+        $nimAmount.textContent = `${nimAmount} NIM`;
+
+        /** @type HTMLElement */
+        const $fiatAmount = (this.$el.querySelector('.fiat-amount'));
+        if (paymentInfo.fiatAmount !== undefined && paymentInfo.fiatCurrency !== undefined) {
```
sisou: Can `fiatAmount` also be `null` and would that be handled differently? Otherwise I think `fiatAmount` can not be `0` (should be validated in the RequestParser), thus we can use the simple truthy check here instead:

```suggestion
        if (paymentInfo.fiatAmount && paymentInfo.fiatCurrency) {
```

src/components/PaymentInfoLine.js
```diff
@@ -0,0 +1,78 @@
...
+            userFriendlyAddress: paymentInfo.recipient,
+            label: paymentInfo.label || null,
+            imageUrl: paymentInfo.imageUrl || null,
+            accountLabel: null,
+        });
+        recipientInfo.renderTo(/** @type HTMLElement */ (this.$el.querySelector('.recipient')));
+
+        /** @type HTMLElement */
+        const $timer = (this.$el.querySelector('.timer'));
+        if (paymentInfo.time !== undefined && paymentInfo.expires !== undefined) {
```
sisou: I think both `time` and `expires` can also not be the number `0`, so it's safe to use truthy checks here:

```suggestion
        if (paymentInfo.time && paymentInfo.expires) {
```

src/components/Timer.css
```diff
@@ -0,0 +1,105 @@
...
+    height: 8rem;
+    padding: 2rem 1.25rem .875rem 1.5rem;
+    opacity: 0;
+    font-size: 1.75rem;
+    line-height: 1.5;
+    font-weight: 600;
+    color: white;
+    z-index: 1;
+    pointer-events: none;
+    background-image: url('data:image/svg+xml,<svg viewBox="0 0 136 63.9" xmlns="http://www.w3.org/2000/svg"><path d="M136 59.9a4 4 0 0 1-4 4H4a4 4 0 0 1-4-4v-47a4 4 0 0 1 4-4h99a4 4 0 0 0 3.2-1.7l4.6-6.6c.6-.8 1.8-.8 2.4 0l4.6 6.6a4 4 0 0 0 3.3 1.7H132a4 4 0 0 1 4 4z" fill="url(%23a)"/><defs><radialGradient id="a" cx="0" cy="0" r="1" gradientTransform="matrix(-190 0 0 -64.999 136 70.9)" gradientUnits="userSpaceOnUse"><stop stop-color="%23260133" offset="0"/><stop stop-color="%231F2348" offset="1"/></radialGradient></defs></svg>');
```
sisou, line 98: Does the timer tooltip background need to be an SVG instead of `background-image: var(--nq-blue-bg)`?

src/components/Timer.js
sisou: While I agree on making it non-static and the return type (string), I think the name is alright. `_toSmallestPossibleTimeUnit` would suggest to me that it's always in seconds (or milliseconds), which is not the case.

Maybe `_toRoundedTime`?

src/components/Timer.js
```diff
@@ -0,0 +1,297 @@
...
+    get _totalTime() {
+        return Math.max(0, this._endTime - this._startTime);
+    }
+
+    /**
+     * @private
+     * @returns {number}
+     */
+    get _timeLeft() {
+        return Math.max(0, Math.min(this._totalTime, this._endTime - Date.now()));
```
sisou, line 120: I think in the Hub we are working with offsets to the browser time to prevent mis-set browser clocks from breaking the timer. So when the page loads, the offset between the `startTime` and `Date.now()` is calculated and used in all relative calculations. Or is this offset already applied outside the timer?

src/lib/NumberFormatting.js
```diff
@@ -0,0 +1,45 @@
...
+            currency,
+            currencyDisplay: 'symbol',
+            useGrouping: false,
+        });
+        const integerMatch = formatted.match(/\d+/); // first match is the integer part
+        if (!integerMatch) return formatted; // should never happen
+        const integerPart = integerMatch[0];
+        return formatted
+            .replace(/(\d)\D(\d)/, '$1.$2') // enforce a dot as decimal separator for
+            // consistency. Using capturing groups instead of lookahead/lookbehind to avoid browser support limitations.
```
sisou: I would be happier if this two-line comment is above the line that it's about.

src/lib/Tweenable.js
```diff
@@ -0,0 +1,68 @@
+class Tweenable {
+    /**
+     * @param {number} [targetValue=0]
+     * @param {number} [startValue=targetValue]
+     * @param {number} [tweenTime=0]
+     * @param {number} [startTime=Date.now()]
+     * @param {(function(number): number)} [easing=Tweenable.Easing.EASE_IN_OUT_CUBIC]
```
sisou: _All_ arguments are optional?

Also, what is `tweenTime`? Can that get an explaining doc comment? Is that the _duration_?

src/request/sign-transaction/SignTransaction.js
```diff
@@ -46,19 +49,37 @@ class SignTransaction {
...
-            label: request.shopOrigin
-                ? request.shopOrigin.split('://')[1]
-                : request.recipientLabel || null,
-            imageUrl: request.shopLogoUrl || null,
+            userFriendlyAddress: recipientAddress,
+            label: recipientLabel,
+            imageUrl: 'shopLogoUrl' in request && !!request.shopLogoUrl
+                ? request.shopLogoUrl
+                : null,
             accountLabel: null,
```
sisou: I think for consistency the image assignment conditional should also be pulled above the object.

src/request/sign-transaction/SignTransaction.js
```diff
@@ -102,6 +123,10 @@ class SignTransaction {
                 this._onConfirm(request, resolve, reject, password);
             },
         );
+
+        if ('expires' in request && request.expires !== undefined) {
```
sisou: Again, I think expires can not be `0`, so a truthy check is enough.

```suggestion
        if (request.expires) {
```

src/request/sign-transaction/SignTransactionApi.js
sisou: I have the same question.

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -23,14 +24,29 @@ class SignTransactionApi extends TopLevelApi {
...
             parsedRequest.shopOrigin = this.parseShopOrigin(request.shopOrigin);
             parsedRequest.shopLogoUrl = this.parseShopLogoUrl(request.shopLogoUrl);
             if (parsedRequest.shopLogoUrl && parsedRequest.shopLogoUrl.origin !== parsedRequest.shopOrigin) {
                 throw new Errors.InvalidRequestError('origin of shopLogoUrl must be same as referrer');
```
sisou: Since the referrer can't be checked in the Keyguard, this error message is technically wrong. And I think this is also checked in the Hub already? Why not:

```suggestion
                 throw new Errors.InvalidRequestError('origin of shopLogoUrl must be same as shopOrigin');
```

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -23,14 +24,29 @@ class SignTransactionApi extends TopLevelApi {
...
+            parsedRequest.fiatAmount = this.parseNonNegativeFiniteNumber(request.fiatAmount);
+            parsedRequest.fiatCurrency = this.parseFiatCurrency(request.fiatCurrency);
+            if ((parsedRequest.fiatAmount === undefined) !== (parsedRequest.fiatCurrency === undefined)) {
+                throw new Errors.InvalidRequestError('fiatAmount and fiatCurrency must be both defined or undefined.');
+            }
+
+            parsedRequest.time = this.parseNonNegativeFiniteNumber(request.time);
+            parsedRequest.expires = this.parseNonNegativeFiniteNumber(request.expires);
+            if (parsedRequest.expires !== undefined) {
+                if (parsedRequest.time === undefined) {
```
sisou, line 44: `time` and `expires` should never be `0`, right? So all these tests for `undefined` specifically are unnecessary IMO.

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -23,14 +24,29 @@ class SignTransactionApi extends TopLevelApi {
...
+                    throw new Errors.InvalidRequestError('If `expires` is defined `time` must be defined too.');
```
sisou:
```suggestion
                    throw new Errors.InvalidRequestError('If `expires` is given, `time` must be given, too.');
```

tsconfig.json
```diff
@@ -11,7 +11,7 @@
     // "sourceMap": true,                     /* Generates corresponding '.map' file. */
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
     "outDir": "./.ts-out",                    /* Redirect output structure to the directory. */
-    "incremental": true,
+    "incremental": false,
```
sisou, line 14: What problem did you run into that you had to disable incremental compilation?

types/Keyguard.d.ts
```diff
@@ -54,15 +54,29 @@ type ConstructTransaction<T extends KeyguardRequest.TransactionInfo> = Transform
...
         | Is<T, KeyguardRequest.RemoveKeyRequest>
         | Is<T, KeyguardRequest.ExportRequest> ? KeyId2KeyInfo<T> :
-    T extends Is<T, KeyguardRequest.ImportRequest> ? Transform<KeyguardRequest.ImportRequest,
-        'isKeyLost', { isKeyLost: boolean }> :
+    T extends Is<T, KeyguardRequest.ImportRequest> ?
+        Transform<
+            KeyguardRequest.ImportRequest,
+            'isKeyLost',
+            { isKeyLost: boolean }
+        > :
```
sisou, line 81: Not really more readable IMO.

sisou (changes requested):

sisou:
I would appreciate it if updates to this PR are done as commits on top of the existing commits, and _not_ as rebases ;). Makes reviewing the updates much easier.

My requested changes are mostly small, simple stuff, that I nontheless think will make readability and understanding easier for devs unfamiliar with the code.

src/request/sign-transaction/SignTransactionApi.js
danimoh: I see.
But I'd say, that should be part of a separate PR that makes this change across the whole project.

nibhar: I agree.

client/src/PublicRequest.ts
danimoh: Suggestions?

src/components/PaymentInfoLine.js
danimoh: Unfortunately typescript doesn't understand that sub properties of an object in this notation that are put into square brackets are optional, therefore the need to explicitly specify `undefined`.

src/components/Timer.js
danimoh: I think from the suggested names, `_toSimplifiedTime` is still my favorite.

Why do you think that non-static here would be nicer?

But I agree on the return type.

src/request/sign-transaction/SignTransactionApi.js
danimoh: It's required for ts to be able to infer that the checkout specific properties exist on `request` and `parsedRequest`.

src/components/PaymentInfoLine.js
danimoh: I guess I put it that way for easier distinction from the fiat amount.

src/lib/NumberFormatting.js
danimoh: It's indeed not.
This was before `Math.min(2, maxDecimals))` but since I changed it to `0`, the `Math.min` is not required anymore.

src/lib/NumberFormatting.js
danimoh: Technically, this could also be used to format other currencies like `BTC`, `NIM`, etc.

src/components/Timer.js
nibhar: As I see it, it is not really static the way it is used. It is always used on `this.timeLeft`. It is also private so it cannot get used anywhere else either.  I would prefer it to be `get _simplifiedTimeLeft()` but the includeUnit argument prevents that.

Therefore I would have it like this
```js
_toSimplifiedTime(includeUnit = true) {
        // find appropriate unit, starting with second
        let resultTime = this._timeLeft / 1000;
...
```

src/components/PaymentInfoLine.js
danimoh: `fiatAmount` can be `number|undefined`. The check for `undefined` was just to make this explicit.
But I can change it.

src/components/PaymentInfoLine.js
danimoh: Changed this too.

src/components/Timer.css
danimoh: Yes, the background image includes the small tooltip dent.

src/components/Timer.js
danimoh: The keyguard expects the times already to be converted to the local time.

src/lib/NumberFormatting.js
danimoh: Alright, then let's make you happy.

src/lib/Tweenable.js
danimoh: Yes, it's the duration. Renamed it to `tweenDuration`.

src/request/sign-transaction/SignTransaction.js
danimoh: The `'expires' in request` is required for ts to know that this is a request on which `expires` exists.

src/request/sign-transaction/SignTransactionApi.js
danimoh: In this case typescript can't infer that in line 46 `parsedRequest.time` can be `undefined` if not checking for `undefined` here.
In other cases it's essentially a matter of taste.

tsconfig.json
danimoh: The build was using cached results that did not resemble what was in the code.
The non-incremental build is still very quick and for an app as important as the Keyguard it's worth to wait the extra time to get a proper build or rather typecheck.
The time difference is not really noticeable anyways.

types/Keyguard.d.ts
danimoh: In my opinion it is.

danimoh:
Well, reviewing force-pushed changes is not really much harder. You just have to diff from the commit hash that you last reviewed at (which github tells you) to the latest commit. All this can be done in github too.

src/components/Timer.js
sisou: I'm with Sebastian. If not used statically and only used from an instance, it should not be static and just be on the instance.

src/request/sign-transaction/SignTransactionApi.js
sisou: For me it's not just about taste, but also about defensive programming. Just because we tell Typescript that it can be `string | undefined` (for example), does not mean that outside calls cannot pass `null` or other types in. Typescript does not do type checks during runtime, and when it's not checked by whitelist anywhere, types are passed through as they are. So to be sure, I would always do a falsy or whitelist check for formerly untested arguments.

src/lib/NumberFormatting.js
sisou: Last I checked, `value.toLocaleString()` did not support crypto currencies :stuck_out_tongue: So it cannot, in fact, be used for crypto _currencies_ (https://www.currency-iso.org/dam/downloads/lists/list_one.xml).

src/components/PaymentInfoLine.js
sisou: I still think a `@type` definition makes sense for this instead.

src/lib/NumberFormatting.js
sisou: Last I checked, `value.toLocaleString()` did not support crypto currencies :stuck_out_tongue: So it cannot, in fact, be used for crypto _currencies_ (https://www.currency-iso.org/dam/downloads/lists/list_one.xml).

sisou (approved):

src/lib/NumberFormatting.js
danimoh: Well, here is the specification for it: https://www.ecma-international.org/ecma-402/2.0/#sec-iswellformedcurrencycode

src/components/PaymentInfoLine.js
danimoh: Changed it to a `@typedef`

src/lib/NumberFormatting.js
sisou: OK, I have tested it and it actually accepts any three letter code and if it doesn't know it it just displays the uppercase three letters in front of the number formatted to (at least) two decimals.

OK fine.

PR #383, opened 2019-11-07
add ExportResult to RedirectResult

nibhar:


client/src/PublicRequest.ts
```diff
@@ -161,7 +161,8 @@ export type IFrameResult = ListResult
     | DerivedAddress[]
     | SimpleResult;
 
-export type RedirectResult = KeyResult
+export type RedirectResult = ExportResult
+    | KeyResult
```
sisou, line 169: Would it be possible to have all combined results in separate lines, instead of the first one not? Just for code cleanliness :stuck_out_tongue:

sisou (approved):

client/src/PublicRequest.ts
nibhar: Like so?

danimoh (approved):

client/src/PublicRequest.ts
sisou: Yes, lgtm.

PR #384, opened 2019-11-15
Delay hash change during file export on iOS 13.x+

### LINKED ISSUE #382: LoginFile download partially broken on iOS 13+
iOS Safari 13 finally introduced support for the [download attribute](https://caniuse.com/#feat=download) such that on Safari 13+ our long tap download work-around is not used anymore.
Unfortunately the implementation of the download attribute seems to be broken or not compatible with our usage of it. Downloading the login file only succeeds sometimes and no pattern has been recognized so far when it works and when not.

So far two possible causes for that behavior have been identified but not investigated yet:
- The link with the download attribute gets set to `display: none` before the download concludes.
- iOS Safari might have problems downloading data urls. We could use an [object url](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) for a [blob](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) instead.

nibhar:
While Safari on iOS 13 does support the download attribute, the resulting download prompt does not like the hash change  in the background happening in our setup. 
This PR changes that behaviour by delaying the hash change until after the prompt was confirmed.

This PR also increases core-web to Version 1.5.3.

Resolves #382

### DIFF
--- a/src/components/DownloadLoginFile.js
+++ b/src/components/DownloadLoginFile.js
@@ -46,10 +46,16 @@ class DownloadLoginFile extends Nimiq.Observable {
         this.$loginfile.addEventListener('mousedown', e => this._onMouseDown(e));
         this.$loginfile.addEventListener('touchstart', () => this._onTouchStart());
         this.$downloadButton.addEventListener('click', () => this._onDownloadStart());
-        $continueButton.addEventListener('click', this._onDownloadEnd.bind(this));
+        $continueButton.addEventListener('click', () => {
+            this._onDownloadEnd();
+            // Remove previously added classes after a short delay to restore the initial state.
+            window.setTimeout(() => {
+                this.$el.classList.remove('maybe-downloaded');
+                this.fire(DownloadLoginFile.Events.RESET);
+            }, 300);
+        });
         $backToDownloadButton.addEventListener('click', () => {
-            this.$el.classList.remove('maybe-downloaded');
-            this.fire(DownloadLoginFile.Events.RESET);
+            this.$downloadButton.click();
         });
     }
 
@@ -101,6 +107,7 @@ class DownloadLoginFile extends Nimiq.Observable {
         }
         // Remove previously added classes to restore the initial state.
         this.$el.classList.remove('maybe-downloaded');
+        this.fire(DownloadLoginFile.Events.RESET);
 
         const color = IqonHash.getBackgroundColorIndex(firstAddress.toUserFriendlyAddress());
         this.file = new LoginFile(Nimiq.BufferUtils.toBase64(encryptedEntropy), color);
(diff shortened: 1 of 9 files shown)


src/components/DownloadLoginFile.js
```diff
@@ -166,8 +177,16 @@ class DownloadLoginFile extends Nimiq.Observable {
 
             // If window gets blurred, show 'Continue' button in interface and do not automatically
             // consider the download successful.
-            if (!document.hasFocus()) {
+            // As mobile Safari on iOS 13.x+ does support the download attribute, it no longer uses the long tap
+            // fallback. However, if the page changes its hash in the background while the promt asking to download
```
sisou: Typo `promt => prompt`

sisou (approved):
Perfect! I'll add the Nimiq lib hashes in a few minutes ;).

sisou (approved):

src/components/DownloadLoginFile.js
```diff
@@ -42,7 +46,17 @@ class DownloadLoginFile extends Nimiq.Observable {
...
+            this._onDownloadEnd();
+            // Remove previously added classes after a short delay to restore the initial state.
+            window.setTimeout(() => {
+                this.$el.classList.remove('maybe-downloaded');
+                this.fire(DownloadLoginFile.Events.RESET);
+            }, 300);
+        });
+        $backToDownloadButton.addEventListener('click', () => {
+            this.$downloadButton.click();
+        });
```
danimoh: No need to open a block here.

src/components/DownloadLoginFile.js
```diff
@@ -90,6 +105,9 @@ class DownloadLoginFile extends Nimiq.Observable {
         if (encryptedEntropy.byteLength !== KeyStore.ENCRYPTED_SECRET_SIZE) {
             throw new Errors.KeyguardError('Can only export encrypted Entropies');
         }
+        // Remove previously added classes to restore the initial state.
+        this.$el.classList.remove('maybe-downloaded');
+        this.fire(DownloadLoginFile.Events.RESET);
```
danimoh: You're using this twice. Might be worth moving this to a `reset` method.

src/components/DownloadLoginFile.js
```diff
@@ -166,8 +184,16 @@ class DownloadLoginFile extends Nimiq.Observable {
...
             // consider the download successful.
-            if (!document.hasFocus()) {
+            // As mobile Safari on iOS 13.x+ does support the download attribute, it no longer uses the long tap
+            // fallback. However, if the page changes its hash in the background while the prompt asking to download
+            // the file was not confirmed yet, the download will simply do nothing. To prevent that behaviour the
+            // hash change is delayed by the .maybe-downloaded confirmation mechanism.
+            if (!document.hasFocus()
+                || (BrowserDetection.isIOS()
+                    && BrowserDetection.isSafari()
+                    && BrowserDetection.iOSVersion()[0] > 12)) {
```
danimoh: I think we should test whether we need to include iOS Chrome too.
Or didn't we even say we want to have the `Continue` button always on mobile?

danimoh (changes requested):

src/components/DownloadLoginFile.js
nibhar: Actually I think we should probably just get rid of the `BrowserDetection.isSafari()` condition.
For non iOS devices I think the continue button is simply not necessary, but on iOS 13 we can just declare this as default behaviour, no matter what browser.

danimoh (approved):
Resolves https://github.com/nimiq/keyguard/issues/382

PR #387, opened 2019-11-27
More specific SignTransactionRequest typing

danimoh:


client/src/PublicRequest.ts
```diff
@@ -93,16 +93,19 @@ export type ExportResult = {
     wordsExported: boolean,
 };
 
-export type SignTransactionRequestSimple = SimpleRequest & TransactionInfo & {
-    layout?: Exclude<SignTransactionRequestLayout, 'checkout'>,
+type SignTransactionRequestCommon = SimpleRequest & TransactionInfo & {
```
nibhar, line 96: 👍

nibhar (approved):

client/src/PublicRequest.ts
sisou: What do you think about `SignTransactionRequestBase`?

client/src/PublicRequest.ts
```diff
@@ -111,7 +113,14 @@ export type SignTransactionRequestCheckout = Omit<SignTransactionRequestSimple,
...
 
-export type SignTransactionRequest = SignTransactionRequestSimple | SignTransactionRequestCheckout;
+export type SignTransactionRequestCashlink = SignTransactionRequestCommon & {
+    layout: 'cashlink',
+    cashlinkMessage?: string,
+};
+
+export type SignTransactionRequest = SignTransactionRequestStandard
+    | SignTransactionRequestCheckout
+    | SignTransactionRequestCashlink;
```
sisou: I believe in the Hub we have the indentation all the way to the equals sign.

sisou (approved):
I hate that it requires the conditions to be doubled...

client/src/PublicRequest.ts
nibhar: `Common` is good in my opinion.

PR #389, opened 2019-11-30
Reset password option

nibhar:
This Pr adds a reset password option to the change password request following a similar strategy to the go to create button in Login.

Also included is an adjustment to the import request allowing the enforcement of the Recovery Words portion of it as well as an expectedKeyId which needs to be matched by said recovery words in order to continue.

### DIFF
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -61,6 +61,8 @@ export type ImportRequest = BasicRequest & {
     requestedKeyPaths: string[],
     isKeyLost?: boolean,
     enableBackArrow?: boolean,
+    wordsOnly?: boolean,
+    expectedKeyId?: string,
 };
 
 export type ReleaseKeyRequest = {
@@ -230,6 +232,8 @@ export type KeyguardError = {
     Messages: {
         // specifically used to trigger a redirect to create after returning to caller
         GOTO_CREATE: 'GOTO_CREATE',
+        // Specifically used to trigger a redirect to a special import after returning to caller
+        GOTO_RESET_PASSWORD: 'GOTO_RESET_PASSWORD',
         // used to signal a user initiated cancelation of the request
         CANCELED: 'CANCELED',
         // used to signal that the request expired
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -62,7 +62,10 @@ export type ImportRequest = BasicRequest & {
     isKeyLost?: boolean,
     enableBackArrow?: boolean,
     wordsOnly?: boolean,
-    expectedKeyId?: string,
+};
+
+export type ResetPasswordRequest = ImportRequest & {
+    expectedKeyId: string,
 };
 
 export type ReleaseKeyRequest = {
@@ -203,7 +206,7 @@ export type ResultType<T extends RedirectRequest> =
         | Is<T, SignTransactionRequestCashlink>
         ? SignatureResult :
     T extends Is<T, DeriveAddressRequest> ? DerivedAddress[] :
-    T extends Is<T, CreateRequest> | Is<T, ImportRequest> ? KeyResult :
+    T extends Is<T, CreateRequest> | Is<T, ImportRequest> | Is<T, ResetPasswordRequest> ? KeyResult :
     T extends Is<T, ExportRequest> ? ExportResult :
     T extends Is<T, RemoveKeyRequest> | Is<T, SimpleRequest> ? SimpleResult :
     never;
--- a/src/components/PasswordBox.css
+++ b/src/components/PasswordBox.css
@@ -109,3 +109,12 @@
     text-align: center;
     width: 100%;
 }
+
+.password-box .skip {
+    opacity: 1;
+    margin-top: -3.5rem;
+}
+
+.password-box.input-eligible .skip {
+    opacity: 0;
+}
--- a/src/components/PasswordBox.js
+++ b/src/components/PasswordBox.js
@@ -5,11 +5,21 @@
 /* global TemplateTags */
 /* global Key */
 
+/**
+ *  @typedef {{
+ *      bgColor: string,
+ *      hideInput: boolean,
+ *      buttonI18nTag: string,
+ *      minLength: number,
+ *      showResetPassword: boolean,
+ *  }} PasswordBoxOptions
+ */
+
 class PasswordBox extends Nimiq.Observable {
     // eslint-disable-next-line valid-jsdoc
     /**
      * @param {HTMLFormElement} [$el]
-     * @param {{bgColor?: string, hideInput?: boolean, buttonI18nTag?: string, minLength?: number}} [options]
+     * @param {Partial<PasswordBoxOptions>} [options]
      */
     constructor($el, options = {}) {
         const defaults = {
@@ -17,11 +27,12 @@ class PasswordBox extends Nimiq.Observable {
             hideInput: false,
             buttonI18nTag: 'passwordbox-confirm-tx',
             minLength: PasswordInput.DEFAULT_MIN_LENGTH,
+            showResetPassword: false,
         };
 
         super();
 
-        /** @type {{bgColor: string, hideInput: boolean, buttonI18nTag: string, minLength: number}} */
+        /** @type {PasswordBoxOptions} */
         this.options = Object.assign(defaults, options);
 
         this.$el = PasswordBox._createElement($el, this.options);
@@ -36,11 +47,19 @@ class PasswordBox extends Nimiq.Observable {
         this._isInputValid = false;
 
         this.$el.addEventListener('submit', event => this._onSubmit(event));
+
+        if (options.showResetPassword) {
+            /** @type {HTMLAnchorElement} */
+            (this.$el.querySelector('.skip')).addEventListener('click', /** @param {Event} event */ event => {
+                event.preventDefault();
+                this.fire(PasswordBox.Events.RESET_PASSWORD);
+            });
+        }
     }
 
     /**
      * @param {HTMLFormElement} [$el]
-     * @param {{bgColor: string, hideInput: boolean, buttonI18nTag: string, minLength: number}} options
+     * @param {PasswordBoxOptions} options
      * @returns {HTMLFormElement}
      */
     static _createElement($el, options) {
@@ -64,6 +83,16 @@ class PasswordBox extends Nimiq.Observable {
             'passwordbox-sign-msg': '<button class="submit" data-i18n="passwordbox-sign-msg">Sign message</button>',
         };
 
+        const resetPasswordHtml = options.showResetPassword
+            ? TemplateTags.noVars`
+                <a href="#" class="skip nq-link">
+                    <span data-i18n="passwordbox-reset-password">Reset with Recovery Words</span>
+                    <svg class="nq-icon">
+                        <use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-caret-right-small"/>
+                    </svg>
+                <a>`
+            : '';
+
         /** @type {{[i18nTag: string]: string}} */
         const promptVersions = {
             'passwordbox-enter-password': '<div class="prompt nq-text-s" data-i18n="passwordbox-enter-password">Enter your password</div>',
@@ -74,7 +103,7 @@ class PasswordBox extends Nimiq.Observable {
         if (!buttonVersions[options.buttonI18nTag]) throw new Error('PasswordBox button i18n tag not defined');
 
         /* eslint-disable max-len */
-        $el.innerHTML = TemplateTags.hasVars(2)`
+        $el.innerHTML = TemplateTags.hasVars(3)`
             ${promptVersions[options.minLength === Key.PIN_LENGTH ? 'passwordbox-enter-pin' : 'passwordbox-enter-password']}
             <div password-input></div>
             ${buttonVersions[options.buttonI18nTag]}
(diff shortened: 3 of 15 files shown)


src/components/PasswordBox.js
```diff
@@ -64,6 +90,16 @@ class PasswordBox extends Nimiq.Observable {
             'passwordbox-sign-msg': '<button class="submit" data-i18n="passwordbox-sign-msg">Sign message</button>',
         };
 
+        const resetPassword = options.showResetPassword
```
sisou:
```suggestion
        const resetPasswordHtml = options.showResetPassword
```

src/lib/Errors.js
```diff
@@ -81,6 +81,12 @@ Errors.RequestExpired = class extends Errors.KeyguardError {
     }
 };
 
+Errors.ResetPassword = class extends Errors.KeyguardError {
```
sisou: To keep in the spirit of the other GOTO error.

I would also define this error _after_ the other GOTO one (that's where I first looked for it and R comes _after_ C in the alphabet).

```suggestion
Errors.GoToResetPassword = class extends Errors.KeyguardError {
```

src/lib/Errors.js
```diff
@@ -81,6 +81,12 @@ Errors.RequestExpired = class extends Errors.KeyguardError {
...
+    constructor() {
+        super(ErrorConstants.Messages.RESET_PASSWORD);
```
sisou:
```suggestion
        super(ErrorConstants.Messages.GOTO_RESET_PASSWORD);
```

client/src/PublicRequest.ts
```diff
@@ -234,6 +236,8 @@ export type KeyguardError = {
         CANCELED: 'CANCELED',
         // used to signal that the request expired
         EXPIRED: 'EXPIRED',
+        // Specifically used to trigger a redirect to a special import after returning to caller
+        RESET_PASSWORD: 'RESET_PASSWORD',
```
sisou:
```suggestion
        GOTO_RESET_PASSWORD: 'GOTO_RESET_PASSWORD',
```

Please also move this message next to the existing GOTO_CREATE message.

src/lib/ErrorConstants.js
```diff
@@ -17,6 +17,8 @@ const ErrorConstants = {
         CANCELED: 'CANCELED',
         // used to signal that the request expired
         EXPIRED: 'EXPIRED',
+        // Specifically used to trigger a redirect to a special import after returning to caller
+        RESET_PASSWORD: 'RESET_PASSWORD',
```
sisou:
```suggestion
        GOTO_RESET_PASSWORD: 'GOTO_RESET_PASSWORD',
```
Please also move this message next to the existing GOTO_CREATE message.

src/request/import/ImportWords.js
```diff
@@ -130,10 +131,14 @@ class ImportWords {
             window.location.hash = ImportWords.Pages.DOWNLOAD_LOGINFILE;
         });
 
-        $skipDownloadButton.addEventListener('click', e => {
-            e.preventDefault();
-            this._resolve(this._keyResults);
-        });
+        if (this._request.wordsOnly && this._request.requestedKeyPaths) {
+            $skipDownloadButton.remove();
```
sisou: Is this the correct condition? What do the requested key paths have to do with removing the skipDownloadButton? Do you mean to check for a given `expectedKeyId` instead?

src/request/import/index.html
```diff
@@ -95,7 +95,8 @@ <h1 data-i18n="import-heading-enter-recovery-words" class="nq-h1">Enter Recovery
                 <div class="page-body nq-card-body">
                     <div class="recovery-words"></div>
                     <p class="nq-label input-hint" data-i18n="import-words-hint">Press Tab to Jump to the next field</p>
-                    <p class="nq-label nq-red" data-i18n="import-words-error">This is not a valid account. Typo?</p>
+                    <p class="nq-label nq-red invalid" data-i18n="import-words-error">This is not a valid account. Typo?</p>
+                    <p class="nq-label nq-red wrong-seed-phrase" data-i18n="import-words-wrong-seed-phrase">These Recovery Words belong to a different account</p>
```
sisou: Should we capitalize "account" here? => "Account"?

src/translations/index.js
```diff
@@ -237,6 +240,7 @@ const TRANSLATIONS = {
         'passwordbox-repeat-password-long': 'Keine Übereinstimmung, bitte versuche es erneut',
         'passwordbox-repeat-password-short': 'Das Passwort ist zu kurz',
         'passwordbox-password-skip': 'Erstmal überspringen',
+        'passwordbox-reset-password': 'Mit Wiederherstellungswörtern zurücksetzen',
```
sisou, line 247: This german translation is propably too long for the passwordbox to fit into one line...

sisou (changes requested):

src/translations/index.js
nibhar: That is likely going to be an issue in many more places. If you don't have a good short suggestion I suggest doing it when the translation is actually being used.

sisou (approved):

src/components/PasswordBox.js
```diff
@@ -5,23 +5,41 @@
...
+ *      minLength?: number,
+ *      showResetPassword?: boolean
+ *  }} PasswordBoxOptions
+ *
+ *  @typedef {{
+ *      bgColor: string,
+ *      hideInput: boolean,
+ *      buttonI18nTag: string,
+ *      minLength: number,
+ *      showResetPassword: boolean
```
danimoh: Missing trailing comma.

src/components/PasswordBox.js
```diff
@@ -5,23 +5,41 @@
...
+ *  }} FullyDefinedPasswordBoxOptions
+ */
+
 class PasswordBox extends Nimiq.Observable {
     // eslint-disable-next-line valid-jsdoc
     /**
      * @param {HTMLFormElement} [$el]
-     * @param {{bgColor?: string, hideInput?: boolean, buttonI18nTag?: string, minLength?: number}} [options]
+     * @param {PasswordBoxOptions} [options]
```
danimoh: Instead of defining types `FullyDefinedPasswordBoxOptions` and `PasswordBoxOptions` I would just define the full options and call them `PasswordBoxOptions` and then use `Partial<PasswordBoxOptions>` here.

src/components/PasswordBox.js
```diff
@@ -36,11 +54,19 @@ class PasswordBox extends Nimiq.Observable {
         this._isInputValid = false;
 
         this.$el.addEventListener('submit', event => this._onSubmit(event));
+
+        if (options.showResetPassword) {
+            /** @type {HTMLLinkElement} */
```
danimoh: This is not a `<link>`.
```suggestion
            /** @type {HTMLAnchorElement} */
```

src/components/PasswordBox.css
```diff
@@ -109,3 +109,12 @@
...
     width: 100%;
 }
+
+.password-box .skip {
+    opacity: 1;
+    margin-top: -3.5rem;
+}
+
+.password-box.input-eligible .skip {
+    opacity: 0;
```
danimoh, line 119: [Optional]
For other elements here, the opacity is transitioned.

src/components/PasswordBox.css
```diff
@@ -109,3 +109,12 @@
     text-align: center;
     width: 100%;
 }
+
+.password-box .skip {
```
danimoh, line 113: I would prefer `reset-password` over `skip`.

src/components/PasswordBox.js
```diff
@@ -167,4 +204,5 @@ class PasswordBox extends Nimiq.Observable {
 
 PasswordBox.Events = {
     SUBMIT: 'passwordbox-submit',
+    RESET_PASSWORD: 'reset-password',
```
danimoh: As for `SUBMIT` the event is namespaced with `passwordbox-`, I would do the same here.

src/request/import/ImportApi.js
```diff
@@ -18,12 +19,16 @@ class ImportApi extends TopLevelApi {
         parsedRequest.requestedKeyPaths = this.parsePathsArray(request.requestedKeyPaths, ' requestedKeyPaths');
         parsedRequest.isKeyLost = this.parseBoolean(request.isKeyLost);
         parsedRequest.enableBackArrow = this.parseBoolean(request.enableBackArrow);
+        parsedRequest.wordsOnly = this.parseBoolean(request.wordsOnly);
+        parsedRequest.expectedKeyId = request.expectedKeyId ? await this.parseKeyId(request.expectedKeyId) : undefined;
```
danimoh: As we're actually using only the key id and not the whole `KeyInfo` and the property is called `expectedKeyId`, I would assign only the `.id` here.

Also: `parseKeyId` throws for an unknown key id which does not necessarily make sense in the import case.
I would suggest:

```suggestion
        parsedRequest.expectedKeyId = request.expectedKeyId ? request.expectedKeyId.toString() : undefined;
```

danimoh (changes requested):

src/components/PasswordBox.css
nibhar: This element should transition as well. Does it not transition on your machine?

src/components/PasswordBox.css
nibhar: `.skip` is defined with every single property needed for this link except for the 2 defined here. I could copy it and use it as `.reset-password`, but the duplication was something I thought was not desirable.
I can change it if you think it is beneficial, however I think it is more beneficial to have these styles in a single place.

src/request/import/ImportApi.js
nibhar: This is the same as an remove-key followed by an import, except for this very `expectedKeyId`. The `expectedKeyId` is only ever used for the specific purpose of resetting your password of an existing Account. 
If that account does not exist, or the keyId is something like `'-'` other tests in this request will always fail, as in this case you are only allowed to import with the recovery words which result in this specific id. 
The `expectedKeyId` is of course also expected to exist anyways, as it existed a few moments ago, but I think the parsing should not rely on that.

I could change it to 
```suggestion
        parsedRequest.expectedKeyId = request.expectedKeyId ? (await this.parseKeyId(request.expectedKeyId)).id : undefined;
```
and update the type accordingly, but I would keep the parsing as it is, because I think the throw is the desired behaviour here.

src/components/PasswordBox.css
danimoh: I just now saw that `.skip` is defined in `common.css` as common style, so makes sense to use it :+1:

src/request/import/ImportApi.js
danimoh: By "unknown key id" I meant a valid key id string for which no key exists in the database.
For this case, `parseKeyId` throws a `KeyNotFoundError`.
However, I think that for an import request it's totally acceptable that the key that I want to import is not currently present in the database. All we care about is that the `expectedKeyId` matches for the key resulting from the words the user entered.
I think the change I suggested serves this purpose.
If I'm not mistaken (correct me if I'm wrong) the following holds true: `(await this.parseKeyId(someId)).id === someId` for a Key existing in the new `KeyStore`.
For keys in the `AccountStore` this does not hold true which should however not be of relevance as such do not exist anymore after the migration process.
If this is the case, we should be able to replace the term `(await this.parseKeyId(request.expectedKeyId)).id` with `request.expectedKeyId.toString()` to remove the dependency on the existence of the key in the database.

nibhar: In the case the `expectedKeyId` is set this is basically a pseudo change-password. It is not really an import, but a re-import of an existing account. I do not see a benefit in opening it up to arbitrary keyIds or arbitrary strings for that matter. They could end up giving the user a usable interface, which he is unable to complete. If this is used as it is intended this will most likely never throw, as the request comes from a change-password to this specific keyId moments ago, which in turn did not throw for the same validation. If it is used as it was not intended it is going to throw.

What use-case do you have in mind for a non-existing keyId?

src/request/import/ImportWords.js
```diff
@@ -22,7 +22,7 @@
 
 class ImportWords {
     /**
-     * @param {Parsed<KeyguardRequest.ImportRequest>} request
+     * @param {Parsed<KeyguardRequest.ImportRequest | KeyguardRequest.ResetPasswordRequest>} request
```
danimoh, line 25: Might be that this is something different than `Parsed<KeyguardRequest.ImportRequest> | Parsed<KeyguardRequest.ResetPasswordRequest>` but I'm not sure.

types/Keyguard.d.ts
```diff
@@ -78,7 +78,13 @@ type Parsed<T extends KeyguardRequest.Request> =
...
             KeyguardRequest.ImportRequest,
+            'isKeyLost' | 'wordsOnly',
+            { isKeyLost: boolean, wordsOnly: boolean }
+        > :
+    T extends Is<T, KeyguardRequest.ResetPasswordRequest> ?
+        Transform<
+            KeyguardRequest.ResetPasswordRequest,
             'isKeyLost' | 'expectedKeyId' | 'wordsOnly',
-            { isKeyLost: boolean, expectedKeyId?: KeyInfo, wordsOnly: boolean }
+            { isKeyLost: boolean, expectedKeyId: KeyInfo, wordsOnly: boolean }
```
danimoh: I would still prefer `expectedKeyId` to be a string.

danimoh (approved):

src/request/import/ImportApi.js
danimoh: Now resolved by separate `ResetPasswordRequest` type.

types/Keyguard.d.ts
nibhar: I simply forgot that I wanted to include that suggestion. Is a string now.

src/request/import/ImportWords.js
nibhar: I don't think it is, at least not in this scenario.
```
request: Transform<KeyguardRequest.ImportRequest, ...> | Transform<KeyguardRequest.ResetPasswordRequest, ...>
```
is the resulting type.

mraveux (approved):

danimoh:
Three approvals, LGTM :)

sisou:
These changes have been released in @nimiq/keyguard-client v1.2.0.

PR #391, opened 2020-01-26
Tab-width selector

### LINKED ISSUE #204: Adding a selector for `tab-size` in the Keyguard Signature UI
Currently the Keyguard only accepts strings or Uint8Arrays as input, and I think this is fine.  But to help with ease of development, [NimiqWrapper](https://github.com/MatthewDLudwig/NimiqWrapper) allows the user to sign any javascript object.  This is done by using the `JSON.stringify(obj, NULL, "\t")` method, however this leads to the message being formatted with a lot of empty space due to the [default value of 8 spaces](https://www.w3schools.com/cssref/css3_pr_tab-size.asp) being used by all browsers (there is no way to change this in browser settings afaik, meaning the user has no say in how the message should look).

I propose adding a selector that allows setting (and remembering for future calls) the `tab-size` for the `textarea` (with id `message`) to either 2, 4, or 8 spaces.  This way the object can be displayed as the user wants (there is much debate for 2 vs 4, though no one I know likes 8), but the signature returned will be the same no matter which setting the user selects (since it's still tabs, no matter how it's displayed).

I envision a simple 3 switch button, where you can click on the number of spaces you want to be used.  Below is an image I threw together to show my idea:

image

sisou:
First proposed by @MatthewDLudwig in #368.

Add a tab-width selector component to the sign-message UI, when the message contains tabs (`\t`). The selected tab-width is stored in localStorage and automatically applied next time.

The component is styled according to Julians design (https://github.com/nimiq/keyguard/pull/368#issuecomment-549115231).

Closes https://github.com/nimiq/hub/issues/204.

### DIFF
--- a/src/request/sign-message/index.html
+++ b/src/request/sign-message/index.html
@@ -43,6 +43,7 @@
 
     <script defer src="../../lib/IqonHash.js"></script>
     <script defer src="../../lib/Iqons.js"></script>
+    <script defer src="../../components/TabWidthSelector.js"></script>
     <script defer src="../../components/Identicon.js"></script>
     <script defer src="SignMessageApi.js"></script>
     <script defer src="SignMessage.js"></script>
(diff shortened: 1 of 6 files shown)


src/request/sign-message/index.html
```diff
@@ -71,6 +73,9 @@ <h1 data-i18n="sign-msg-heading" class="nq-h1">Sign Message</h1>
 
             <div class="page-body nq-card-body">
                 <textarea id="message" readonly="readonly"></textarea>
+                <div class="tab-width">
+                    <div id="tabwidthselector"></div>
```
nibhar: Only a remark:
I liked the kebab-case we used previously throughout keyguard. Same for the translation string.

nibhar (approved):

danimoh (approved):

PR #395, opened 2020-03-06
Mark both file and words as exported after import

### LINKED ISSUE #394: Mark both file and words as exported when logging in
For the new wallet, the warnings about missing LoginFile and Recovery Words backups should only be displayed for created accounts, not for accounts that have been logged in.

So whenever an account is logged in, both the file and words exported flags have to be set to true.

sisou:
Resolves #394.

### DIFF
--- a/src/request/import/ImportFile.js
+++ b/src/request/import/ImportFile.js
@@ -156,8 +156,11 @@ class ImportFile {
             keyId: key.id,
             keyType: key.type,
             addresses,
+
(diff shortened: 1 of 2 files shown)


danimoh (approved):
Looks ok to me but the intention became just clear after reading the issue.
This should be added either as a comment in the code or in the commit message.

nibhar (approved):
Looks good to me.

danimoh (approved):

PR #398, opened 2020-04-21
Checkout timer always visible

danimoh:
Make the time in the checkout timer always visible / the timer always expanded.
Additionally, change the tooltip implementation to support content of arbitrary length.


sisou (approved):
:+1:

nibhar (approved):
Good to go I think.

PR #400, opened 2020-04-27
Checkout tooltip

danimoh:


src/components/PaymentInfoLine.js
```diff
@@ -21,20 +25,15 @@ class PaymentInfoLine { // eslint-disable-line no-unused-vars
      * @param {HTMLElement} [$el]
      */
     constructor(paymentInfo, $el) {
+        this.paymentInfo = paymentInfo;
         this.$el = PaymentInfoLine._createElement($el);
 
         /** @type HTMLElement */
         const $nimAmount = (this.$el.querySelector('.nim-amount'));
-        const nimAmount = NumberFormatting.formatNumber(Nimiq.Policy.lunasToCoins(paymentInfo.lunaAmount));
+        const nimAmount = NumberFormatting.formatNumber(Nimiq.Policy.lunasToCoins(paymentInfo.lunaAmount), 4);
```
sisou, line 33: Why `4` decimals here?

src/lib/FiatApi.js
```diff
@@ -0,0 +1,85 @@
+class FiatApi {
```
sisou, line 2: Is this basically a copy of the FiatApi from @nimiq/utils? If so, you can put a comment on top saying that's where it's from, so we have a reference for the implementation and if updates happen there.

src/lib/FiatApi.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * @readonly
+ * @enum { 'nim' }
+ * Crypto currencies supported by the coingecko api that are currently of interest for us.
+ */
+FiatApi.SupportedCryptoCurrency = {
+    NIM: /** @type {'nim'} */ ('nim'),
+};
```
sisou, line 44: I think you should add BTC already.

src/lib/NumberFormatting.js
```diff
@@ -23,24 +23,43 @@ class NumberFormatting { // eslint-disable-line no-unused-vars
...
-            .replace(integerPart, NumberFormatting.formatNumber(parseInt(integerPart, 10)));
+            // undefined to start with decimal count typical for this currency, e.g. 2 for eur
+            minimumFractionDigits: /** @type number | undefined */ (undefined),
+        };
+        let formatted;
+        let integers;
+        let relativeDeviation;
+
+        do {
+            formatted = value.toLocaleString([localeWithLatinNumbers, 'en-US'], formattingOptions)
```
sisou, line 47: For localization, the locale here should be adapted to what the I18n class is set to.

src/components/PaymentInfoLine.js
```diff
@@ -78,4 +76,170 @@ class PaymentInfoLine { // eslint-disable-line no-unused-vars
...
+        } else {
+            $vendorMarkupTemplate.remove();
+        }
+
+        /** @type {HTMLElement} */
+        const $effectiveRate = ($tooltip.querySelector('.effective-rate'));
+        // Fiat/crypto rate. Higher fiat/crypto rate means user is paying less crypto for the requested fiat amount
+        // and is therefore better for the user. Note: precision loss should be acceptable here.
+        const effectiveRate = fiatAmount / Nimiq.Policy.lunasToCoins(lunaAmount);
+        $effectiveRate.textContent = `${NumberFormatting.formatCurrency(effectiveRate, fiatCurrency, 0.0001)} / NIM`;
```
sisou, line 157: Why a relative deviation of 0.0001? Why not simply 0?

src/lib/NumberFormatting.js
```diff
@@ -23,24 +23,43 @@ class NumberFormatting { // eslint-disable-line no-unused-vars
     /**
      * @param {number} value
      * @param {string} currency
+     * @param {number} [maxRelativeDeviation = .1]
      * @param {string} [locale = navigator.language]
      * @returns {string}
      */
-    static formatCurrency(value, currency, locale = navigator.language) {
+    static formatCurrency(value, currency, maxRelativeDeviation = 0.1, locale = navigator.language) {
```
sisou: I don't understand how `maxRelativeDeviation` works. I think the default should be 0 (e.g. _no_ deviation) to get an exact number back, no? Why would I want a rounded number to two decimals (the starting `minimumFractionDigits`) when I not explicitly asked for it? And why is `minimumFractionDigits` limiting the decimals anyway? Shouldn't that be the job of `maximumFractionDigits`?

src/translations/index.js
```diff
@@ -250,6 +264,20 @@ const TRANSLATIONS = {
...
+        'payment-info-line-effective-rate': 'Effektiver Wechselkurs',
+        'payment-info-line-free-service': 'Nimiq bietet diesen Service kostenfrei an.',
+        'payment-info-line-total': 'Gesamt',
+        'payment-info-line-network-fee': 'Netzwerkgebühr',
+        'payment-info-line-paying-more': 'Du zahlst etwa %RATE_DEVIATION% mehr als zum aktuellen '
+                                       + 'Durchschnittswechselkurs (coingecko.com).',
+        'payment-info-line-paying-less': 'Du zahlst etwa %RATE_DEVIATION% weniger als zum aktuellen '
+                                       + 'Durchschnittswechselkurs (coingecko.com).',
+        'payment-info-line-actual-discount': 'Dein tatsächlicher Rabatt beträgt etwa %RATE_DEVIATION% zum aktuellen '
+                                           + 'Durchschnittswechselkurs (coingecko.com).',
```
sisou: Ich würde "Durchschnitts" hier weg lassen, im englischen steht auch nicht "average". Einfach "Wechselkurs".

sisou (changes requested):

src/components/PaymentInfoLine.js
```diff
@@ -78,4 +76,170 @@ class PaymentInfoLine { // eslint-disable-line no-unused-vars
...
+
+        /** @type {HTMLElement} */
+        const $total = ($tooltip.querySelector('.total'));
+        $total.textContent = `${NumberFormatting.formatNumber(Nimiq.Policy.lunasToCoins(lunaAmount))} NIM`;
+
+        // Note that in the Keyguard the fee is never undefined.
+        if (networkFee !== 0) {
```
nibhar, line 164: Out of curiosity:
Is there a specific reason as to why for one potentially not displayed property `vendorMarkup` `template` is used to replace in the positiv case while for `networkFee` it is the other way around, removing it in the negativ case?

src/lib/FiatApi.js
nibhar: I think BTC should be added when it is actually added as currently it would not serve any purpose and would therefore be unrelated to this PR.

src/lib/NumberFormatting.js
nibhar: I think `.1` is a good choice for the default. With `0` a value with 10 decimals formatted as a currency would still display 10 decimals by default which kind of defeats the purpose. Could also be something like 1e-(minimumFractionDigits) which I think should preserve the do loop to finish immediately for the default values. Still I think `.1` is perfectly fine.

nibhar (approved):
Pending @sisou comments I approve with just a minor question.

src/components/PaymentInfoLine.js
danimoh: To make it the same precision as used in `PaymentInfoLine.vue` in `vue-components`.

src/lib/FiatApi.js
danimoh: Yes, I think it should not be added now.

src/components/PaymentInfoLine.js
danimoh: The network fee sits in a root element which can be removed easily as a whole, so I'm doing that.
The elements in the template must be direct children of the grid, therefore I can't add an additional root element and am using a template instead.
I could also have put the elements directly as direct children of the grid, and remove them based on a common css class but I felt using a template is the more "intended" way of doing this.

src/lib/FiatApi.js
danimoh: Yes, basically. It enforces strict input sanitization though.
That might actually also in general be a good idea, wdyt?

src/components/PaymentInfoLine.js
danimoh: 0 would mean printing up to 20 decimal positions which is an unnecessary high precision, especially if floating point precision errors occurred in the number to be displayed.

src/lib/NumberFormatting.js
danimoh: @sisou
`minimumFractionDigits` and `maximumFractionDigits` both default to the decimal count typical for a currency. To get a higher precision, `minimumFractionDigits` has to be increased. `maximumFractionDigits` then defaults to `minimumFractionDigits` and does not have to be set manually.

@nibhar 
Not sure what you are suggesting in your last sentence but the method currently guarantees a max deviation, not a minimal decimal count, which is better I think and also what we discussed with overnice.

src/lib/NumberFormatting.js
sisou: Can the locale be adapted here too? Or what does the `en-US` do here?

sisou (approved):

src/lib/NumberFormatting.js
danimoh: Yes, it can be adapted. `en-us` is the fallback.

PR #401, opened 2020-05-25
i18n - initial setup

mraveux:
This PR add to the already existing i18n:
- transifex support
- french language
- `lang` cookie detection, fallback to browser language
- the `src/translations/index.js`file is now `.gitignore`ed because it's auto-generated by the `src/translations/i18n.sh` script executed by the `prebuild` npm script.
- npm scripts:
  - `i18n:pull` to retrieve translations from transifex (only languages min 95% translated)
  - `i18n:push` to push the source file (`src/translations/en.json`) to transifex

One question: Do you prefer translation files to be consistent between all applications (`.po` files) or the `.json` version of the Keyguard is enough (current version)?


src/lib/I18n.js
```diff
@@ -20,8 +20,37 @@ class I18n { // eslint-disable-line no-unused-vars
         this._fallbackLanguage = fallbackLanguage;
 
         // Re-enable next line (and remove line after) to enable internationalisation support
```
sisou: This comment can now be removed :wink:

src/lib/I18n.js
```diff
@@ -20,8 +20,37 @@ class I18n { // eslint-disable-line no-unused-vars
...
+
+    /**
+     * @returns {string} The detected language set in the 'lang' cookie. Fallback to the browser language.
+     */
+    static detectLanguage() {
+        const cookieMatch = document.cookie.match(new RegExp('(^| )lang=([^;]+)'));
+        const cookieLang = cookieMatch && decodeURIComponent(cookieMatch[2]);
+        const browserLang = navigator.language.split('-')[0];
+
+        return cookieLang || browserLang;
```
sisou: Calling `split` on a string does not take many CPU cycles, but it's still more work than _not_ doing it. So I suggest to only do it when actually necessary.

```suggestion

        return cookieLang || navigator.language.split('-')[0];
```

package.json
```diff
@@ -13,7 +15,10 @@
     "pr": "yarn typecheck && yarn lint && yarn test && yarn checkdeps && yarn checklangs && yarn build && cd client && yarn lint && yarn build",
     "watch": "tsc-watch --pretty --onSuccess \"yarn lint\"",
     "watchfix": "tsc-watch --pretty --onSuccess \"yarn lintfix\"",
-    "postinstall": "cd client && yarn"
+    "postinstall": "cd client && yarn",
+    "i18n:pull": "tx pull --all --parallel --no-interactive --force --minimum-perc 95",
```
sisou: Instead of setting the minimum percentage here, why not set it in the .tx/config file (where it's currently set to `0`)?

src/translations/de.json
```diff
@@ -0,0 +1,165 @@
+{
+    "_language": "Deutsch",
+
+    "language-changed": "The display language changed but some translations can't be updated. Do you want to reload the page to update thoses?",
```
sisou:
```suggestion
    "language-changed": "Die Anzeigesprache hat sich geändert, aber manche Übersetzungen können nicht aktualisiert werden. Möchtest du die Seite neu laden um auch diese zu aktualisieren?"
```

src/translations/i18n.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+FICHIERS=$(ls -1 src/translations/*.json)
```
sisou: What is `FICHIERS`? Is that french for `FILES`?

src/translations/i18n.sh
```diff
@@ -0,0 +1,19 @@
...
+OUTPUT_FILE=src/translations/index.js
+
+echo "/* eslint-disable */
+const TRANSLATIONS = {" > $OUTPUT_FILE
+
+for fichier in $FICHIERS; do
+    basename=${fichier##*/}
+    basename=${basename%.json}
```
sisou: What kind of formatting is going on here? Maybe explain it in a comment?

sisou (approved):

package.json
mraveux: It was just easier for testing to not set it in the .tx/config file. I change it now.

src/translations/i18n.sh
mraveux: oh, yes. 😅

src/translations/i18n.sh
mraveux: it's retrieving the basename without the file extension from the file path.
For example: `de` from `src/translations/de.json`.
I'll add a comment 👍

sisou: But, then you are adding `.json` again, so it's useless? Or don't I understand?

mraveux: starting from `src/translations/de.json`.
then basename = `de.json`
then basename = `de`
So i'm removing `.json`, not adding it 😉

sisou (approved):

sisou (approved):

danimoh (approved):

PR #404, opened 2020-07-28
I18n-ed Login File color & filename

mraveux:
This PR allow translations for Login File Color/Label & filename.

This is part of the i18n feedbacks from Overnice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "watchfix": "tsc-watch --pretty --onSuccess \"yarn lintfix\"",
     "postinstall": "cd client && yarn",
     "i18n:build-dictionary": "tools/i18n.sh",
-    "i18n:pull": "tx pull --all --parallel --no-interactive --force",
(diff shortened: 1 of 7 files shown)


sisou (approved):

package.json
```diff
@@ -15,7 +15,7 @@
     "watchfix": "tsc-watch --pretty --onSuccess \"yarn lintfix\"",
     "postinstall": "cd client && yarn",
     "i18n:build-dictionary": "tools/i18n.sh",
-    "i18n:pull": "tx pull --all --parallel --no-interactive --force",
+    "i18n:pull": "tx pull --all --parallel --no-interactive --force --mode sourceastranslation",
```
danimoh, line 18: :+1:

danimoh (approved):

package.json
danimoh: Unfortunately, this doesn't actually seem to work for me.

mraveux: Your problem was due to the `minimum_perc: 95` in the `.tx/config` file

PR #405, opened 2020-07-29
i18n/FlippableHandler: Automatically adapt the height of the element to fit it’s content

mraveux:
Because of i18n some translations caused the content to overflow on the bottom, and that broke design. With this commit now it automatically adapt the height of the `#rotation-container` element so that it fit its content and do not break design anymore.

_Btw, in the `FlippableHandler` you used a lot `window.setTimeout(..., 0)`.
Is there a reason you used this instead of `requestAnimationFrame` ?_

### DIFF
--- a/src/components/FlippableHandler.js
+++ b/src/components/FlippableHandler.js
@@ -48,12 +48,10 @@ class FlippableHandler {
                             window.setTimeout(() => $rotationContainer.classList.remove('disable-transition'), 10);
                         }, 0);
                     } else {
+                        $rotationContainer.classList.add('disable-transition');
+                        FlippableHandler._updateContainerHeight($newEl || undefined);
                         window.setTimeout(() => {
-                            $rotationContainer.classList.add('disable-transition');
-                            FlippableHandler._updateContainerHeight($newEl || undefined);
-                            window.setTimeout(() => {
-                                $rotationContainer.classList.remove('disable-transition');
-                            }, 0);
+                            $rotationContainer.classList.remove('disable-transition');
                         }, 0);
                     }
                 }
@@ -65,7 +63,7 @@ class FlippableHandler {
     /**
      * Update the height of the #rotation-container element to match its content.
      * The default behavior is to look for every visible `.page` element,
-     * and, if there is multiple, take the height of the higher.
+     * and, if there are multiple, take the height of the higher.
      * @param {Element} [$enforcedElement] - Enforce which element the function is taking the height from.
      *  Must be a child of `#rotation-container`
      */
@@ -78,7 +76,7 @@ class FlippableHandler {
             /** @type {Array<HTMLElement>} */
             const $pages = Array.from($rotationContainer.querySelectorAll('.page'));
             if ($pages && $pages.length > 0) {
-                const heights = $pages.map($el => ($el.offsetParent ? $el.clientHeight : 0));
+                const heights = $pages.map($el => $el.clientHeight);
                 const visiblePageHeight = Math.max(...heights);
                 $rotationContainer.style.height = visiblePageHeight > 0 ? `${visiblePageHeight}px` : '';
             }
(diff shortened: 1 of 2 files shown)


src/components/FlippableHandler.js
```diff
@@ -32,21 +33,60 @@ class FlippableHandler {
...
+        if ($enforcedElement && $rotationContainer.contains($enforcedElement)) {
+            $rotationContainer.style.height = `${$enforcedElement.clientHeight}px`;
+        } else {
+            /** @type {Array<HTMLElement>} */
+            const $pages = Array.from($rotationContainer.querySelectorAll('.page'));
+            if ($pages && $pages.length > 0) {
+                const $visiblePageHeight = $pages.reduce(
+                    (acc, $el) => ($el.offsetParent && $el.clientHeight > acc ? $el.clientHeight : acc),
+                    0,
+                );
```
sisou: You can find the biggest height with
```javascript
const heights = $pages.map(($el) => $el.offsetParent ? $el.clientHeight : 0);

Math.max.apply(Math, heights);
// or
Math.max(...heights);
```
which is better/simpler than `.reduce()`.

src/components/FlippableHandler.js
```diff
@@ -32,21 +33,60 @@ class FlippableHandler {
...
+                if ($visiblePageHeight > 0) {
+                    $rotationContainer.style.height = `${$visiblePageHeight}px`;
```
sisou: Would it make sense to revert it to its default (or `auto`), when no height was found?

sisou (changes requested):

src/components/FlippableHandler.js
```diff
@@ -32,21 +33,60 @@ class FlippableHandler {
...
                     if ($newEl && $newEl.classList.contains(classname)) {
                         $rotationContainer.classList.add('disable-transition');
+                        FlippableHandler._updateContainerHeight($newEl);
                         window.setTimeout(() => {
                             $rotationContainer.classList.toggle('flipped', $newEl.classList.contains(classname));
                             window.setTimeout(() => $rotationContainer.classList.remove('disable-transition'), 10);
                         }, 0);
+                    } else {
+                        window.setTimeout(() => {
+                            $rotationContainer.classList.add('disable-transition');
```
danimoh: Are the timeouts and setting / unsetting `disable-transition` necessary here?

src/components/FlippableHandler.js
```diff
@@ -32,21 +33,60 @@ class FlippableHandler {
...
                 }
             });
             FlippableHandler.flippableHandlerInitialised = true;
         }
     }
+
+    /**
+     * Update the height of the #rotation-container element to match its content.
+     * The default behavior is to look for every visible `.page` element,
+     * and, if there is multiple, take the height of the higher.
```
danimoh: `if there are multiple`

src/components/FlippableHandler.js
```diff
@@ -32,21 +33,56 @@ class FlippableHandler {
...
+    static _updateContainerHeight($enforcedElement) {
+        /** @type {HTMLElement} */
+        const $rotationContainer = (document.getElementById('rotation-container'));
+        if ($enforcedElement && $rotationContainer.contains($enforcedElement)) {
+            $rotationContainer.style.height = `${$enforcedElement.clientHeight}px`;
+        } else {
+            /** @type {Array<HTMLElement>} */
+            const $pages = Array.from($rotationContainer.querySelectorAll('.page'));
+            if ($pages && $pages.length > 0) {
+                const heights = $pages.map($el => ($el.offsetParent ? $el.clientHeight : 0));
```
danimoh: The check here isn't necessary as `clientHeight` is automatically 0 if the element is not attached to the DOM.

src/components/FlippableHandler.js
mraveux: The first timeout is not needed anymore, you're right.
But the second timeout as well as setting / unsetting `disable-transition` is needed to avoid animating the height change at the loading of the page.

src/components/FlippableHandler.js
mraveux: 😅

sisou (approved):
Superb! :+1:

danimoh (approved):

PR #408, opened 2020-08-05
Add `yarn serve` via `serve` dependecy + script in package.json

svub:
add `serve` devDep
building i18n dicts
serve -p 8000

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "build": "yarn i18n:build-dictionary && tools/build.sh",
     "test": "karma start",
+    "serve": "yarn i18n:build-dictionary && serve -p 8000",
     "typecheck": "tsc",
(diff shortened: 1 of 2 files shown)


sisou (approved):
The only concern I have is that it adds a huge load of dependencies. Maybe we can either find a simpler server package with less cruft, or build a simple nodejs server script that serves the current folder.

svub:
`serve` is just a devDep, should be no problem.

hst-m:
that is a lot of dependencies to serve a couple files, just write a small node script

hst-m:
I am counting 37 new external libraries in your PR for what, to do what Node.js can do in a couple lines of code?

svub:
Feel free to add the Node.js script.

danimoh (changes requested):
I also think a simple node script would be preferable.
Also the instructions in the Readme should be updated then.

sisou:
I created https://github.com/nimiq/keyguard/pull/411 that adds a basic static server NodeJS script. Please check it out. I tested it, it works the same.

PR #411, opened 2020-08-11
Add static server script to run `yarn serve`

sisou:
Add a basic NodeJS script without external dependencies to statically serve the Keyguard files. Adjusted from a public Github gist, added handling of non-existent `*/index.html` files.

Replaces #408.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ yarn
 
 Then you can:
 
+- serve the Keyguard locally with `yarn serve` (starts on port 8000).
 - run the build script with `yarn build [config]`.
 - run the tests with `yarn test`.
 - run the typechecker with `yarn typecheck`.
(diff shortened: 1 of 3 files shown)


mraveux (approved):
It works as expected. Nice.

svub (approved):
Cool, thanks Soeren! Just thinking, would be great to turn this into an NPM package. After all, this is something very generic and IMO, just adding a dev package that has this functionality should be the way to go instead of adding source code for dev purposes only.

sisou:
@svub generally I agree, however for the Keyguard which is a security relevant component, we need to try to depend on third-party/external packages as little as possible.

That's why we use our own scripts for language extraction, dependency checking, bundling and more tooling. And now serving for development ;)

PR #416, opened 2020-11-06
Add Github Actions workflow to replace Travis CI

sisou:


.github/workflows/complete-check.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
mraveux, line 11: We maybe want to explicitly set the NodeJs version?
```
- uses: actions/setup-node@v2
  with:
    node-version: '10'
    check-latest: true
```

mraveux (approved):
Works perfectly fine.

Just one suggestion, otherwise it's good 👍

.github/workflows/complete-check.yml
sisou: Can you give an argument of _why_ we would want to specify a Node version, instead of using the latest LTS release (which is the default behavior, currently Node 14)?

mraveux: ok i was wondering what was the default Node version in the runner since i didn't find it in the docs. But if it's already the latest LTS then i guess we don't need this. Unless there was a reason to use Node 10.x with travis.

sisou: I think it was a left-over from when a package failed to build in a newer version. Seems not relevant anymore, as all tests pass with default version also.

PR #418, opened 2020-11-11
Add HTLC and atomic swap signing

sisou:
This PR adds two new capabilities to the Keyguard:
- Signing HTLC funding and redeeming transactions for atomic swaps
- Signing HTLC refund transactions for when a swap fails

Signing a swap happens in two steps:
1. The display-relevant data and the involved key paths are handed to the top-level `SignSwap` request. Here the user reviews the swap data and enters their password. Then the Keyguard derives the required private keys at the respective keyPaths for NIM and BTC and stores them in SessionStorage (similar to how we do it during login for the key to derive addresses for discovery in the Hub).
2. After the user confirmed the swap by entering their password, the Hub confirms the swap with Fastspot and lets the Keyguard sign the actual transactions via the `swap-iframe` and it's `SignSwapTransactionsRequest`.

This two step process is necessary, because only when **confirming** a swap with Fastspot does Fastspot provide the necessary data for the HTLC transactions, such as the BTC target address, the BTC HTLC script, Fastspot's refund address (required to create a NIM HTLC) and the common hash (also called `hashRoot` in CoreJS). **However**, once a swap is _confirmed_ with Fastspot, its value counts against the user's 30-day limit. Plus, once confirmed, Fastspot directly locks up their part of the swap, reducing liquidity and incurring transactions fees, which we want to avoid if users have no intention of completing a swap. So we use the entering of the password as the final "yes, let's do this for real" signal of the user. But because we don't want to communicate with Fastspot from inside the Keyguard (no outside connections), we need to delegate the Fastspot interaction back to the Hub. Then, we don't want to require the user to enter their password again, so we store the private keys in SessionStorage to use by a Keyguard iframe. That's why it's a two-step process.

---
~~There is one detail which currently doesn't sit quite right with me about how I implemented it:
I believe there is currently no verification if what is signed in the iframe (in the second step) is actually what the user got displayed and approved in the first step. Meaning, the transactions that get signed with the user's keys _could_ be different from what was shown in the first step.~~

The only way around this, as I see it, is to give _all transaction_ info (except the unavailable ones that only become available with swap confirmation) to the Keyguard in the first step and store this data together with the private keys in SessionStorage. Then in the second step, only accept the missing information necessary to build valid transactions.

**Update:** this is is now how it's implemented.

~~Currently, our Keyguard can only be called from our Hub, where we don't do such weird things, so I think this is fine for now. Albeit obviously not ideal.~~

What do you think?


client/src/PublicRequest.ts
```diff
@@ -26,23 +26,34 @@ export type SingleKeyResult = {
...
-    sender: Uint8Array
-    senderType: Nimiq.Account.Type
-    recipient: Uint8Array
-    value: number
-    fee: number
-    validityStartHeight: number
-    recipientType?: Nimiq.Account.Type
-    data?: Uint8Array
+    keyPath: string,
+    senderLabel?: string,
```
sisou, line 32: I moved the `keyPath` and `senderLabel` into the `TransactionInfo` type to simplify types futher down and when altering/extending them for the swap types.

client/src/PublicRequest.ts
```diff
@@ -26,23 +26,34 @@ export type SingleKeyResult = {
...
+}
+
 export type BitcoinTransactionInput = {
     keyPath: string,
     transactionHash: string,
     outputIndex: number,
     outputScript: string,
     value: number,
+    witnessScript?: string,
+    type?: BitcoinTransactionInputType,
```
sisou: `witnessScript` is the HTLC script necessary for HTLC redeem and refund transactions. `type` is used to tell the Keyguard which of the two it is, as the witness data the Keyguard has to generate is different with otherwise same request data.

client/src/PublicRequest.ts
```diff
@@ -178,6 +185,73 @@ export type SignBtcTransactionRequest
...
+        {type: 'NIM'}
+            & Omit<TransactionInfo,
+                'senderType'
+                | 'recipient'
+                | 'recipientType'
+                | 'recipientLabel'
+                | 'validityStartHeight'
+                | 'data'
+                | 'flags'
+            >
```
sisou: We don't need quite a bit of information for only displaying the swap info visually, that's why it's all omitted here.

client/src/PublicRequest.ts
```diff
@@ -178,6 +185,73 @@ export type SignBtcTransactionRequest
...
+        balance: number, // Sats
+    },
+};
+
+// Used in swap-iframe
+export type SignSwapTransactionsRequest = {
+    swapId: string,
+    fund:
+        {type: 'NIM'}
+            & Omit<TransactionInfo, 'keyPath' | 'recipient' | 'recipientType' | 'data'>
```
sisou: - `keyPath` was already provided and the private key derived in the top-level request above
- `recipient` is generated automatically for HTLC creation transactions
- `recipientType` is clear, must be `HTLC`
- `data` is omitted because it's optional in `TransactionInfo`, but I am re-adding it below without the optionality.

src/lib/Key.js
```diff
@@ -28,7 +28,7 @@ class Key {
      * @returns {Nimiq.PublicKey}
      */
     derivePublicKey(path) {
-        return Nimiq.PublicKey.derive(this._derivePrivateKey(path));
+        return Nimiq.PublicKey.derive(this.derivePrivateKey(path));
```
sisou, line 31: I needed to make the `derivePrivateKey` method public, as I need to access the raw private key when signing a swap, to store it in SessionStorage for the iframe.

src/lib/RequestParser.js
```diff
@@ -188,9 +185,15 @@ class RequestParser { // eslint-disable-line no-unused-vars
      * @returns {Nimiq.Address}
      */
     parseAddress(address, name) {
+        if (address === 'CONTRACT_CREATION') {
+            return Nimiq.Address.CONTRACT_CREATION;
+        }
```
sisou, line 200: For a NIM HTLC creation transaction to work, the recipient must be passed as the _exact_ object reference to `Nimiq.Address.CONTRACT_CREATION`. This can obviously not be done through RPC, so a constant is used and replaced here with the correct object reference.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,234 @@
...
+
+// eslint-disable-next-line valid-jsdoc
+/**
+ * A mixin to add Bitcoin-related request parsers to a RequestParser class.
+ *
+ * @template {Constructor<RequestParser>} TBase
+ * @param {TBase} clazz
+ * @returns {typeof Clazz}
+ */
+function BitcoinRequestParserMixin(clazz) { // eslint-disable-line no-unused-vars
```
sisou: This file contains the Bitcoin-related request parsing methods that you already reviewed in the `bitcoin` branch PR. Only minor things changed in the actual methods, which I tried to mark with these review comments.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,234 @@
+/* global Nimiq */
+/* global Errors */
+/* global BitcoinJS */
+/* global BitcoinUtils */
+
+/**
+ * @template {{}} T
+ * @typedef {new (...args: any[]) => T} Constructor;
+ */
```
sisou, line 9: This method is taken from official TS docs: https://www.typescriptlang.org/docs/handbook/mixins.html

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,234 @@
...
+ * @returns {typeof Clazz}
+ */
+function BitcoinRequestParserMixin(clazz) { // eslint-disable-line no-unused-vars
+    class Clazz extends clazz {
+        /**
+         * @param {unknown} type
+         * @param {string} name
+         * @returns {'default' | 'htlc-redeem' | 'htlc-refund'}
+         */
+        parseInputType(type, name) {
```
sisou, line 31: This method is new.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,234 @@
...
+                        /** @type {number} */
+                        (this.parseNonNegativeFiniteNumber(input.outputIndex, false, `input[${index}].outputIndex`)),
+                    witnessUtxo: {
+                        script,
+                        value: Math.round(
+                            /** @type {number} */
+                            (this.parseNonNegativeFiniteNumber(input.value, false, `input[${index}].value`)),
+                        ),
+                    },
+                    type: this.parseInputType(input.type, `input[${index}].type`),
```
sisou, line 70: `type` is new.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,234 @@
...
+                    keyPath: this.parseBitcoinPath(input.keyPath, `input[${index}].keypath`),
+                    // Address added only for display
+                    // @ts-ignore Argument of type 'Uint8Array' is not assignable to parameter of type 'Buffer'.
+                    address: BitcoinJS.address.fromOutputScript(script, BitcoinUtils.Network),
+                };
+
+                if (input.witnessScript) {
+                    parsed.witnessScript = BitcoinJS.Buffer.from(Nimiq.BufferUtils.fromAny(input.witnessScript));
+                }
```
sisou, line 83: `witnessScript` parsing is new.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,234 @@
...
+                return parsed;
+            });
+        }
+
+        /**
+         * @param {unknown} paths
+         * @param {string} name - name of the property, used in error case only
+         * @returns {string[]}
+         */
+        parseBitcoinPathsArray(paths, name) {
```
sisou, line 94: This method is new.

src/lib/bitcoin/BitcoinUtils.js
```diff
@@ -114,4 +116,34 @@ class BitcoinUtils { // eslint-disable-line no-unused-vars
     static satoshisToCoins(satoshis) {
         return satoshis / BitcoinConstants.SATOSHIS_PER_COIN;
     }
+
+    /**
+     * @param {string | Uint8Array | Buffer} bytes
+     * @param {'BIP49' | 'BIP84'} bip
+     * @returns {string}
+     */
+    static addressBytesToAddress(bytes, bip) {
```
sisou, line 125: This is a left-over from when I did BTC HTLC validation in the Keyguard before the change to the two-step process.

Will probably be re-used once HTLC data is validated again.

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,168 @@
+/* global Nimiq */
+/* global BitcoinJS */
+/* global Errors */
+
+class HtlcUtils { // eslint-disable-line no-unused-vars
+    /**
+     * @param {unknown} data - Uint8Array
+     * @returns {NimHtlcContents}
+     */
+    static decodeNimHtlcData(data) {
```
sisou, line 12: The web-offline build of CoreJS does not yet include the HTLC class (https://github.com/nimiq/core-js/pull/572), so we need to recreate its `dataToPlain()` method ourselves to be able to parse NIM HTLC data.

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,168 @@
...
+            hash: hashRoot,
+            timeoutBlockHeight: timeout,
+        };
+    }
+
+    /**
+     * @param {unknown} script - Uint8Array
+     * @returns {BtcHtlcContents}
+     */
+    static decodeBtcHtlcScript(script) {
```
sisou, line 41: This was fun to figure out :smile:

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,168 @@
...
+        const error = new Errors.InvalidRequestError('Invalid BTC HTLC script');
+
+        if (!script || !(script instanceof Uint8Array) || !script.length) throw error;
+        const chunks = BitcoinJS.script.decompile(/** @type {Buffer} */ (script));
+        if (!chunks) throw error;
+        const asm = BitcoinJS.script.toASM(chunks).split(' ');
+
+        let branchesVerifiedIndividually = false;
```
sisou, line 50: A BTC HTLC script can be one of two versions. One where both the redeem and refund branch is verified independently, and one where only one verification happens at the end. This method can parse both. (Should probably have tests...)

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,168 @@
...
+            hash,
+            timeoutTimestamp,
+        };
+    }
+
+    /**
+     * @param {Buffer[]} witness
+     * @returns {Buffer}
+     */
+    static witnessStackToScriptWitness(witness) {
```
sisou, line 112: This is a re-implementation of a method included in BitcoinJS, which is sadly not exported anywhere.

https://github.com/bitcoinjs/bitcoinjs-lib/blob/533d6c2e6d0aa4111f7948b1c12003cf6ef83137/ts_src/psbt.ts#L1512

src/request/derive-btc-xpub/index.html
```diff
@@ -71,7 +71,8 @@
             <div class="page-body nq-card-body">
                 <img src="../../assets/icons/bitcoin.svg" class="bitcoin-logo">
 
-                <h1 class="nq-h1" data-i18n="derive-btc-xpub-heading">Add Bitcoin\nto your account</h1>
+                <h1 class="nq-h1" data-i18n="derive-btc-xpub-heading">Add Bitcoin
+                    to your account</h1>
```
sisou, line 75: This fixes a `\n` character appearing in the translated text on-screen.

src/request/sign-btc-transaction/SignBtcTransaction.js
```diff
@@ -216,43 +216,82 @@ class SignBtcTransaction {
...
+            throw new Error('Invalid or missing signature(s) for BTC transaction.');
         }
+
+        // Finalize
+        let i = 0;
+        for (const parsedInput of request.inputs) {
+            if (parsedInput.type === 'default') {
+                psbt.finalizeInput(i);
+            }
+            if (parsedInput.type === 'htlc-redeem' || parsedInput.type === 'htlc-refund') {
```
sisou: The type `htlc-redeem` is currently not used with the standard `SignBtcTransaction` request, but I though it can't hurt to include it, as the code is pretty much the same as for `htlc-refund`, except 2 lines in the middle.

src/request/sign-swap/SignSwapApi.js
```diff
@@ -0,0 +1,132 @@
...
+            parsedRequest.fund = {
+                type: 'NIM',
+                keyPath: this.parsePath(request.fund.keyPath, 'fund.keyPath'),
+                transaction: this.parseTransaction({
+                    validityStartHeight: 0, // Dummy
+                    data: new Uint8Array(78), // Dummy
+                    ...request.fund,
+                    flags: Nimiq.Transaction.Flag.CONTRACT_CREATION,
+                    recipient: 'CONTRACT_CREATION',
+                    recipientType: Nimiq.Account.Type.HTLC,
```
sisou, line 50: Here we set the fixed properties that we _know_ must be present when an HTLC is supposed to be created.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,283 @@
...
+
+        /** @type {{nim: string, btc: string[]}} */
+        const privateKeys = JSON.parse(storedPrivateKeys);
+
+        if (!privateKeys.nim) throw new Error('No NIM key stored in SessionStorage');
+        if (privateKeys.nim.length !== 64) throw new Error('Invalid NIM key stored in SessionStorage');
+        if (!privateKeys.btc) throw new Error('No BTC key list stored in SessionStorage');
+        if (!privateKeys.btc.length) throw new Error('No BTC keys stored in SessionStorage');
+        if (privateKeys.btc.some(key => !key)) throw new Error('Empty BTC key stored in SessionStorage');
+        if (privateKeys.btc.some(key => key.length !== 64)) throw new Error('Invalid BTC key stored in SessionStorage');
```
sisou: These checks are supposed to protect against corrupted or non-existent SessionStorage, and provide the correct error message to help us and the user understand what happened.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,283 @@
...
+
+        // Decode HTLC contents
```
sisou: All the commented code is technically TODO, but not required as long as we trust Fastspot and the Hub.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,283 @@
...
+            await loadNimiq();
+
+            const privateKey = new Nimiq.PrivateKey(Nimiq.BufferUtils.fromHex(privateKeys.nim));
+            const publicKey = Nimiq.PublicKey.derive(privateKey);
+
+            const transaction = this.parseTransaction({
+                ...request.fund,
+                flags: Nimiq.Transaction.Flag.CONTRACT_CREATION,
+                recipient: 'CONTRACT_CREATION',
+                recipientType: Nimiq.Account.Type.HTLC,
```
sisou: Again, we force the three properties that we _know_ must be present for HTLC creation.

client/src/KeyguardClient.ts
```diff
@@ -33,6 +34,9 @@ import {
     ResultByCommand,
     SignBtcTransactionRequest,
     DeriveBtcXPubRequest,
+    SignSwapRequest,
```
danimoh, line 37: ~~`SetupSwapRequest` or `SignSwapSetupRequest` or `ConfirmSwapRequest` could be better names.
Similarly for `SIGN_SWAP`, etc.
But would require many changes...~~

Nah, forget it. It's too many places to change.

client/src/PublicRequest.ts
danimoh: This can be typed more precisely:
```ts
export type BitcoinTransactionInput = {
    ...
} & ({
    type?: BitcoinTransactionInputType.DEFAULT,
} | {
    type: BitcoinTransactionInputType.HTLC_REDEEM | BitcoinTransactionInputType.HTLC_REFUND,
    witnessScript: string,
})
```

client/src/PublicRequest.ts
```diff
@@ -26,23 +26,34 @@ export type SingleKeyResult = {
...
-    validityStartHeight: number
-    recipientType?: Nimiq.Account.Type
-    data?: Uint8Array
+    keyPath: string,
+    senderLabel?: string,
+    sender: Uint8Array,
+    senderType: Nimiq.Account.Type,
+    recipient: Uint8Array,
+    recipientType?: Nimiq.Account.Type,
+    recipientLabel?: string,
```
danimoh: `recipientLabel` is ignored for `layout: 'checkout'` and `layout: 'cashlink'` and should therefore remain only in the definition of `layout: 'standard'`

client/src/PublicRequest.ts
```diff
@@ -26,23 +26,34 @@ export type SingleKeyResult = {
...
+    value: number,
+    fee: number,
+    validityStartHeight: number,
+    data?: Uint8Array,
     flags?: number,
 };
 
+export enum BitcoinTransactionInputType {
+    DEFAULT = 'default',
```
danimoh: Could also be `standard` to be more in line how the default for `layout` is called.

client/src/PublicRequest.ts
```diff
@@ -178,6 +185,73 @@ export type SignBtcTransactionRequest
     = SignBtcTransactionRequestStandard
     | SignBtcTransactionRequestCheckout;
 
+export type SignSwapRequest = SimpleRequest & {
+    swapId: string,
+    fund:
+        {type: 'NIM'}
```
danimoh, line 196: Inconsistent spacing around curly brackets in the newly added types.

client/src/KeyguardClient.ts
```diff
@@ -162,6 +170,11 @@ export class KeyguardClient {
         return this._iframeRequest<EmptyRequest, SimpleResult>(KeyguardCommand.MIGRATE_ACCOUNTS_TO_KEYS);
     }
 
+    public async signSwapTransactions(request: SignSwapTransactionsRequest): Promise<SignSwapTransactionsResult> {
+        const swapIframeBehavior = new SwapIFrameRequestBehavior();
+        return swapIframeBehavior.request(this._endpoint, KeyguardCommand.SIGN_SWAP_TRANSACTIONS, [request]);
```
danimoh: I would move this logic into `_iframeRequest`.
Otherwise, the type for `_iframeRequest` would need to be updated to exclude the `SignSwapTransactionsRequest` as it's in fact not able to handle that request type.

demos/Swap.html
```diff
@@ -0,0 +1,163 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Swap | Keyguard Demo</title>
+    <script src="../node_modules/@nimiq/rpc/dist/rpc.umd.js"></script>
+    <script>window.rpc = Rpc;</script>
+    <script src="../client/dist/KeyguardClient.umd.js"></script>
+    <script src="../node_modules/@nimiq/core-web/web-offline.js"></script>
+    <script src="../src/lib/bitcoin/vendor/NodeBuffer.js"></script>
+    <script src="../src/lib/bitcoin/vendor/BitcoinJS.js"></script>
```
danimoh: These dependencies are not found.

demos/Swap.html
```diff
@@ -0,0 +1,163 @@
...
+                address: 'NQ41 D9UN RC7G UMHE 4XA7 R1PL DNCF TH2N 7JQ8',
+                balance: 10000e5, // Luna
+            }, {
+                address: 'NQ39 X5FQ K296 97T0 VMJQ 1JJE SNLE 13B1 SJ4B',
+                balance: 10000e5, // Luna
+            }],
+            bitcoinAccount: {
+                balance: 0.001e8, // Sats
+            },
+        });
```
danimoh: Seems like this is still using the old request.

src/common.js
```diff
@@ -55,18 +55,24 @@ async function loadNimiq() {
...
 
+/**
+ * @typedef {{loadNimiq: boolean, whitelist: string[]}} Options
+ */
+
 /**
  * @template {KeyguardRequest.RedirectRequest} T
  * @param {Newable} RequestApiClass - Class object of the API which is to be exposed via RPC
- * @param {object} [options]
+ * @param {Partial<Options>} [opts]
```
danimoh, line 65: Why not mark `loadNimiq` and `whitelist` as `?` instead?

src/lib/bitcoin/BitcoinKey.js
```diff
@@ -23,27 +23,20 @@ class BitcoinKey {
...
-                break;
-            case BitcoinConstants.BIP.BIP84:
-                payment = BitcoinUtils.keyPairToNativeSegwit(keyPair);
-                break;
-            default:
-                throw new Errors.KeyguardError(`Invalid BIP: ${bip}`);
+        switch (BitcoinUtils.parseBipFromDerivationPath(path)) {
+            case BitcoinConstants.BIP.BIP49: address = BitcoinUtils.keyPairToNestedSegwit(keyPair).address; break;
+            case BitcoinConstants.BIP.BIP84: address = BitcoinUtils.keyPairToNativeSegwit(keyPair).address; break;
+            default: break;
```
danimoh, line 32: `default` case can be omitted.

src/lib/swap/HtlcUtils.js
nibhar: Agreed. Tests would be good here.

src/request/sign-btc-transaction/SignBtcTransaction.js
```diff
@@ -216,43 +216,82 @@ class SignBtcTransaction {
         // Sort outputs by value ASC, then address ASC
         outputs.sort((a, b) => (a.value - b.value) || (a.address < b.address ? -1 : 1));
 
-        try {
```
nibhar, line 219: What made this `try {}` block unnecessary?

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,423 @@
...
+                ? (redeemTx.value + redeemTx.fee + request.serviceRedeemingNetworkFee) / 1e8
+                : 0; // Should never happen, if parsing works correctly
+
+        if (!nimExchangeValue || !btcExchangeValue) {
+            throw new Errors.KeyguardError(
+                `UNEXPECTED: Swap values are invalid - NIM: ${nimExchangeValue}, BTC: ${btcExchangeValue}`,
+            );
+        }
+
+        const exchangeRate = Math.round(btcExchangeValue / nimExchangeValue * 1e8) / 1e8;
```
nibhar: Personally I would prefer to have this as either Btc -> Nim rate or Nim -> Btc rate depending on which direction I swap.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,418 @@
...
+            Nimiq.Policy.lunasToCoins(newNimBalance) * request.nimFiatRate,
+            request.fiatCurrency,
+        );
+        $newBtcBalance.textContent = NumberFormatting.formatNumber(BitcoinUtils.satoshisToCoins(newBtcBalance), 8);
+        const newBtcBalanceFiat = BitcoinUtils.satoshisToCoins(newBtcBalance) * request.btcFiatRate;
+        $newBtcBalanceFiat.textContent = NumberFormatting.formatCurrency(newBtcBalanceFiat, request.fiatCurrency);
+
+        // Draw distribution graph
+
+        const nimDistributionData = request.nimiqAddresses.map(addressInfo => {
```
nibhar, line 166: Could potentially have been made a component to make this class a little more lightweight.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,418 @@
...
+        )}`;
+        $swapBtcValueFiat.textContent = NumberFormatting.formatCurrency(
+            BitcoinUtils.satoshisToCoins(swapBtcValue) * request.btcFiatRate,
+            request.fiatCurrency,
+        );
+
+        $swapValues.classList.add(`${fundTx.type.toLowerCase()}-to-${redeemTx.type.toLowerCase()}`);
+
+        // Exchange rate
+        const nimExchangeValue = fundTx.type === 'NIM' // eslint-disable-line no-nested-ternary
```
nibhar, line 99: The whole exchange values and exchange rate topic is in my opinion in need of some explanatory comments.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,418 @@
...
+
+        I18n.translateDom($tooltip);
+
+        // All variables are in FIAT!
+
+        const myNimFee = Nimiq.Policy.lunasToCoins(fundTx.type === 'NIM' // eslint-disable-line no-nested-ternary
+            ? fundTx.transaction.fee
+            : redeemTx.type === 'NIM'
+                ? redeemTx.transaction.fee
+                : 0) * request.nimFiatRate;
```
nibhar, line 307: So all display within this tooltip is fiat exclusively. I don't know if I like this all that much, as this is also talking about network fees which are (rate-independent) values in crypto.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,418 @@
...
+            },
+        );
+    }
+
+    /**
+     * @param {Parsed<KeyguardRequest.SignSwapRequest>} request
+     * @param {number} exchangeAmount - In Luna or Satoshi, depending on which currency is funded
+     * @returns {HTMLDivElement}
+     */
+    _makeFeeTooltip(request, exchangeAmount) {
```
nibhar, line 251: Could also be moved to a component to clear this class up.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,283 @@
...
+
+        const storedPrivateKeys = sessionStorage.getItem(storageKey);
+        sessionStorage.removeItem(storageKey); // Delete storage
+
+        if (!storedPrivateKeys) throw new Error('No keys stored in SessionStorage');
+
+        /** @type {{nim: string, btc: string[]}} */
+        const privateKeys = JSON.parse(storedPrivateKeys);
+
+        if (!privateKeys.nim) throw new Error('No NIM key stored in SessionStorage');
```
nibhar: From what I take from the `SignSwapRequest` these do not have to be set specifically.
Only if the fund or the redeem tx is of type Nim this key will be present.

I realise that will always be the case currently, but in the other request it was handled with flexibility in mind while here it is not.

Same for `privateKeys.btc`.

src/request/swap-iframe/SwapIFrameApi.js
nibhar: As it is already there (just commented), why did you leave it out? It appears important to me to have this working.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,283 @@
...
+            psbt.addInputs(inputs);
+            // Add outputs
+            psbt.addOutput(output);
+
+            // Sign
+            const keyPair = BitcoinJS.ECPair.fromPrivateKey(
+                // @ts-ignore Argument of type 'import("...").Buffer' is not assignable to parameter of type 'Buffer'.
+                BitcoinJS.Buffer.from(privateKeys.btc[0], 'hex'),
+            );
+            psbt.signInput(0, keyPair);
```
nibhar, line 350: Is there only ever going to be 1 input?

nibhar (changes requested):
Good Job.

From my comments I only deem the one about the commented out todos essential.

src/request/sign-btc-transaction/SignBtcTransaction.js
sisou: I believe I removed it because I'm now sure that it will always work and there are now bugs :D

But I see that there are still lots of places where an Error could be thrown, and having a try-catch wrapper doesn't hurt, so I guess I'll re-add it.

src/request/swap-iframe/SwapIFrameApi.js
sisou: Yes, BTC HTLCs are only funded with one UTXO.

Doesn't _have_ to, it's just a BTC address, can have multiple UTXOs as well. But the way Fastspot works, there's only one.

src/request/sign-swap/SignSwap.js
sisou: I added comments, let me know if they explain it well enough. :slightly_smiling_face:

src/request/sign-swap/SignSwap.js
sisou: This is determined by the designs, which have been approved for a long time.

client/src/PublicRequest.ts
```diff
@@ -187,32 +187,52 @@ export type SignBtcTransactionRequest
...
-                | 'recipientLabel'
-                | 'validityStartHeight'
-                | 'data'
-                | 'flags'
-            >
-            & { senderLabel: string }
-        | {type: 'BTC', keyPaths: string[], value: number, fee: number },
-    redeem:
+        & Omit<TransactionInfo,
+            | 'recipient' // Only known in second step (in swap-iframe), derived from htlcData
```
danimoh, line 198: There's nothing before the first `|`.
Is that syntax actually working?

Same also at other places here in this file.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,228 @@
...
+
+            return /** @type {'htlc-redeem' | 'htlc-refund'} */ (typeString);
+        }
+
+        /**
+         * @param {unknown} inputs
+         * @returns {ParsedBitcoinTransactionInput[]}
+         */
+        parseInputs(inputs) {
+            if (!inputs || !Array.isArray(inputs)) {
```
danimoh: `!inputs` check already covered by `!Array.isArray()`.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,228 @@
...
+                throw new Errors.InvalidRequestError('inputs must not be empty');
+            }
+
+            // Construct inputs
+            return inputs.map((input, index) => {
+                const script = BitcoinJS.Buffer.from(Nimiq.BufferUtils.fromAny(input.outputScript));
+
+                /** @type {ParsedBitcoinTransactionInput} */
+                const parsed = {
+                    hash: Nimiq.BufferUtils.toHex(Nimiq.BufferUtils.fromAny(input.transactionHash)),
```
danimoh, line 64: Might want to validate the hash length

src/lib/bitcoin/BitcoinRequestParserMixin.js
danimoh: Can length / content of outputScript and witnessScript be validated?

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,228 @@
...
+            });
+        }
+
+        /**
+         * @param {unknown} paths
+         * @param {string} name - name of the property, used in error case only
+         * @returns {string[]}
+         */
+        parseBitcoinPathsArray(paths, name) {
+            if (!paths || !Array.isArray(paths)) {
```
danimoh: `!paths` part can again be omitted.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,228 @@
...
+            }
+            return path;
+        }
+
+        /**
+         * @param {string} path
+         * @returns {boolean}
+         */
+        isValidBitcoinPath(path) {
+            if (path.match(/^m(\/[0-9]+'?)*$/) === null) return false;
```
danimoh, line 137: Could be more restrictive: require at least purpose', coin_type', account' levels of Bip44 and make sure they are hardened.

Regex can optionally use `\d` instead of `[0-9]` and instead of `path.match(regex)`, `regex.test(path)` could be used.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,228 @@
...
+
+            // Overflow check.
+            const segments = path.split('/');
+            for (let i = 1; i < segments.length; i++) {
+                if (!Nimiq.NumberUtils.isUint32(parseInt(segments[i], 10))) return false;
```
danimoh: Indices must additionally be `< 2 ** 31` according to Bip32.

Loop can optionally also be simplified using `for (const ... of ...)` or `segments.some()`.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,228 @@
...
+                throw new Error(`${parameterName} is not a valid output`);
+            }
+
+            /** @type {KeyguardRequest.BitcoinTransactionOutput} */
+            const parsed = {
+                address: this.parseBitcoinAddress(
+                    /** @type {{address: unknown}} */ (output).address,
+                    `${parameterName}.address`,
+                ),
+                label: this.parseLabel(/** @type {{label: unknown}} */ (output).label),
```
danimoh: Parameter name not specified here.

src/lib/bitcoin/BitcoinRequestParserMixin.js
```diff
@@ -0,0 +1,228 @@
...
+         * @param {boolean} allowUndefined
+         * @param {string} parameterName
+         * @returns {KeyguardRequest.BitcoinTransactionOutput | undefined}
+         */
+        parseOutput(output, allowUndefined, parameterName) {
+            if (output === undefined && allowUndefined) {
+                return undefined;
+            }
+
+            if (!output || typeof output !== 'object') {
```
danimoh: `!output` check is redundant.
Same in `parseChangeOutput`.

src/lib/bitcoin/BitcoinUtils.js
```diff
@@ -114,4 +116,34 @@ class BitcoinUtils { // eslint-disable-line no-unused-vars
...
+        if (typeof bytes === 'string') {
+            bytes = Nimiq.BufferUtils.fromAny(bytes);
+        }
+
+        if (bip === BitcoinConstants.BIP.BIP49) {
+            return BitcoinJS.address.toBase58Check(
+                // @ts-ignore Argument of type 'Uint8Array | Buffer' is not assignable to parameter of type 'Buffer'.
```
danimoh, line 132: Could be avoided by actually ensuring that bytes is a Buffer via `bytes = Buffer.from()`

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,173 @@
...
+        const buf = new Nimiq.SerialBuffer(data);
+
+        const sender = Nimiq.Address.unserialize(buf).toUserFriendlyAddress();
+        const recipient = Nimiq.Address.unserialize(buf).toUserFriendlyAddress();
+        const hashAlgorithm = /** @type {Nimiq.Hash.Algorithm} */ (buf.readUint8());
+        const hashRoot = Nimiq.Hash.unserialize(buf, hashAlgorithm).toHex();
+        const hashCount = buf.readUint8();
+        const timeout = buf.readUint32();
+
+        if (!sender || !recipient || !hashRoot || !timeout) {
```
danimoh: I think this check is never true.

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,173 @@
...
+
+        if (!script || !(script instanceof Uint8Array) || !script.length) throw error;
+        // @ts-ignore Type 'import(...).Buffer' is not assignable to type 'Buffer'.
+        const chunks = BitcoinJS.script.decompile(BitcoinJS.Buffer.from(script));
+        if (!chunks) throw error;
+        const asm = BitcoinJS.script.toASM(chunks).split(' ');
+
+        let branchesVerifiedIndividually = false;
+
+        /* eslint-disable no-plusplus */
```
danimoh, line 52: Feel free to also globally allow usage of `++`.

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,173 @@
...
+
+        // Start redeem branch
+        if (asm[i] !== 'OP_IF') throw error;
+
+        // Check secret size
+        if (asm[++i] !== 'OP_SIZE' || asm[++i] !== (32).toString(16) || asm[++i] !== 'OP_EQUALVERIFY') throw error;
+
+        // Check hash
+        if (asm[++i] !== 'OP_SHA256' || asm[i + 2] !== 'OP_EQUALVERIFY') throw error;
+        const hash = Nimiq.BufferUtils.toHex(Nimiq.BufferUtils.fromHex(asm[++i]));
```
danimoh, line 63: Checking hash length might be a good idea.

Same for all other data read from the script.

And I assume the `fromHex` followed by `toHex` is for verification purposes?

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,173 @@
...
+        if (asm[++i] !== 'OP_ELSE') {
+            branchesVerifiedIndividually = true;
+            if (asm[i] !== 'OP_EQUALVERIFY' || asm[++i] !== 'OP_CHECKSIG' || asm[++i] !== 'OP_ELSE') throw error;
+        }
+
+        // Check timeout
+        // Bitcoin HTLC timeouts are backdated 1 hour, to account for Bitcoin's
+        // minimum age for valid transaction locktimes (6 blocks).
+        // @ts-ignore Argument of type 'Buffer' is not assignable to parameter of type 'Buffer'
+        const timeoutTimestamp = BitcoinJS.script.number.decode(BitcoinJS.Buffer.from(asm[++i], 'hex')) + (60 * 60);
```
danimoh, line 81: Does 60 * 60 stand for 60 minutes with 60 seconds each, so is the timestamp in seconds?

src/lib/swap/HtlcUtils.js
```diff
@@ -0,0 +1,173 @@
...
+     */
+    static witnessStackToScriptWitness(witness) {
+        /** @type {number[]} */
+        let buffer = [];
+
+        /**
+         * @param {Buffer} slice
+         */
+        function writeSlice(slice) {
+            buffer = buffer.concat([...slice.subarray(0)]);
```
danimoh, line 120: What's the `subarray` for?
Could also use spread syntax: `buffer = [...buffer, ...slice];`

src/request/sign-swap/SignSwapApi.js
```diff
@@ -0,0 +1,142 @@
...
+                }),
+                /** @type {string} */
+                senderLabel: (this.parseLabel(request.fund.senderLabel, false, 'fund.senderLabel')),
+            };
+        } else if (request.fund.type === 'BTC') {
+            parsedRequest.fund = {
+                type: 'BTC',
+                inputs: this.parseInputs(request.fund.inputs),
+                recipientOutput: {
+                    value: this.parsePositiveInteger(request.fund.recipientOutput.value),
```
danimoh: `parameterName` not set.
Might also apply to some other things parsed here.

src/request/sign-swap/SignSwapApi.js
```diff
@@ -0,0 +1,142 @@
...
+        const nimAddress = parsedRequest.fund.type === 'NIM' // eslint-disable-line no-nested-ternary
+            ? parsedRequest.fund.transaction.sender.toUserFriendlyAddress()
+            : parsedRequest.redeem.type === 'NIM'
+                ? parsedRequest.redeem.transaction.recipient.toUserFriendlyAddress()
+                : ''; // Should never happen, if parsing works correctly
+        if (!parsedRequest.nimiqAddresses.some(addressInfo => addressInfo.address === nimAddress)) {
+            throw new Errors.InvalidRequestError(
+                'The address details of the NIM address doing the swap must be provided',
+            );
+        }
```
danimoh, line 137: Could also validate that the balance is `>=` the funding amount.

src/request/sign-btc-transaction/SignBtcTransaction.js
```diff
@@ -158,65 +158,70 @@ class SignBtcTransaction {
...
+        try {
+            // For BIP49 (nested SegWit) inputs, a redeemScript needs to be added to inputs
+            for (const input of request.inputs) {
+                if (BitcoinUtils.parseBipFromDerivationPath(input.keyPath) !== BitcoinConstants.BIP.BIP49) continue;
+
+                // Add redeemScripts for BIP49 inputs
+                const keyPair = btcKey.deriveKeyPair(input.keyPath);
+                const output = BitcoinUtils.keyPairToNativeSegwit(keyPair).output;
+                if (!output) {
+                    TopLevelApi.setLoading(false);
```
nibhar: Unnecessary as the catch block also does it.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,492 @@
...
+            BitcoinUtils.satoshisToCoins(swapBtcValue) * request.btcFiatRate,
+            request.fiatCurrency,
+        );
+
+        $swapValues.classList.add(`${fundTx.type.toLowerCase()}-to-${redeemTx.type.toLowerCase()}`);
+
+        // Exchange rate
+        const nimExchangeValue = fundTx.type === 'NIM' // eslint-disable-line no-nested-ternary
+            // When the user funds NIM, the service receives the HTLC balance - their network fee.
+            ? (fundTx.transaction.value - request.serviceFundingNetworkFee) / 1e5
```
danimoh: Could use `lunasToCoins`.
Same below.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,492 @@
...
+            : redeemTx.type === 'NIM'
+                // When the user redeems NIM, the service lost the HTLC balance + their network fee.
+                // The transaction value is "HTLC balance - tx fee", therefore the "HTLC balance"
+                // is the transaction value + tx fee.
+                ? (redeemTx.transaction.value + redeemTx.transaction.fee + request.serviceRedeemingNetworkFee) / 1e5
+                : 0; // Should never happen, if parsing works correctly
+
+        const btcExchangeValue = fundTx.type === 'BTC' // eslint-disable-line no-nested-ternary
+            // When the user funds BTC, the service receives the HTLC balance - their network fee.
+            ? (fundTx.recipientOutput.value - request.serviceFundingNetworkFee) / 1e8
```
danimoh: Could use `satoshisToCoins`.
Same below.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,492 @@
...
+        const theirBtcFee = BitcoinUtils.satoshisToCoins(fundTx.type === 'BTC'
+            ? serviceFundingNetworkFee
+            : serviceRedeemingNetworkFee) * request.btcFiatRate;
+
+        const theirExchangeFee = fundTx.type === 'NIM'
+            ? Nimiq.Policy.lunasToCoins(serviceExchangeFee) * request.nimFiatRate
+            : BitcoinUtils.satoshisToCoins(serviceExchangeFee) * request.btcFiatRate;
+
+        const theirExchangeFeePercentage = NumberFormatting.formatNumber(
+            request.serviceExchangeFee / (exchangeAmount - request.serviceExchangeFee) * 100,
```
danimoh: Can use `serviceExchangeFee` instead of `request.serviceExchangeFee`

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,492 @@
...
+        if (request.fund.type === 'NIM') {
+            const privateKey = key.derivePrivateKey(request.fund.keyPath);
+            privateKeys.nim = privateKey.toHex();
+        }
+
+        if (request.fund.type === 'BTC') {
+            const keyPairs = request.fund.inputs.map(input => btcKey.deriveKeyPair(input.keyPath));
+            const privKeys = keyPairs.map(keyPair => Nimiq.BufferUtils.toHex(
+                /** @type {Buffer} */ (keyPair.privateKey),
+            ));
```
danimoh: Could also directly do `toString('hex')` on the `Buffer`.
Same below.

src/request/sign-btc-transaction/SignBtcTransaction.js
```diff
@@ -226,17 +231,62 @@ class SignBtcTransaction {
...
             if (!psbt.validateSignaturesOfAllInputs()) {
-                throw new Error('Invalid or missing signature(s).');
+                throw new Error('Invalid or missing signature(s) for BTC transaction.');
             }
 
             // Finalize
-            psbt.finalizeAllInputs();
+            let i = 0;
+            for (const parsedInput of request.inputs) {
+                if (parsedInput.type === 'default') {
```
nibhar: I assume there are more types than the 3 that we are currently using. If we are likely going to be using more in the future i would prefer this being a switch:
```js
switch (parsedInput.type) {
  case 'default':
    psbt.finalizeInput(i);
    break;
  case 'htlc-redeem':
  case 'htlc-refund':
    ...
    break;
  default:
    throw new Errors.KeyguardError('Unsupported input type');
}
```
I realise that the Parser takes care of different types and the default arm is technically unreachable.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,492 @@
...
+                                        witnessUtxo: {
+                                            ...input.witnessUtxo,
+                                            script: Nimiq.BufferUtils.toHex(input.witnessUtxo.script),
+                                        },
+                                    })),
+                                },
+                            } : {}
+                        ),
+                    },
+                }),
```
danimoh: This can be simplified passing a custom replacer to `JSON.stringify` which serializes transactions and buffers.

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,492 @@
...
+
+    run() {
+        // Go to start page
+        window.location.hash = SignSwap.Pages.CONFIRM_SWAP;
+    }
+}
+
+SignSwap.Pages = {
+    CONFIRM_SWAP: 'confirm-swap',
+};
```
danimoh, line 467: As there is only a single page, is this necessary?

src/request/sign-swap/index.html
```diff
@@ -0,0 +1,156 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <link rel="icon" href="/favicon.ico">
+    <meta http-equiv="Content-Security-Policy"
+          content="default-src 'self' 'unsafe-eval';
+          img-src http: https: blob: data:;
+          child-src 'self' blob:;
+          style-src 'self';
+          font-src 'self';
+          connect-src 'self' https://api.coingecko.com">
```
danimoh: I think coingecko is currently not used in the sign-swap request.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+     * @param {KeyguardRequest.SignSwapTransactionsRequest} request
+     * @returns {Promise<KeyguardRequest.SignSwapTransactionsResult>}
+     */
+    async signSwapTransactions(state, request) {
+        const storageKey = SwapIFrameApi.SESSION_STORAGE_KEY_PREFIX + request.swapId;
+
+        const storedData = sessionStorage.getItem(storageKey);
+        sessionStorage.removeItem(storageKey); // Delete storage
+
+        if (!storedData) throw new Error('No swap stored in SessionStorage');
```
danimoh, line 23: In error case should delete all stored swap data to avoid the swap id being brute-forceable.
Note that the session storage can be iterated with https://developer.mozilla.org/en-US/docs/Web/API/Storage.key in combination with `sessionStorae.length` to find the entries to delete. But apparently also `Object.keys(sessionStorage)` seems to work.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+            if (privateKeys.nim.length !== 64) throw new Error('Invalid NIM key stored in SessionStorage');
+        }
+
+        if (request.fund.type === 'BTC' || request.redeem.type === 'BTC') {
+            if (!privateKeys.btc) throw new Error('No BTC key list stored in SessionStorage');
+            if (!privateKeys.btc.length) throw new Error('No BTC keys stored in SessionStorage');
+            if (privateKeys.btc.some(key => !key)) throw new Error('Empty BTC key stored in SessionStorage');
+            if (privateKeys.btc.some(key => key.length !== 64)) {
+                throw new Error('Invalid BTC key stored in SessionStorage');
+            }
```
danimoh, line 39: These could be combined.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+                            }),
+                        ),
+                    },
+                } : {}
+            ),
+        };
+
+        // Parse request
+        /** @type {Parsed<KeyguardRequest.SignSwapTransactionsRequest>} */
+        // @ts-ignore Missing the following properties: fund, redeem
```
danimoh: To avoid this, you can probably also first set `fund` and `redeem` as variables and then later do
```js
const parsedRequest = {
    fund,
    redeem,
    swapId: request.swapId,
};
```

src/request/sign-swap/SignSwap.js
```diff
@@ -0,0 +1,492 @@
...
+
+        try {
+            sessionStorage.setItem(
+                SignSwapApi.SESSION_STORAGE_KEY_PREFIX + request.swapId,
+                JSON.stringify({
+                    keys: privateKeys,
+
+                    // Serialize request to store in SessionStorage
+                    request: {
+                        ...request,
```
nibhar: In my opinion a little overboard with inlining.

i.e
```js
let copiedRequest = Object.assign({}, request);
if (copiedRequest.fund.type === 'NIM') {
    copiedRequest.fund.transaction = request.fund.transaction.toPlain();
}
```
Or something along those lines would have worked too. Potentially even in place. In my opinion is easier to read and shorter, as well.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+                transactionHash: tx.getId(),
+                raw: tx.toHex(),
+            };
+        }
+
+        return result;
+    }
+}
+
+SwapIFrameApi.SESSION_STORAGE_KEY_PREFIX = 'swap_id_';
```
nibhar: This should be put into `Constants.js` as it is duplicated for `SignSwapApi.js` as the "same" constant.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+            const htlcDetails = HtlcUtils.decodeBtcHtlcScript(request.fund.htlcScript);
+
+            if (htlcDetails.refundAddress !== storedRequest.fund.refundAddress) {
+                throw new Errors.InvalidRequestError(
+                    'BTC HTLC refund address must be same as given in top-level request',
+                );
+            }
+
+            const htlcAddress = BitcoinJS.payments.p2wsh({
+                // @ts-ignore Type 'Uint8Array' is not assignable to type 'Buffer'.
```
danimoh, line 143: Are you still getting this warning even though you're doing `Buffer.from`?

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+            if (!privateKeys.btc.length) throw new Error('No BTC keys stored in SessionStorage');
+            if (privateKeys.btc.some(key => !key)) throw new Error('Empty BTC key stored in SessionStorage');
+            if (privateKeys.btc.some(key => key.length !== 64)) {
+                throw new Error('Invalid BTC key stored in SessionStorage');
+            }
+        }
+
+        // Deserialize stored request
+        /** @type {Parsed<KeyguardRequest.SignSwapRequest>} */
+        const storedRequest = {
```
nibhar: Again this could be done with less inlining for a more readable code in my opinion.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+                    'Redeem validityStartHeight must be lower than HTLC timeout block height',
+                );
+            }
+
+            parsedRequest.redeem = {
+                type: 'NIM',
+                htlcDetails,
+                htlcData: request.redeem.htlcData,
+                htlcAddress: this.parseAddress(request.redeem.htlcAddress, 'redeem.htlcAddress')
+                    .toUserFriendlyAddress(),
```
danimoh, line 129: Could also be calculated from the actual contract creation tx, which would thus need to be passed as input.

src/request/swap-iframe/SwapIFrameApi.js
nibhar: I am not too fond of this `@ts-ignore`.
Maybe assemble the parts first and the `parsedRequest` last?

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+            };
+        }
+
+        if (request.fund.type === 'BTC' && storedRequest.fund.type === 'BTC') {
+            const htlcDetails = HtlcUtils.decodeBtcHtlcScript(request.fund.htlcScript);
```
danimoh, line 134: The timeout timestamp could be roughly compared to our local time.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+            const tx = psbt.extractTransaction();
+
+            /** @type {KeyguardRequest.SignedBitcoinTransaction} */
+            result.btc = {
+                transactionHash: tx.getId(),
+                raw: tx.toHex(),
+            };
+        }
+
+        if (parsedRequest.redeem.type === 'NIM' && storedRequest.redeem.type === 'NIM') {
```
nibhar, line 299: Pretty much the same as the funding except for about 2 lines. Can maybe be simplified into a single block.

nibhar (commented):
Only comments about some minor things, which also in part come down to preference.

Nice work!

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+            }
+
+            // Sort outputs by value ASC, then address ASC
+            outputs.sort((a, b) => (a.value - b.value) || (a.address < b.address ? -1 : 1));
+
+            // Construct transaction
+            const psbt = new BitcoinJS.Psbt({ network: BitcoinUtils.Network });
+
+            // Add inputs
+            // @ts-ignore Argument of type 'Uint8Array' is not assignable to parameter of type 'Buffer'.
```
danimoh, line 267: Does `Buffer.from` help?

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+
+            // Construct outputs
+            const outputs = [{
+                address: parsedRequest.fund.htlcAddress,
+                value: storedRequest.fund.recipientOutput.value,
+            }];
+
+            // Add change output
+            if (storedRequest.fund.changeOutput) {
+                // The address is set in SignSwap after password entry.
```
danimoh, line 252: Could also validate this, doesn't hurt.
Same below.

src/request/swap-iframe/index.js
```diff
@@ -0,0 +1,9 @@
+/* global runKeyguard */
+/* global SwapIFrameApi */
+
+runKeyguard(SwapIFrameApi, {
+    loadNimiq: false,
```
danimoh, line 5: I think it's always required?
So could also be true.

src/translations/nl.json
```diff
@@ -0,0 +1,219 @@
+{
```
danimoh: As Dutch is not being released as part of this PR, it should also not be added here.

tools/build.sh
```diff
@@ -120,7 +121,7 @@ fi
 output "🧐  Validating BitcoinJS file integrity"
 
 # For bitcoinjs-lib v5.2.0 and Buffer v5.6.0
-bitcoinjs_hashsum="9cb30b789ff1a79cc1925d7f49ee827b2e9eb89531a58c51ca6ca996b15597eb  src/lib/bitcoin/BitcoinJS.js"
+bitcoinjs_hashsum="5a02ae59046a7ee4f386f5828097aa21bdf52657acca4ae472d5e89e8332ba43  src/lib/bitcoin/BitcoinJS.js"
```
danimoh, line 124: What changed about BitcoinJS?

client/src/PublicRequest.ts
```diff
@@ -178,6 +185,93 @@ export type SignBtcTransactionRequest
...
+            | 'recipientType' // Must be HTLC
+            | 'recipientLabel' // Not used
+            | 'data' // Only known in second step (in swap-iframe)
+            | 'flags' // Must be CONTRACT_CREATION
+        >
+        & { senderLabel: string }
+    ) | (
+        {type: 'BTC'}
+        & Omit<BitcoinTransactionInfo,
+            | 'recipientOutput' // Replaced below
```
danimoh: Could also use `Transform` helper type from `Keyguard.d.ts`

client/src/PublicRequest.ts
```diff
@@ -187,32 +187,52 @@ export type SignBtcTransactionRequest
...
-                | 'recipientType'
-                | 'recipientLabel'
-                | 'validityStartHeight'
-                | 'data'
-                | 'flags'
-            >
-            & { senderLabel: string }
-        | {type: 'BTC', keyPaths: string[], value: number, fee: number },
-    redeem:
+        & Omit<TransactionInfo,
```
danimoh, line 197: Using `Pick` instead of `Omit` would probably be shorter and give a better idea of what's actually included.

client/src/PublicRequest.ts
```diff
@@ -187,32 +187,52 @@ export type SignBtcTransactionRequest
...
+            | 'recipient' // Only known in second step (in swap-iframe), derived from htlcData
+            | 'recipientType' // Must be HTLC
+            | 'recipientLabel' // Not used
+            | 'data' // Only known in second step (in swap-iframe)
+            | 'flags' // Must be CONTRACT_CREATION
+        >
+        & { senderLabel: string }
+    ) | (
+        {type: 'BTC'}
+        & Omit<BitcoinTransactionInfo,
```
danimoh: The fee is not validated in the Keyguard which is not ideal.

client/src/PublicRequest.ts
```diff
@@ -187,32 +187,52 @@ export type SignBtcTransactionRequest
...
-                | 'senderType'
-                | 'senderLabel'
-                | 'recipientType'
-                | 'validityStartHeight'
-                | 'data'
-                | 'flags'
-            >
-            & { recipientLabel: string }
-        | {type: 'BTC', keyPaths: string[], value: number, fee: number },
+        & Omit<TransactionInfo,
```
danimoh, line 217: Could use `Pick` again.

danimoh (changes requested):

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -0,0 +1,385 @@
...
+                sender: parsedRequest.redeem.htlcAddress,
+            });
+
+            const signature = Nimiq.Signature.create(privateKey, publicKey, transaction.serializeContent());
+
+            /** @type {KeyguardRequest.SignatureResult} */
+            result.nim = {
+                publicKey: publicKey.serialize(),
+                signature: signature.serialize(),
+            };
```
danimoh, line 306: I was wondering why the transaction proof including the pre-image is not constructed in the KG already?

And in the hub, currently a constant pre-image seems to be used.
How does that work?

src/common.js
sisou: Because they do exist on the `Options` type (as it has defaults) and would otherwise also be `?` when used in the rest of the function and would require type annotations to force that they are indeed defined.

client/src/PublicRequest.ts
sisou: It's working, I was also surprised. I'm doing it for alignment.

src/lib/bitcoin/BitcoinUtils.js
sisou: That only changes the type error to `Type Buffer of import(...) is not assignable to type Buffer`, because the type `Buffer` that is required is the NodeJS `Buffer`, but the library for the browser is a different `Buffer`.

src/lib/swap/HtlcUtils.js
sisou: Yes and yes.

src/lib/swap/HtlcUtils.js
sisou: It's to convert the Buffer into an Uint8Array.

src/request/swap-iframe/SwapIFrameApi.js
sisou: I had them separate for the different type of error message, which I though useful for debugging.

src/request/swap-iframe/index.js
sisou: Not if we use this iframe for other swaps in the future as well, e.g EUR-BTC.

tools/build.sh
sisou: Had to add the `ECPair` to the exports.

client/src/PublicRequest.ts
sisou: How should we validate the fee?

src/lib/swap/HtlcUtils.js
sisou: Test are added in the `oasis` branch: b04b3648f9aaa6f34619a2fd5f5ccfe770db724b

src/request/sign-swap/SignSwap.js
sisou: Made into a component in the `oasis` branch: ca5d6ba963c5c271b2d74ccea5c5ac728293f7bc

src/request/sign-swap/SignSwap.js
sisou: Its now extracted in the `oasis` branch: 2950cd5884eb0c9cad55aa1c9ca574810bbbefb2

src/request/sign-swap/SignSwap.js
sisou: I was actually quite proud of this construct :laughing:

demos/Swap.html
sisou: The swap is not easily demo-able anymore. I removed the Swap.html demo page.

src/lib/bitcoin/BitcoinKey.js
sisou: eslint expects a default case

src/lib/bitcoin/BitcoinRequestParserMixin.js
sisou: An invalid tx hash will fail to get converted into a BTC transaction input and throw eventually, likely after password-entry, but before the request would successfully return.

src/lib/bitcoin/BitcoinRequestParserMixin.js
sisou: I believe not. They can be totally custom.

src/lib/bitcoin/BitcoinRequestParserMixin.js
sisou: This check is taken from Nimiq.ExtendedPrivateKey. The method `parseBitcoinPath` (the only user of this method here) actually checks for BIP49 vs BIP84.

sisou: I think the Keyguard should make no assumption about the derivation path used, to also allow non-standard usage. Does the BTC Ledger app restrict the paths?

src/lib/bitcoin/BitcoinRequestParserMixin.js
sisou: Again, this is taken straight from the Nimiq lib and I think it's fine this way.

src/lib/bitcoin/BitcoinRequestParserMixin.js
sisou: Actually, `!output` checks for `null`, which is `typeof 'object'`.

src/lib/swap/HtlcUtils.js
sisou: You are right, the length of the buffer is checked before, so all deserialization will succeed.

src/lib/swap/HtlcUtils.js
sisou: I added a hash length check. And yes, the fromHex(toHex()) is for verification that it is valid hex data.

src/request/sign-swap/SignSwapApi.js
sisou: I believe all other checks have the parameter name set.

src/request/sign-swap/SignSwap.js
sisou: I think it's convention, and is also used for the DOM selector.

src/request/swap-iframe/SwapIFrameApi.js
sisou: Yes, because the Buffer expected and the polyfilled Buffer for browsers are different.

sisou: No, actually the error about Buffer and the polyfilled Buffer being not the same.

rezarostam2453 (approved):

src/request/swap-iframe/SwapIFrameApi.js
sisou: I think that is unnecessary because if the HTLC address actually isn't this address, the swap does not get past step 2 in the wallet.

And as a precaution against the HTLC address getting faked in the request: a whole transaction can just as easily be faked - the Keyguard cannot validate if a transaction is included in the chain or not.

src/request/swap-iframe/SwapIFrameApi.js
sisou: No, because the browser-polyfilled "Buffer" (which is a Uint8Array internally) is still not the same as the _type_ `Buffer`.

sisou: And making the type use `Buffer` instead, just shifts the TS errors from this here usage to the creation of the object, plus it would be using non-browser types.

sisou: Maybe this would be cleaner though. I opened https://github.com/nimiq/keyguard/issues/432 to investigate this after this is merged.

client/src/PublicRequest.ts
sisou: The type has 11 properties, so `Pick` would be longer.

It would become this:
```
        & Pick<TransactionInfo,
            | 'keyPath'
            | 'senderLabel'
            | 'sender'
            | 'senderType'
            | 'value'
            | 'fee'
            | 'validityStartHeight'
        >
```

But it would loose all the explanation of why the other ones are omitted. I prefer the `Omit<>` method.

client/src/PublicRequest.ts
sisou: See https://github.com/nimiq/keyguard/pull/418#discussion_r672147730

PR #422, opened 2021-03-14
feat: add account label on login file

mariofriz:
This PR adds a new feature as suggest here: [Be able to add a label on a Login Files ](https://github.com/nimiq/wallet/issues/25)

## Description

When you have multiple accounts, it can quickly become messy with all the login files. Colors help to organize but still can be confusing. There is already a possibility to rename accounts in the wallet, so why not re-use that label and add it to the login file.

## Changes
- Add label to login file
- Move date to the side of the login file
- Use "color" when account label is not customized
- Use account label when user has set it
- Update demo files to work with new code

## Screenshots
Standard "color" account:

Custom labeled account:


sisou:
Hello Mario, 

Thank you very much for this PR! 👍 

We have discussed it with @julianbauer internally and would like to do a few additions:
- Julian will propose a few design adaptions to the Login File layout
- We want to also store the account label in the QR code, so that it can be applied when logging in with the file

We'll get back to you over the next days with updates!

julianbauer:
Hi Mario, thanks a lot for your proposal! We made some small adjustments and ended up with the following design, which I think is very close to what you had in mind: https://www.figma.com/file/ttceUyaZDAVAIpJVJwKfci/21-03-23-Named-Login-Files?node-id=0%3A1

sisou:
@julianbauer I believe the only things that change to what you shared is
- Removal of stars around the key
- Moving the date to next to the QR code and coloring it opaque blue instead of white

Anything else?

julianbauer:
@sisou exactly, the key's size is slightly adjusted as well. Please note that on the darker login files (red, light blue, brown, purple, indigo) the date is set in white, also the opacities of the text are different depending on the color, ranging from 35-45% – this is to keep max visual distance from the main content in the center, while still having enough contrast to be readable.

mariofriz:
@julianbauer Thanks for the mockups, I'll get on it.
@sisou You mentioned: 

> We want to also store the account label in the QR code, so that it can be applied when logging in with the file

I also saw that popping up on Telegram or Discord recently, I'll see if that easily doable.

mariofriz:
New changes:
- remove the stars and reduce key size
- reposition and resize date text
- change date text opacity depending on background

Here's the result:
nimiq-login-files

@julianbauer IMO, changing the date text to dark blue for lighter background values look a bit weird and makes it pop a lot more than on darker backgrounds.

sisou:
Hey Mario, thank you for your work!
I tried to commit on this branch, but it seems you disabled that repo PR-target-repo owners can push on your branch. Can you enable that?
https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

julianbauer:
> IMO, changing the date text to dark blue for lighter background values look a bit weird and makes it pop a lot more than on darker backgrounds.

@mariofriz The reasoning behind this was to maintain a good contrast delta to the main content on light background colors. But I agree with your assessment, and the version you posted works fine for me as well!

mariofriz:
<strike>😞 I don't have that option. </strike>

@sisou Should be good now!

sisou:
I have also prepared the Hub to receive the label and apply it during login, that branch will be updated and merged when this PR is merged and a new KeyguardClient is published.
https://github.com/nimiq/hub/commit/6bad290a02ccf60ad7fe6cc4feedee8a2a1cde03

sisou (approved):

src/request/import/ImportFile.js
```diff
@@ -161,6 +170,7 @@ class ImportFile {
         const result = [{
             keyId: key.id,
             keyType: key.type,
+            ...(this._label ? { keyLabel: this._label } : {}),
```
mariofriz, line 176: question: could this be simplified?

```js
keyLabel: this_label ? this_label : null, 
// or null-coalescing if allowed
// or || would probably be fine also
```

src/lib/LoginFile.js
```diff
@@ -83,24 +92,64 @@ class LoginFile {
...
+        this._ctx.fillStyle = 'rgba(255, 255, 255, 0.9)';
+        this._ctx.font = `600 36px ${LoginFile.FONT_FAMILY}`;
+        this._ctx.fillText(this._getLabelForDisplay(), x, y, LoginFile.WIDTH - LoginFile.BORDER_WIDTH * 4);
+        this._setFont(); // reset font
+    }
+
+    /**
+     * @returns {string}
+     */
+    _getLabelForDisplay() {
```
mariofriz, line 148: praise: good idea, better than just having the color name!

src/components/FileImporter.js
```diff
@@ -103,7 +103,9 @@ class FileImporter extends Nimiq.Observable {
 
             // Make sure the data size is correct
             if (bytes.byteLength !== KeyStore.ENCRYPTED_SECRET_SIZE
-                && bytes.byteLength !== KeyStore.ENCRYPTED_SECRET_SIZE_V2) {
+                && bytes.byteLength !== KeyStore.ENCRYPTED_SECRET_SIZE_V2
+                // If an extension is present, it must consist of at least one length byte and one character byte
+                && bytes.byteLength < KeyStore.ENCRYPTED_SECRET_SIZE + 2) {
```
mariofriz: suggestion: I saw there are some import/export tests, I think it would be worth adding a test for new format and old format to ensure backwards compatibility

mariofriz (commented):
@sisou Thanks for the cleanup, looks much better now. 
I left a few comments, please let me know what you think.

src/request/import/ImportFile.js
sisou: The alternative would be returning an `undefined`. I think omitting the key instead is nicer.

src/lib/LoginFile.js
sisou: Thanks! Translations are a fickle thing :D

src/components/FileImporter.js
sisou: Added!

sisou:
I rebased on top of `master`, as there were conflicts with the LoginFile tests, and this way all tests are run now (which was fixed in `master`).

mariofriz (commented):
👍

src/lib/LoginFile.js
```diff
@@ -83,24 +92,64 @@ class LoginFile {
...
+
+        return this._label;
     }
 
     _drawWarningText() {
         const x = LoginFile.WIDTH / 2;
         const y = LoginFile.HEIGHT - 86 - 2;
         this._ctx.fillStyle = 'white';
-        this._ctx.fillText('Keep safe and confidential.', x, y);
+        this._ctx.fillText('Keep safe and confidential', x, y);
```
mraveux, line 173: I think i would keep this dot :)

mraveux (approved):
Nice job Mario and Sisou, nicely done. 👌

danimoh (approved):
Thank you for your contribution.
Great work.

I added some fixes and improvements on top:
- Fix for truncation of Unicode astral symbols consisting of a surrogate character pair.
- If available, use the actual account label in the filename instead of the default label.
- Fix grammatical gender of colors in account label for Spanish translation.
- Stricter label validation by checking for the expected byte length and utf8 validity.
- Render emojis in account label white instead of their original color.
- Remove the stars in the `LoginFileAnimation` as they have been removed in the `LoginFile` too.
- Improve the demo a bit.

@julianbauer 
This is the filtering which I'm applying to the account label to render emojis in white to avoid that they clash with the background color and draw too much attention.
nimiq-login-file-emojis

This PR is now good to go from my side.
Feel free to merge if you agree with my changes.

PR #423, opened 2021-03-23
DownloadLoginFile: increase download dialog detection wait time

danimoh:
Occasionally (but rarely) the download dialog seems to need a longer time to open.
I'm proposing to increase the wait time to 1.5s.
As browsers usually play a download animation during that time, this should still be an acceptable wait time.
Maybe even 2s would still be acceptable.

### DIFF
--- a/src/components/DownloadLoginFile.js
+++ b/src/components/DownloadLoginFile.js
@@ -180,8 +180,8 @@ class DownloadLoginFile extends Nimiq.Observable {
                     // ("Save as" always opens a dialog.)
 
                     // If no download dialog opens, consider the download successful after a delay. On Macs it takes a
-                    // bit longer until the dialog gets focused and the window blurred, therefore waiting 1s.
-                    window.setTimeout(resolve, 1000);
+                    // bit longer until the dialog gets focused and the window blurred, therefore waiting 1.5s.
+                    window.setTimeout(resolve, 1500);
                 }
                 window.addEventListener('blur', resolve, { once: true });
                 this._cancelDownload = reject;


sisou:
This doesn't fix the issue for all browsers and situations, though. I have noticed that when I download the loginfile on my Android Chrome, and I already have a file named the same, it asks me if I want to overwrite that, but that prompt does not prevent the success-redirect.

Is 1.5s sufficient to trigger the download everytime you tested, also on the original reporter's laptop?

danimoh:
It is very hard to tell whether 1.5s is enough as even with 1s the issue occurs only very rarely.
For me, yesterday was the first time this happened to me.
Generally, 1.5s are likely _not_ enough in all cases but already better than 1s.

Typical occasions I could imagine for this to occur are when your computer is busy:
- high CPU load by other tabs or programs
- high memory usage such that your OS first needs to swap some memory to disk
- garbage collection kicking in on Android
- ...

Thus, increasing the time is nothing but a quick fix.
But I also don't really have a better idea right now.
One possible solution we discussed some time ago was implementing a service worker which detects whether the file to download has been requested.
However, Safari will potentially struggle with this again and I believe most browsers already start the download in the background while the download dialog is open but no download destination has been chosen yet.

sisou:
I believe the only _real_ solution is a separate "Continue" button that only becomes enabled when we think the user downloaded it (the event that we currently use for triggering the redirect).

danimoh:
I am also fine with introducing that additional click.
It's essentially the `maybe-downloaded` flow we already have.

mraveux:
It's not the best user experience imo, but since there's not a lot of other options anyway this seems to be the best compromise. I'm totally fine with it.

PR #424, opened 2021-03-29
Always require confirmation click when downloading LoginFile

sisou:
Replaces #423.

### DIFF
--- a/src/components/DownloadLoginFile.js
+++ b/src/components/DownloadLoginFile.js
@@ -31,10 +31,7 @@ class DownloadLoginFile extends Nimiq.Observable {
         this.$downloadButton = (this.$el.querySelector('.download-button'));
 
         /** @type {HTMLButtonElement} */
-        const $continueButton = (this.$el.querySelector('.continue'));
-
-        /** @type {HTMLButtonElement} */
-        const $backToDownloadButton = (this.$el.querySelector('.back-to-download'));
+        this.$continueButton = (this.$el.querySelector('.continue'));
 
         /** @type {SVGElement} */
         this.$longTouchIndicator = (this.$el.querySelector('.long-touch-indicator'));
@@ -42,12 +39,11 @@ class DownloadLoginFile extends Nimiq.Observable {
         this.$loginfile.addEventListener('mousedown', e => this._onMouseDown(e));
         this.$loginfile.addEventListener('touchstart', () => this._onTouchStart());
         this.$downloadButton.addEventListener('click', () => this._onDownloadStart());
-        $continueButton.addEventListener('click', () => {
+        this.$continueButton.addEventListener('click', () => {
             this._onDownloadEnd();
             // Remove previously added classes after a short delay to restore the initial state.
             window.setTimeout(this._reset.bind(this), 300);
         });
-        $backToDownloadButton.addEventListener('click', () => this.$downloadButton.click());
     }
 
     /**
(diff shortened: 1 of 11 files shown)


src/components/DownloadLoginFile.js
```diff
@@ -195,12 +197,10 @@ class DownloadLoginFile extends Nimiq.Observable {
             // hash change is delayed by the .maybe-downloaded confirmation mechanism.
             if (!document.hasFocus()
                 || (BrowserDetection.isIOS() && BrowserDetection.iOSVersion()[0] > 12)) {
-                this.$el.classList.add('maybe-downloaded');
                 this.fire(DownloadLoginFile.Events.INITIATED);
```
danimoh: This should now always be fired, which means the `if` condition can be removed.

src/components/DownloadLoginFile.js
```diff
@@ -195,12 +197,10 @@ class DownloadLoginFile extends Nimiq.Observable {
             // hash change is delayed by the .maybe-downloaded confirmation mechanism.
```
danimoh: The comments in this method are not accurate anymore.

src/translations/fr.json
```diff
@@ -122,10 +122,9 @@
     "identicon-selector-generate-new": "Nouveaux avatars",
 
     "download-loginfile-download": "Télécharger",
-    "download-loginfile-successful": "Téléchargement réussi ?",
-    "download-loginfile-tap-and-hold": "Appuyez et maintenez l'image pour la télécharger",
-    "download-loginfile-continue": "Continuez lorsque vous êtes prêt",
-    "download-loginfile-download-again": "Télécharger à nouveau",
+    "download-loginfile-successful": "Téléchargement réussi?",
```
danimoh: This was not a typo.
French people apparently have a space before the question mark.
We're using non-breaking spaces to avoid breaking the question mark into the next line.

src/components/DownloadLoginFile.js
```diff
@@ -244,8 +244,14 @@ class DownloadLoginFile extends Nimiq.Observable {
         }
     }
 
+    _maybeDownloaded() {
+        this.$el.classList.add('maybe-downloaded');
```
danimoh, line 249: The `maybe-downloaded` class is currently not being used anymore.

danimoh (changes requested):
- This PR does not implement the final design yet.
- `back-to-download` is still referenced in `Export.css` and can be removed.

src/translations/fr.json
mariofriz: I can confirm there is a space in french before the ?

Yukiioz: Yes, there is one. And please use a non-breaking space for it :)

src/components/DownloadLoginFile.js
sisou: The class is now used again for changing the buttons in this state.

mraveux (approved):
Simple and effective. Nice 👍 
I would've loved to have a transition between the buttons, with the text fading out or smth. But i guess that since we're opening a popup the "jump" is not meant to be seen so we don't really need one.

danimoh (approved):
Looks good.
I agree with Matheo that animating between the two states would be nice.
Note that not always a popup dialogue is opening, specifically if the user has his browser setup to download files immediately.
In this case he'd be seeing the animation.

For the `fallback-download` case I would personally not unveil the `Continue` text in the continue button as it breaks the layout to a degree.

sisou:
I created issue #426 for the button transitions, as it's too much work now to figure this out and not worth the delay in releasing this fix.

PR #425, opened 2021-04-07
QR video scanning

### LINKED ISSUE #364: Login File Scanning with camera
Add camera support to the Keyguard QR Scanner for scanning Login Files via camera.

sisou:
Fixes #364

This PR extends the `QrScanner` library with the ability to scan live video for QR codes, converted from https://github.com/nimiq/qr-scanner.

It then adds a `QrVideoScanner` component to encapsulate the scanner. Also includes already-translated strings, taken from the nimiq/vue-components translations.

~~There is a workaround for the "BarcodeScanner", which is sadly not yet recognized by Typescript and thus cannot be used for types.~~

### DIFF
--- a/src/lib/QrScanner.js
+++ b/src/lib/QrScanner.js
@@ -14,7 +14,7 @@
 
 class QrScanner {
     static async hasCamera() {
-        if (!navigator.mediaDevices) return Promise.resolve(false);
+        if (!navigator.mediaDevices) return false;
         // note that enumerateDevices can always be called and does not prompt the user for permission. However, device
         // labels are only readable if served via https and an active media stream exists or permanent permission is
         // given. That doesn't matter for us though as we don't require labels.
--- a/src/lib/QrScanner.js
+++ b/src/lib/QrScanner.js
@@ -383,7 +383,14 @@ class QrScanner {
             /** @type {string | undefined} */
             let result;
             try {
-                result = await QrScanner.scanImage(this.$video, this._scanRegion, this._qrEnginePromise, this.$canvas);
+                result = await QrScanner.scanImage(
+                    this.$video,
+                    this._scanRegion,
+                    this._qrEnginePromise,
+                    this.$canvas,
+                    /* fixedCanvasSize */ true,
+                    /* alsoTryWitoutScanRegion */ false,
+                );
             } catch (error) {
                 if (!this._active) return;
                 const errorMessage = error.message || error;
--- a/src/lib/QrScanner.js
+++ b/src/lib/QrScanner.js
@@ -212,7 +212,7 @@ class QrScanner {
     /**
      * @param {HTMLImageElement | HTMLVideoElement | string} imageOrFileOrUrl
      * @param {RectArea} [scanRegion]
-     * @param {Promise<Worker>} [givenEngine] // TODO: Or BarcodeDecoder
+     * @param {Promise<Worker>} [givenEngine] // TODO: Or BarcodeDetector
      * @param {HTMLCanvasElement} [givenCanvas]
      * @param {boolean} [fixedCanvasSize = false]
      * @param {boolean} [alsoTryWithoutScanRegion = false]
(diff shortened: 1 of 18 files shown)


src/lib/QrScanner.js
```diff
@@ -1,141 +1,565 @@
...
+            }
+            this.$video.srcObject = null;
+            this._offTimeout = undefined;
+        }, 300);
+    }
+
+    /**
+     * @param {HTMLImageElement | HTMLVideoElement | string} imageOrFileOrUrl
+     * @param {RectArea} [scanRegion]
+     * @param {Promise<Worker>} [givenEngine] // TODO: Or BarcodeDetector
```
mraveux: Is there really no way to use `BarcodeDetector` in JsDoc?

src/lib/QrScanner.js
```diff
@@ -1,141 +1,565 @@
...
+            if (exact) {
+                const exactFacingMode = { exact: facingMode };
+                constraintsToTry.forEach(constraint => {
+                    constraint.facingMode = exactFacingMode;
+                });
+            } else {
+                constraintsToTry.forEach(constraint => {
+                    constraint.facingMode = facingMode;
                 });
             }
```
mraveux: Couldn't this be simplified to the following?
```
constraintsToTry.forEach(constraint => {
    constraint.facingMode = exact ? { exact: facingMode } : facingMode;
})
```

src/lib/QrScanner.js
```diff
@@ -1,141 +1,564 @@
...
+     * @returns {Promise<MediaStream>}
+     */
+    _getMatchingCameraStream(constraintsToTry) {
+        if (!navigator.mediaDevices || constraintsToTry.length === 0) {
+            return Promise.reject('Camera not found.');
+        }
+        return navigator.mediaDevices.getUserMedia({
+            video: constraintsToTry.shift(),
+        }).catch(() => this._getMatchingCameraStream(constraintsToTry));
+    }
```
mraveux: Wouldn't it be cleaner as an async function?
```
async _getMatchingCameraStream(constraintsToTry) {
    if (!navigator.mediaDevices || constraintsToTry.length === 0) {
        return Promise.reject('Camera not found.');
    }
    try {
        return navigator.mediaDevices.getUserMedia({
            video: constraintsToTry.shift(),
        });
    } catch (e) {
        return await this._getMatchingCameraStream(constraintsToTry);
    }
}
```

mraveux (commented):
Looks nice, as always 👍 
The only thing that prevent me from approving the PR is that, as discussed, it doesn't seem to work as intended on Chrome android (and maybe other browsers but that'll require more testing) but that should not be a big issue to fix imo.

src/lib/QrScanner.js
sisou: It's not included in TS. And probably won't be until there's wider support: https://caniuse.com/?search=barcodedetector

src/lib/QrScanner.js
sisou: With a slight change, that the `navigator.mediaDevices.getUserMedia()` function must be `await`ed, but yes, that works.

mraveux (approved):
Perfect! It now works perfectly on mobile devices as well 😃

danimoh (approved):
Nice work :+1:
I applied a variety of improvements and a fix for scanning invalid LoginFiles / arbitrary QR codes, see commit history.
Please have a look at the changes.
I also rebased the branch onto master to avoid later merge conflicts as I changed code touching the current LoginFile validation that already supports the encoded wallet label.
As the QrScanner and QrVideoScaner are largely adapted from the `qr-scanner` lib and `@nimiq/vue-components`, I took the freedom, to also add me as a co-author to the commits adding these chunks.
However, I'm also not upset if you object to this and want to force-push it away.

src/lib/QrScanner.js
sisou: Daniel added a type definition, so now it's included in the types.

PR #431, opened 2021-07-18
Fiat currency updates

danimoh:


src/lib/CurrencyInfo.js
```diff
@@ -0,0 +1,241 @@
...
+    'VES', // sub-unit rarely used
+]);
+
+/**
+ * Cached auto-generated CurrencyInfos such that they do not need to be recalculated.
+ * @private
+ * @readonly
+ * @type {{[currencyAndLocale: string]: CurrencyInfo}}
+ */
+CurrencyInfo.CACHED_AUTO_GENERATED_CURRENCY_INFOS = {};
```
sisou, line 279: Do we really need to cache currency infos? How long does it actually take to calculate?

src/lib/NumberFormatting.js
```diff
@@ -85,6 +109,11 @@ class NumberFormatting { // eslint-disable-line no-unused-vars
             case 'cny':
                 return 'zh';
             default:
+                // Return the country from the currency code which is typically (but not necessarily) the first two
+                // letters (see https://en.wikipedia.org/wiki/ISO_4217#National_currencies), in the hope that it
+                // coincides with a locale.
+                // TODO oftentimes this results in the wrong locale, e.g. ARS (Argentinan Peso) -> AR (Arabic),
+                //  CAD (Canadian Dollar) -> CA (Catalan). Can we come up with a better heuristic?
```
sisou, line 116: For these cases, they need to be added to the excemption list above, like for EUR and GBP.

sisou (approved):
I think the tests from @nimiq/utils should also be ported.

mraveux (approved):
Nice job, well done. And seems to work perfectly 👍 

(i agree with sisou about the tests, if that's not too much effort)

sisou (approved):
Good to go!

PR #437, opened 2021-08-11
Password input alignment on mobile (except iOS)

mraveux:
This PR add a small change to prevent the password input, or the submit button present below it, from being (partially) hidden behind the on-screen keyboard when typing password.
What it does is simply scroll back the PasswordBox (or any PasswordInput parent) into view on focus of input.

**This does not work on iOS!**

### DIFF
--- a/src/components/PasswordInput.js
+++ b/src/components/PasswordInput.js
@@ -1,6 +1,7 @@
 /* global Nimiq */
 /* global I18n */
 /* global TemplateTags */
+/* global BrowserDetection */
 
 class PasswordInput extends Nimiq.Observable {
     /**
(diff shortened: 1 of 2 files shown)


src/components/PasswordInput.js
```diff
@@ -108,6 +117,17 @@ class PasswordInput extends Nimiq.Observable {
...
+
+    scrollParentIntoView() {
+        const $parent = this.$el.parentElement;
+
+        if (!$parent) return;
+
+        $parent.scrollIntoView({
+            behavior: 'smooth',
+            block: 'end',
+        });
```
sisou, line 129: According to https://caniuse.com/scrollintoview, the `smooth` option is not supported on all non-iOS browsers. E.g. the UC and Samsung browsers don't support an options object.

I don't know for sure, but I expect this to throw an error in these browsers, and I would suggest wrapping it into a `try...catch` block and call `scrollIntoView(false)` (false means [the bottom of the element will be aligned to the bottom of the visible area](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#parameters)) in the catch case.

Except if we don't want any scrolling when it cannot be smooth?

sisou (approved):

src/components/PasswordInput.js
mraveux: if i try to use the `smooth` option on a browser that don't support it, it will just ignore the `smooth` option and scroll instantly. The question now is: do we want it to scroll instantly when `smooth` is not supported, or not scroll at all in those cases and let the browser do its thing?

sisou: Ah, because an object is truthy! I'd say yes.

sisou: In that case this PR is good to go!

PR #454, opened 2022-10-24
Ten31 Pass support

danimoh:
Display a "Swap Authorized" indicator for KYC-enabled swaps.
Additionally, includes a bit of code cleanup and a bug fix in the PasswordBox component before adding this change there.

The current check for the KYC token is very primitive, but good enough for now, because not much harm can be done by providing a fake KYC token (only causes the Keyguard to display the "Swap authorized") and because the swap request can only be sent to the hub by trusted apps.
Eventually, I plan to check the KYC token signature and content.
For this, I want to use https://github.com/panva/jose/, strip it down to only the required parts, translate those to js and potentially put it into a sandbox as we're doing for the multisig rsa lib.
But all that doesn't have to happen before mainnet launch.

### DIFF
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -213,7 +213,7 @@ export type SettlementInstruction = MockSettlementInstruction | SepaSettlementIn
 
 export type SignSwapRequestLayout = 'standard' | 'slider';
 
-export type KycProvider = 'TEN31 PASS';
+export type KycProvider = 'TEN31 Pass';
 
 export type SignSwapRequestCommon = SimpleRequest & {
     swapId: string,
--- a/client/src/PublicRequest.ts
+++ b/client/src/PublicRequest.ts
@@ -213,6 +213,8 @@ export type SettlementInstruction = MockSettlementInstruction | SepaSettlementIn
 
 export type SignSwapRequestLayout = 'standard' | 'slider';
 
+export type KycProvider = 'TEN31 PASS';
+
 export type SignSwapRequestCommon = SimpleRequest & {
     swapId: string,
     fund: (
@@ -294,6 +296,15 @@ export type SignSwapRequestCommon = SimpleRequest & {
         processing: number,
     },
     serviceSwapFee: number, // Luna, Sats or Cents, depending which one gets funded
+
+    // Optional KYC info for swapping at higher limits.
+    // KYC-enabled swaps facilitated by S3/Fastspot require an s3GrantToken and swaps from or to Euro via OASIS
+    // additionally require a clearing or settlement specific oasisGrantToken.
+    kyc?: {
+        provider: KycProvider,
+        s3GrantToken: string,
+        oasisGrantToken?: string,
+    };
 };
 
 export type SignSwapRequestStandard = SignSwapRequestCommon & {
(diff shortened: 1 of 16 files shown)


client/src/PublicRequest.ts
```diff
@@ -292,6 +294,13 @@ export type SignSwapRequestCommon = SimpleRequest & {
         processing: number,
     },
     serviceSwapFee: number, // Luna, Sats or Cents, depending which one gets funded
+
+    // Optional KYC info for swapping at higher limits
+    kyc?: {
+        provider: KycProvider,
+        s3GrantToken: string,
+        oasisGrantToken?: string,
```
sisou, line 306: Can you add a comment to the OASIS grant token, for which swap pair and direction its required?

src/components/PasswordBox.css
```diff
@@ -43,13 +59,38 @@
...
+    opacity: 0;
     pointer-events: none;
 }
 
-/* S E T T E R */
+.password-box.show-swap-authorization {
+    overflow: hidden;
+}
+.password-box.show-swap-authorization::after {
+    content: attr(data-i18n-swap-authorization-caption);
```
sisou, line 76: Nice! 👍

src/request/sign-swap/SignSwapApi.js
```diff
@@ -189,6 +189,22 @@ class SignSwapApi extends BitcoinRequestParserMixin(TopLevelApi) {
...
         }
 
+        // Parse optional KYC data
+        if (request.kyc) {
+            if (request.kyc.provider !== 'TEN31 Pass') {
+                throw new Errors.InvalidRequestError(`Unsupported KYC provider: ${request.kyc.provider}`);
+            }
+            // TODO verify tokens
+            if (typeof request.kyc.s3GrantToken !== 'string' || !request.kyc.s3GrantToken) {
+                throw new Error('Invalid TEN31 Pass S3 service grant');
```
sisou:
```suggestion
                throw new Error('Invalid KYC S3 grant token');
```

src/request/sign-swap/SignSwapApi.js
```diff
@@ -189,6 +189,22 @@ class SignSwapApi extends BitcoinRequestParserMixin(TopLevelApi) {
...
+            }
+            if (request.kyc.oasisGrantToken !== undefined
+                && (typeof request.kyc.oasisGrantToken !== 'string' || !request.kyc.oasisGrantToken)) {
+                throw new Error('Invalid TEN31 Pass OASIS service grant');
```
sisou:
```suggestion
                throw new Error('Invalid KYC OASIS grant token');
```

sisou (approved):
Please regard my comments as suggestions only.

sisou:
Regarding grant JWT verification: I don't think it's possible to verify the signature of a JWT (if signed with a symmetric key, as most are), without the actual key (which only the KYC service has server-side). We can only decode the token contents and header (it's just base64).

Except if the JWT is using an asymmetric key, then we can verify against the public key I guess. However, whenever the KYC service rotates their keys (which it should do regularly), we would have to update the Keyguard, too. Maybe the service can make the JWT public key available on an API endpoint.

mraveux (approved):

danimoh:
@sisou the KYC server uses an assymetric key for signing the tokens, so verifying them in the Keyguard would certainly be possible. But for now this remains a low priority task. I left the TODO in the code to maybe pick it up in the future.

PR #455, opened 2022-11-21
QrScanner: disable native BarcodeDetector on M1 Macs with Ventura

danimoh:
On Macs with an M1 processor and macOS Ventura (macOS version 13), the BarcodeDetector is broken in Chromium based browsers as of version 107. For that constellation, the BarcodeDetector does not error but does not detect QR codes. Macs without an M1 or before Ventura and Chromium browsers before 107 are fine. See issue nimiq/qr-scanner#209 and https://bugs.chromium.org/p/chromium/issues/detail?id=1382442

The people in the qr-scanner issue claimed that Chromium 106 works fine and the Chromium bug report is also for 107, but it seems like this has not actually been tested on Ventura, so I'm still going to do that on Micha's Mac and adapt the check if required.
Also only after that check I'm going to commit the change also to the qr-scanner repo and release a new package, update the vue-components `QrScanner` and the use in the Wallet.

### DIFF
--- a/src/lib/QrScanner.js
+++ b/src/lib/QrScanner.js
@@ -259,28 +259,31 @@ class QrScanner {
      * @returns {Promise<Worker | BarcodeDetector>}
      */
     static async createQrEngine(workerPath = QrScanner.WORKER_PATH) {
-        const userAgentData = navigator.userAgentData;
-        const useNativeBarcodeDetector = 'BarcodeDetector' in window
+        const supportsBarcodeDetector = 'BarcodeDetector' in window
             && BarcodeDetector.getSupportedFormats
-            && (await BarcodeDetector.getSupportedFormats()).includes('qr_code')
-            // On Macs with an M1 processor and macOS Ventura (macOS version 13), the BarcodeDetector is broken in
-            // Chromium based browsers, regardless of the version. For that constellation, the BarcodeDetector does not
-            // error but does not detect QR codes. Macs without an M1 or before Ventura are fine.
-            // See issue qr-scanner/#209 and https://bugs.chromium.org/p/chromium/issues/detail?id=1382442
-            // TODO update this once the issue in macOS is fixed
-            && !(userAgentData // userAgentData is supported by all Chromium browsers
-                && /macOS/i.test(userAgentData.platform)
-                && userAgentData.brands.some(({ brand }) => /Chromium/i.test(brand))
-                // Check this last as getHighEntropyValues can theoretically trigger a browser prompt.
-                // If browser or user refused to return the requested values, disable the BarcodeDetector, to be safe.
-                && await userAgentData.getHighEntropyValues(['architecture', 'platformVersion'])
-                    .then(({ architecture, platformVersion }) => /arm/i.test(architecture || 'arm')
-                        && Number.parseInt(platformVersion || '13', 10) >= 13)
-                    .catch(() => true)
-            );
-        return useNativeBarcodeDetector
-            ? new BarcodeDetector({ formats: ['qr_code'] })
-            : new Worker(workerPath);
+            && (await BarcodeDetector.getSupportedFormats()).includes('qr_code');
+
+        if (!supportsBarcodeDetector) return new Worker(workerPath);
+
+        // On Macs with an M1/M2 processor and macOS Ventura (macOS version 13), the BarcodeDetector is broken in
+        // Chromium based browsers, regardless of the version. For that constellation, the BarcodeDetector does not
+        // error but does not detect QR codes. Macs without an M1/M2 or before Ventura are fine.
+        // See issue qr-scanner/#209 and https://bugs.chromium.org/p/chromium/issues/detail?id=1382442
+        // TODO update this once the issue in macOS is fixed
+        const userAgentData = navigator.userAgentData;
+        const isChromiumOnMacWithArmVentura = userAgentData // all Chromium browsers support userAgentData
+            && userAgentData.brands.some(({ brand }) => /Chromium/i.test(brand))
+            && /mac ?OS/i.test(userAgentData.platform)
+            // Does it have an ARM chip (e.g. M1/M2) and Ventura? Check this last as getHighEntropyValues can
+            // theoretically trigger a browser prompt, although no browser currently does seem to show one.
+            // If browser or user refused to return the requested values, assume broken ARM Ventura, to be safe.
+            && await userAgentData.getHighEntropyValues(['architecture', 'platformVersion'])
+                .then(({ architecture, platformVersion }) => /arm/i.test(architecture || 'arm')
+                    && Number.parseInt(platformVersion || '13', 10) >= /* Ventura */ 13)
+                .catch(() => true);
+        if (isChromiumOnMacWithArmVentura) return new Worker(workerPath);
+
+        return new BarcodeDetector({ formats: ['qr_code'] });
     }
 
     _onPlay() {
--- a/src/lib/QrScanner.js
+++ b/src/lib/QrScanner.js
@@ -259,11 +259,28 @@ class QrScanner {
      * @returns {Promise<Worker | BarcodeDetector>}
      */
     static async createQrEngine(workerPath = QrScanner.WORKER_PATH) {
-        if ('BarcodeDetector' in window && (await BarcodeDetector.getSupportedFormats()).includes('qr_code')) {
-            return new BarcodeDetector({ formats: ['qr_code'] });
-        }
-
-        return new Worker(workerPath);
+        const userAgentData = navigator.userAgentData;
+        const useNativeBarcodeDetector = 'BarcodeDetector' in window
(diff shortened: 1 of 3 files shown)


src/lib/QrScanner.js
```diff
@@ -259,11 +259,29 @@ class QrScanner {
...
+                // Check this last as getHighEntropyValues can theoretically trigger a browser prompt.
+                // If browser or user refused to return the requested values, disable the BarcodeDetector, to be safe.
+                && await userAgentData.getHighEntropyValues(['architecture', 'platformVersion'])
+                    .then(({ architecture, platformVersion }) => /arm/i.test(architecture || 'arm')
+                        && Number.parseInt(platformVersion || '13', 10) >= 13)
+                    .catch(() => true)
+            );
+        return useNativeBarcodeDetector
+            ? new BarcodeDetector({ formats: ['qr_code'] })
+            : new Worker(workerPath);
```
sisou: I suggest to rewrite this condition into smaller variables, to make the code easier to understand and avoid inverted/negated condition brackets.

```suggestion
        const supportsBarcodeDetector = 'BarcodeDetector' in window
            && typeof BarcodeDetector.getSupportedFormats === 'function'
            && (await BarcodeDetector.getSupportedFormats()).includes('qr_code');

        // On Macs with an M1 processor and macOS Ventura (macOS version 13), the BarcodeDetector is broken in
        // Chromium based browsers as of version 107. For that constellation, the BarcodeDetector does not error but
        // does not detect QR codes. Macs without an M1 or before Ventura and Chromium browsers before 107 are fine.
        // See issue qr-scanner/#209 and https://bugs.chromium.org/p/chromium/issues/detail?id=1382442
        // TODO update this once the issue in Chromium / macOS is fixed
        const userAgentData = navigator.userAgentData; // userAgentData is supported by all Chromium browsers
        const isMacOs = userAgentData && /macOS/i.test(userAgentData.platform);
        const isChromium107 = userAgentData && userAgentData.brands.some(
            ({ brand, version }) => /Chromium/i.test(brand) && Number.parseInt(version, 10) >= 107,
        );
        /** @returns {Promise<boolean>} */
        const isArmMac = async () => {
            // Check this last as getHighEntropyValues can theoretically trigger a browser prompt.
            // If browser or user refused to return the requested values, disable the BarcodeDetector, to be safe.
            try {
                if (!userAgentData) throw new Error('Can only be used when userAgentData is present');
                const { architecture, platformVersion } = await userAgentData.getHighEntropyValues(
                    ['architecture', 'platformVersion'],
                );
                return /arm/i.test(architecture || 'arm') && Number.parseInt(platformVersion || '13', 10) >= 13;
            } catch (error) {
                return true;
            }
        };
        const isBarcodeDetectorBroken = isMacOs && isChromium107 && await isArmMac();

        const useNativeBarcodeDetector = supportsBarcodeDetector && !isBarcodeDetectorBroken;
        return useNativeBarcodeDetector
            ? new BarcodeDetector({ formats: ['qr_code'] })
            : new Worker(workerPath);
```

sisou (changes requested):
I suggest to rewrite the condition, as per reasons outlined in the comment.

It would even be better if we wouldn't have to use user-agent detection, but could instead check if the BarcodeDetector works, e.g. with a short function that tries to scan a base64-encoded image that is known to work? That way we don't depend on browser versions and would never have to update this code again. Capability detection.

Something like this:
```js
/** @returns {Promise<boolean>} */
const useBarcodeDetector = async () => {
    if (!('BarcodeDetector' in window)) return false;
    if (/* check for qr_code support */) return false;

    const image = 'data:image/jpg;base64,<data>';
    const detector = new BarcodeDetector({ formats: ['qr_code'] });
    const timeout = new Promise((_, reject) => window.setTimeout(reject, 1000));

    try {
        await Promise.race([
            timeout,
            detector.detect(image),
        ]);
        return true;
    } catch (e) {
        return false;
    }
}
```

danimoh:
I discussed your proposal to detect the bug by trying to scan a valid qr code image with Matheo.
While the approach is nice indeed, this will cause a delay in initializing the scanner for all users.
For failing devices, it would delay the loading of the fallback worker and for non-failing devices, it would block the BarcodeDetector for one frame, even using the improved detection outlined below and assuming the BarcodeDetector does not process multiple frames in parallel.

The delays we measured were as follows:
On a decent Intel Mac: .2s-.5s, mostly ~.3s
On a decent Android phone: .02s-.04s, mostly .03s
On an old Android phone: 0.055s - 0.13s, mostly ~.08s (Yes that's right: Google's implementation on an old Android is about 4 times as fast as Apple's implementation on a powerful Mac)

Thus, delays on Macs are significant, delays on Android not so much.
The device detection on the other hand is effectively instant.

Matheo and I were a bit undecided which approach would be better, and definitely both have advantages and disadvantages.
But in the end Matheo prefers the device detection over the bug detection, and I'm also leaning towards that side.

I believe that we have a good understanding of which browsers are failing and we can detect them with high precision.
Also, some day, this issue will be fixed in macOS and the check will then only be required for legacy browsers anymore, for which I don't want to have the delay of the bug detection and the added image dataURI for all eternity.
On the other hand, the device detection has the hypothetical drawback that the spec for `getHighEntropyValues()` allows for a browser prompt, but none of the tested browsers is doing that currently.
Also, in all tests on all tested browsers and devices, accurate values were returned without any obfuscation.
Should this change in the future, then the check can be adapted again at that point.

For future reference, this would be a non-flashed-out and non-tested draft of an improved version of your bug detection suggestion with the following improvements:
- it re-uses the created `BarcodeDetector` instance that was used for testing also for the regular scanning. (The first scanned frame takes longer, but don't know whether per `BarcodeDetector` instance or just once overall, but re-using the instance is good anyways.)
- it checks the scan result instead of using a timeout because the `BarcodeDetector` always return a result, it's just empty.
- the image has to be loaded; `BarcodeDetector` can't handle dataURIs.
- makes the `BarcodeDetector` immediately available and just disables it in case that it's failing.

```ts
    /**
     * @param {string} [workerPath]
     * @returns {Promise<Worker | BarcodeDetector>}
     */
    static async createQrEngine(workerPath = QrScanner.WORKER_PATH) {
        const useNativeBarcodeDetector = !QrScanner._disableBarcodeDetector
            && 'BarcodeDetector' in window
            && BarcodeDetector.getSupportedFormats
            && (await BarcodeDetector.getSupportedFormats()).includes('qr_code');
        if (useNativeBarcodeDetector) {
            const detector = new BarcodeDetector({ formats: ['qr_code'] });
            // Check in the background whether the detector is working
            (async () => {
                const image = await QrScanner._loadImage('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyAQMAAAAk8RryAAAABlBMVEX///8AAABVwtN+AAAACXBIWXMAAA7EAAAOxAGVKw4bAAAAB3RJTUUH5gsVDysxMCx+7QAAAJxJREFUGNNl0LEKwyAQBuATVx8gU5/Bkr33Km55hQ6BCh2yBJwDeSQfIqEvUMgUEK6iBrXe8sGB998J8FdI2r0oK5mxirfeT+2GT6kRte9KyWYZ30eRZhtzogA8bVCLp+uXNftQI+06i8vXPqFw4xj6SRjIdmMhCNupViRzTC4rmYabyvq7jmmrDPMLQ/6l/yfc11Y/X6Q8EfOr+gEdfnwjoKSOwQAAAABJRU5ErkJggg==');
                try {
                    if (!(await detector.detect(image)).length) throw new Error('Failed to detect expected qr code');
                } catch (e) {
                    // Also _scanFrame will need to react to this change; see @nimiq/qr-scanner
                    QrScanner._disableBarcodeDetector = true;
                }
            )();
            return detector;
        }
        return new Worker(workerPath);
    }
```

Regarding the check on Chromium 106 I did: it's also failing on M1 Ventura, as I expected, i.e. the comments in nimiq/qr-scanner were misleading in that regard and the issue was not introduced in a change in Chromium, but by Apple in macOS.

Regarding the refactoring of the condition with named variables: makes the code longer, but sure, can do. But I'll wait with changes until Matheo also comments on this.

sisou:
OK, sounds good. Do you know what's up with the CI test? It's failing to load Firefox for running the tests.

danimoh:
I was wondering too, but unfortunately, I don't know, why that is.
Maybe the newest `ubuntu-latest` doesn't include Firefox anymore and we have to install it manually, for example with https://github.com/browser-actions/setup-firefox.

danimoh:
I opened #456 to fix the Github Actions issue.

mraveux (approved):

danimoh:
@sisou I refactored the check for better readability and added more comments, but didn't go quite as far with putting everything into separate variables as your proposal, as most of the checks are self-explanatory and I didn't like defining `isArmMac` as a function.

Also I force-pushed the original commit to remove the version check for Chromium 107.

Please re-review.

sisou (approved):
Yup, this is good now!

PR #458, opened 2022-12-01
Multisig: Connect request and SignMultisigTransaction request

sisou:
Related: https://github.com/nimiq/hub/pull/521

This PR adds two requests:
- Connect
- SignMultisigTransaction

## Connect

For this request, I am adding deterministic RSA key generation from the user's Key's `secret`. The public RSA key is then shared with the connected app, to be able to encrypt data that only the Keyguard can decrypt again. Encryption and decryption uses the `window.crypto.subtle` APIs.

## SignMultisigRequest

Nothing special added, this uses the code from the @nimiq/core SDK to do multisig signatures.


sisou:
@danimoh Do you maybe have a better idea of how to solve the double `KeyguardCommand.KeyguardCommand` in this commit?
https://github.com/nimiq/keyguard/pull/458/commits/253219009034c528395025c8d42168deade52d9c

.eslintignore
```diff
@@ -1,3 +1,4 @@
 *.min.*
 src/lib/bitcoin/BitcoinJS.js
 src/lib/polygon/OpenGSN.js
+src/lib/multisig/wasm/pkg
```
danimoh: This folder has been removed again in fbbcf067a5043d9d56bd24c5a2ac79f63355fac1.

client/src/PublicRequest.ts
```diff
@@ -189,6 +190,40 @@ export type SignTransactionRequest
...
+    },
+    userName?: string,
+};
+
+export type SignMultisigTransactionRequestCommon = Transform<SignTransactionRequestCommon, 'keyLabel' | 'senderLabel', {
+    keyLabel: string, // Not optional
+    senderLabel: string, // Not optional
+}> & {
+    multisigConfig: MultisigConfig,
+};
```
danimoh: Could also be
```suggestion
export type SignMultisigTransactionRequestCommon = Transform<
    SignTransactionRequestCommon,
    'keyLabel' | 'senderLabel',
    Required<Pick<SignTransactionRequestCommon, 'keyLabel' | 'senderLabel'>>
> & {
    multisigConfig: MultisigConfig,
};
```
This ensures that there are no discrepancies if the original types of `keyLabel` or `senderLabel` were to change in `SignTransactionRequestCommon`, or if they were to be removed there. Obviously, in this case this benefit is of a rather theoretical nature. I just wanted to mention this alternative.

client/src/PublicRequest.ts
```diff
@@ -94,6 +94,7 @@ export type BitcoinTransactionInfo = {
 };
 
 export type SignTransactionRequestLayout = 'standard' | 'checkout' | 'cashlink';
+export type SignMultisigTransactionRequestLayout = 'standard';
```
danimoh, line 97: Are you planning to add other layouts?

client/src/PublicRequest.ts
```diff
@@ -593,8 +652,11 @@ export type Result = RedirectResult | IFrameResult;
 // Derived Result types
 
 export type ResultType<T extends RedirectRequest> =
-    T extends Is<T, SignMessageRequest> | Is<T, SignTransactionRequest> ? SignatureResult :
+    T extends Is<T, SignMessageRequest> ? SignatureResult :
```
danimoh, line 666: Should `SignMessageRequest` also receive an alias `SignMessageResult` for `SignatureResult`, similar to `SignMultisigTransactionResult`? Also in `RedirectResult` above and `ResultByCommand` below.

types/Keyguard.d.ts
```diff
@@ -32,12 +32,38 @@ type AccountRecord = AccountInfo & {
     encryptedKeyPair: Uint8Array
 }
 
+type EncryptionKeyParams = {
+    kdf: string
+    iterations: number
+    keySize: number
+}
```
danimoh: Can be used as `KeyguardRequest.EncryptionKeyParams` without copying it from `PublicRequest.ts`.

```suggestion
type EncryptionKeyParams = KeyguardRequest.EncryptionKeyParams;
```

package.json
```diff
@@ -38,13 +38,14 @@
   },
   "homepage": "https://github.com/nimiq/keyguard#readme",
   "devDependencies": {
-    "@nimiq/albatross-wasm": "npm:@nimiq/core@^2.0.6",
+    "@nimiq/albatross-wasm": "https://gitpkg.now.sh/nimiq/core-rs-albatross/web-client/dist?06999e1626b3312a906a0a490339d7feb781811b",
```
danimoh, line 41: Will the Multisig primitives eventually be added to the regular `@nimiq/core` package?

tsconfig.json
```diff
@@ -70,6 +70,7 @@
     "src/lib/QrScannerWorker.js",
     "src/lib/bitcoin/BitcoinJS.js",
     "src/lib/polygon/OpenGSN.js",
+    "src/lib/multisig/wasm/pkg",
```
danimoh: This folder has been removed again in fbbcf067a5043d9d56bd24c5a2ac79f63355fac1.

src/lib/rsa/sandboxed/forge.min.js.map
danimoh, line 1: Doesn't look like these mappings are very useful. (Even though they were taken directly from the `node-forge` package).

src/lib/rsa/sandboxed/forge.min.js
danimoh, line 1: For any other reviewer wondering: this is `forge.min.js` from `node-forge`  version 1.3.1 (sha256 dc67fd132427ad96c9666c844b39565413c40ddb1f2d063c53512fbf6d387dfd)

src/lib/rsa/sandboxed/RSAKeysIframe.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>RSA Keys Iframe | Keyguard Demo</title>
```
danimoh: It's not only used in the demos.
```suggestion
    <title>RSA Keys Iframe</title>
```

src/lib/rsa/sandboxed/RSAKeysIframe.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+const $loading = /** @type {HTMLSpanElement} */ (document.querySelector('#loading'));
+
+/**
+ * @param {any} key
+ * @returns {ArrayBuffer}
+ */
+function key2ArrayBuffer(key) {
+    const bytes = /** @type {string} */ (forge.asn1.toDer(key).getBytes());
```
danimoh: Can be typed more accurately, and I'd personally prefer `keyToArrayBuffer` as a name (do we do that anywhere else to appreviate To as 2?). Additionally, our APIs mostly work on Uint8Arrays, so I'd suggest `keyToUint8Array`:
```js
/**
 * @param {forge.asn1.Asn1} key
 * @returns {Uint8Array}
 */
function keyToUint8Array(key) {
    const bytes = forge.asn1.toDer(key).getBytes();
    const result = new Uint8Array(bytes.length);
    for (let i = 0; i < bytes.length; i++) {
        result[i] = bytes.charCodeAt(i);
    }
    return result;
}
```

src/lib/rsa/sandboxed/RSAKeysIframe.html
```diff
@@ -0,0 +1,13 @@
...
+    <script src="./forge.min.js"></script>
+</head>
```
danimoh, line 16: We should add a CSP meta tag here which enforces the sandbox, even when not opened in the iframe it was designed for, and set the `frame-ancestors` to only the keyguard to disallow the page being embedded in other pages.

Generally, the CSP for this iframe can be very strict, for example:
```html
    <meta http-equiv="Content-Security-Policy" content="
        sandbox allow-scripts;
        frame-ancestors 'self';
        default-src 'none';
        script-src 'self';
    ">
```

However, unfortunately, `sandbox` and `frame-ancestors` do not seem to be allowed to be specified via the `meta` tag. So those should be configured on the server by @curdbecker.

src/lib/rsa/sandboxed/RSAKeysIframe.js
danimoh, line 1: The entire script should refuse running (e.g. by throwing immediately) if it's not inside an iframe or if it's not an iframe within the Keyguard. Otherwise, a malicious page can open the page as iframe (unless disallowed by an appropriate CSP) or as a popup (always possible, I think), and communicate with it.

```js
if (// Refuse to run if:
    // - Run as top level window
    window.self === window.top
    // - Iframe is embedded in another domain
    || window.location.origin !== new URL(document.referrer).origin
    // - Iframe can access parent page and is thus not sandboxed
    || !!window.frameElement
) throw new Error('Only allowed to run in a sandboxed iframe');
```

src/lib/rsa/sandboxed/RSAKeysIframe.js
```diff
@@ -0,0 +1,85 @@
...
+        view[i] = bytes.charCodeAt(i);
+    }
+    return buffer;
+}
+
+window.addEventListener('message', async event => {
+    if (event.source === event.target) {
+        // console.log("Ignored same-window event:", event);
+        return;
+    }
```
danimoh: Where would such same-window events come from? Does `forge` send messages to the window itself?
(Mostly out of curiosity. Checking for it does no harm of course.)

danimoh: Messages which are not coming from the Keyguard should be rejected.
```suggestion
    if (event.origin !== window.location.origin || !event.source) return; // Reject messages from other origins
```

src/lib/rsa/sandboxed/RSAKeysIframe.js
```diff
@@ -0,0 +1,85 @@
+/* global forge */
+
+/**
+ * @typedef {{command: 'generateKey', seed: string, keySize: number}} GenerateKeyCommand
```
danimoh: The `seed` should be better a `Uint8Array` instead of a binary string because it's not very clear that a binary string is expected, and not for example hex encoded data, and binary strings just feel very outdated.
```suggestion
 * @typedef {{command: 'generateKey', seed: Uint8Array, keySize: number}} GenerateKeyCommand
```

src/lib/rsa/sandboxed/RSAKeysIframe.js
```diff
@@ -0,0 +1,85 @@
...
+    await new Promise(res => window.setTimeout(res, 50)); // Give DOM time to update
+
+    // PRNG-seeded RSA key generation
+    const prng = forge.random.createInstance();
+    prng.seedFileSync = needed => {
+        // console.log('Need', needed, 'bytes of randomness, have', data.seed.length);
+        let seed = '';
+        while (seed.length < needed) {
+            seed += data.seed;
+        }
```
danimoh: If not enough entropy was provided, we should throw and reject instead. The caller should make sure to provide enough entropy or stretch it out with a proper key derivation function, instead of us simply repeating the seed here.
```suggestion
        if (needed > seed.length) {
            throw new Error(`Seed does not provide enough entropy. ${needed} bytes needed`);
        }
```

src/lib/rsa/sandboxed/RSAKeysIframe.js
```diff
@@ -0,0 +1,85 @@
...
+        return seed.substring(0, needed);
```
danimoh: There's no good reason to truncate the entropy if we have more than required, as whatever amount of entropy received [will be collected](https://github.com/digitalbazaar/forge/blob/a0a4a4264bedb3296974b9675349c9c190144aeb/lib/prng.js#L210).

Additionally, instead of truncating always to the first bytes of the entropy or returning the entire entropy, I was also thinking about returning a new slice of the entropy each time, because `forge.prng` does in fact [reseed for each generated random data](https://github.com/digitalbazaar/forge/blob/a0a4a4264bedb3296974b9675349c9c190144aeb/lib/prng.js#L90-L95). A quick experiment showed that a `forge.rsa.generateKeyPair` does in fact trigger two reseeds. Therefore one might consider returning two individual slices of `needed` bytes each, and overall requiring a seed of length `2 * needed`. However, looking at the linked comment, it seems like the frequent reseeding is mostly out of paranoia and not strictly required.

Overall, I would say returning the full, non-truncated seed in a single slice on all reseeds is a good middle ground.

src/lib/rsa/sandboxed/RSAKeysIframe.js
```diff
@@ -0,0 +1,85 @@
...
+    //     { bits: data.keySize, prng, workers: -1, workerScript: './node-forge/prime.worker.min.js' },
+    //     (error, keyPair) => {
+    //         if (error) {
+    //             reject(error);
+    //             return;
+    //         }
+
+    //         resolve(keyPair);
+    //     },
+    // ));
```
danimoh: > Sadly, those cannot be created in sandboxed iframes without `allow-same-origin`

Or luckily, because using webworkers could apparently cause the key generation to not be deterministic anymore: if a prng is passed, `generateKeyPair` uses a pure js implementation, which if async mode is enabled by passing a callback,  calls `_generateKeyPair`. `_generateKeyPair` in turn calls `prime.generateProbablePrime` which then calls `primeincFindPrime` and that calls `primeincFindPrimeWithWorkers` which features [the following comment](https://github.com/digitalbazaar/forge/blob/a0a4a4264bedb3296974b9675349c9c190144aeb/lib/prime.js#L150-L153):
```
// NOTE: This algorithm is indeterminate in nature because workers
// run in parallel looking at different segments of numbers. Even if this
// algorithm is run twice with the same input from a predictable RNG, it
// may produce different outputs.
```

Therefore, the async usage should be banned here.
```suggestion
    // Use the synchronous version of forge.rsa.generateKeyPair and not the asynchronous version with WebWorkers because
    // - WebWorkers cannot be created in sandboxed iframes without `allow-same-origin`:
    //   https://stackoverflow.com/a/30567964/4204380.
    // - Using multiple WebWorkers introduces race conditions, which causes the key generation to not be deterministic
    //   anymore, even if a deterministically seeded PRNG is used, see documentation on `primeincFindPrimeWithWorkers`:
    //   https://github.com/digitalbazaar/forge/blob/a0a4a4264bedb3296974b9675349c9c190144aeb/lib/prime.js#L150-L153
    //   If using WebWorkers, that method is called via `generateKeyPair` > `_generateKeyPair` > `generateProbablePrime`
    //   > `primeincFindPrime` > `primeincFindPrimeWithWorkers`.
    const keyPair = forge.pki.rsa.generateKeyPair({ bits: data.keySize, prng });
```

Unfortunately, this blocks the main thread in some browsers like Firefox, where same origin iframes are running on the same thread as the main window. In Chrome, [that is not the case](https://www.chromium.org/developers/design-documents/oop-iframes).

src/lib/rsa/sandboxed/RSAKeysIframe.js
```diff
@@ -0,0 +1,85 @@
...
+    // Wrap into PKCS#8 format (to be understandable by `window.crypto.subtle.importKey`)
+    const privateKeyInfo = forge.pki.wrapRsaPrivateKey(rsaPrivateKey);
+
+    // Export to ASN1 notation (already in SPKI format)
+    const publicKeyInfo = forge.pki.publicKeyToAsn1(keyPair.publicKey);
+
+    /** @type {Window} */ (event.source).postMessage({
+        privateKey: key2ArrayBuffer(privateKeyInfo),
+        publicKey: key2ArrayBuffer(publicKeyInfo),
+    }, '*');
```
danimoh:
```suggestion
    /** @type {Window} */ (event.source).postMessage({
        privateKey: key2ArrayBuffer(privateKeyInfo),
        publicKey: key2ArrayBuffer(publicKeyInfo),
    }, window.location.origin);
```
Either only allow sending the response to known keyguard origins, or the same origin the iframe is running on (also the default if no target origin is specified but the sandbox is messing with that so our `location.origin` has to be specified manually).

src/lib/Key.js
```diff
@@ -11,16 +17,49 @@ class Key {
...
+            keySize: CONFIG.RSA_KEY_BITS,
+        };
+    }
+
+    /**
+     * @param {EncryptionKeyParams} paramsA
+     * @param {EncryptionKeyParams} paramsB
+     * @returns {boolean}
+     */
+    static _isEncryptionKeyParamsEqual(paramsA, paramsB) {
```
danimoh:
```suggestion
    /**
     * @private
     * @param {EncryptionKeyParams} paramsA
     * @param {EncryptionKeyParams} paramsB
     * @returns {boolean}
     */
    static _areEncryptionKeyParamsEqual(paramsA, paramsB) {
```

src/lib/Key.js
```diff
@@ -11,16 +17,49 @@ class Key {
...
-    constructor(secret, hasPin = false) {
+    constructor(secret, config = {}) {
         this._secret = secret;
-        this._hasPin = hasPin;
+        this._hasPin = Boolean(config.hasPin);
         /** @type {string?} */
         this._id = null;
         this._defaultAddress = this.deriveAddress(Constants.DEFAULT_DERIVATION_PATH);
+
+        this.rsaKeyPair = config.rsaKeyPair;
```
danimoh:
```suggestion
        this._rsaKeyPair = config.rsaKeyPair;
```
For consistency with other properties, the name should start with an underscore, and then a getter be added. Also, I would move the definition just below the definition of `_hasPin`, as it also belongs to the `config`.

Additonal getter:
```js
    /**
     * @returns {RsaKeyPairExport | undefined}
     */
    getRsaKeyPairExportIfExists() {
        return this._rsaKeyPair;
    }
```

src/lib/Key.js
```diff
@@ -50,6 +89,37 @@ class Key {
...
+        const privateKey = this.derivePrivateKey(path);
+        const publicKey = Nimiq.PublicKey.derive(privateKey);
+        const [otherPublicKeys, otherCommitments] = otherSigners.reduce((acc, signer) => {
+            acc[0].push(signer.publicKey);
+            acc[1].push(signer.commitments);
+            return acc;
+        }, [
+            /** @type {Nimiq.PublicKey[]} */ ([]),
+            /** @type {Nimiq.Commitment[][]} */ ([]),
+        ]);
```
danimoh:
```suggestion
        const otherPublicKeys = otherSigners.map(({ publicKey }) => publicKey);
        const otherCommitments = otherSigners.map(({ commitments }) => commitments);
```

src/lib/Key.js
```diff
@@ -11,16 +17,49 @@ class Key {
...
+            && paramsA.kdf === paramsB.kdf
+            && paramsA.iterations === paramsB.iterations
+            && paramsA.keySize === paramsB.keySize;
+    }
+
+    /**
+     * @param {EncryptionKeyParams} params
+     * @returns {boolean}
+     */
+    static _isDefaultEncryptionKeyParams(params) {
```
danimoh:
```suggestion
    /**
     * @private
     * @param {EncryptionKeyParams} params
     * @returns {boolean}
     */
    static _areDefaultEncryptionKeyParams(params) {
```

src/lib/Key.js
```diff
@@ -91,6 +161,124 @@ class Key {
...
+     * @param {EncryptionKeyParams} keyParams
+     * @returns {Promise<CryptoKey>}
+     */
+    async getRsaPublicKey(keyParams) {
+        if (!this.rsaKeyPair || !Key._isEncryptionKeyParamsEqual(keyParams, this.rsaKeyPair.keyParams)) {
+            this.rsaKeyPair = await this._computeRsaKeyPair(keyParams);
+            if (Key._isDefaultEncryptionKeyParams(keyParams)) {
+                await KeyStore.instance.setRsaKeypair(this.id, this.rsaKeyPair);
+            }
+        }
```
danimoh: This little snippet could be deduplicated into a method `setRsaKeyPair`.
```suggestion
        const rsaKeyPair = await this.setRsaKeyPair(keyParams);
```

With method:
```js
    /**
     * Calculates and sets the RSA key pair for the given key params. The RSA key is calculated deterministically based
     * on _secret, such that the same RSA key is generated for the same key params.
     * Warning: this method is computationally expensive.
     * @param {EncryptionKeyParams} keyParams
     * @returns {Promise<RsaKeyPairExport>}
     */
    async setRsaKeyPair(keyParams) {
        if (!this._rsaKeyPair || !Key._areEncryptionKeyParamsEqual(keyParams, this._rsaKeyPair.keyParams)) {
            this._rsaKeyPair = await this._computeRsaKeyPair(keyParams);
            if (Key._areDefaultEncryptionKeyParams(keyParams)) {
                await KeyStore.instance.setRsaKeypair(this.id, this._rsaKeyPair);
            }
        }
        return this._rsaKeyPair;
    }
```

src/lib/Key.js
```diff
@@ -91,6 +161,124 @@ class Key {
...
+            ['encrypt'],
+        );
+    }
+
+    /**
+     * @param {EncryptionKeyParams} keyParams
+     * @returns {Promise<RsaKeyPairExport>}
+     */
+    async _computeRsaKeyPair(keyParams) {
+        const iframe = document.createElement('iframe');
```
danimoh, line 277: Would be good to add a comment that explains why we're doing all this fun with the iframe, i.e. browser APIs don't support the deterministic generation of RSA keys from a seed, therefore we use a library, but to avoid an increased attack surface by including a library not audited by us, we run it in a sandbox.
```suggestion
        // Setup the RSAKeysIframe in which the actual computation of the RSA key happens via node-forge. The reason why
        // we don't use the browser crypto API instead, is that it does not support deterministic generation of RSA keys
        // from an existing secret, and the reason why node-forge is not used directly here is that we want to run it in
        // a sandbox, to reduce the attack surface from a big external dependency, not thoroughly audited by us.
        const iframe = document.createElement('iframe');
```

src/lib/Key.js
```diff
@@ -91,6 +161,124 @@ class Key {
...
+        iframe.classList.add('rsa-sandboxed-iframe'); // Styles in common.css hide this class
+        iframe.setAttribute('sandbox', 'allow-scripts');
+        iframe.src = '../../lib/rsa/sandboxed/RSAKeysIframe.html'; // Relative path from a request URL
+        /** @type {Promise<void>} */
+        const loadPromise = new Promise(resolve => iframe.addEventListener('load', () => resolve()));
```
danimoh:
```suggestion
        const loadPromise = new Promise(resolve => iframe.addEventListener('load', resolve));
```

src/lib/Key.js
```diff
@@ -91,6 +161,124 @@ class Key {
...
+                );
+                break;
+            default:
+                throw new Error(`Unsupported KDF function: ${keyParams.kdf}`);
+        }
+
+        // Send computation command to iframe
+        iframe.contentWindow.postMessage({
+            command: 'generateKey',
+            seed: Nimiq.BufferUtils.toUtf8(seed), // seed is a bytestring
```
danimoh: This is a very bad idea, and a security issue.

This interprets the bytes as UTF8 encoding and decodes it to a Javascript string, which is UTF16 encoded. Up to 4 UTF8 bytes can represent a single unicode symbol, which is then encoded as a single UTF16 character or a UTF16 surrogate pair. The following re-encodings are possible:
1 UTF8 byte -> 1 UTF16 character
2 UTF8 bytes -> 1 UTF16 character
3 UTF8 bytes -> 1 UTF16 character
4 UTF8 bytes -> 2 UTF16 characters (surrogate pair)

This means that the resulting string can be as short as a third of the length of the input byte array. As `node-forge` expects a binary string as input, where each character is interpreted as a single byte, overall in the worst case, two thirds of the original data are discarded and an exponentially weaker entropy is used.

Note that fixing this is a breaking change, as the resulting RSA key will be different. Therefore, users who already have a stored RSA key need to logout to clear their old RSA key (or we rename the property in the store). The multisig app must then discard its known RSA public keys and associated encrypted secrets, and connect the account again to get the new RSA public key.

src/lib/Key.js
```diff
@@ -91,6 +161,124 @@ class Key {
...
+                privateKey: new Uint8Array(data.privateKey),
+                publicKey: new Uint8Array(data.publicKey),
+                keyParams,
+            });
+        }
+
+        // Listen for result from iframe
+        window.addEventListener('message', onMessage);
+
+        return resultPromise;
```
danimoh: I think the Promise reads nicer like this:
```suggestion
        return new Promise(resolve => {
            /**
             * @param {MessageEvent} event
             */
            function onMessage(event) {
                if (event.source === event.target) {
                    // console.log("Ignored same-window event:", event);
                    return;
                }

                /** @type {{privateKey: ArrayBuffer, publicKey: ArrayBuffer}} */
                const data = event.data;
                if (!('privateKey' in data) || !('publicKey' in data)) return;

                window.removeEventListener('message', onMessage);

                resolve({
                    privateKey: new Uint8Array(data.privateKey),
                    publicKey: new Uint8Array(data.publicKey),
                    keyParams,
                });
            }

            // Listen for result from iframe
            window.addEventListener('message', onMessage);
        });
```

src/lib/Key.js
```diff
@@ -91,6 +161,124 @@ class Key {
...
+            default:
+                throw new Error(`Unsupported KDF function: ${keyParams.kdf}`);
+        }
+
+        // Send computation command to iframe
+        iframe.contentWindow.postMessage({
+            command: 'generateKey',
+            seed: Nimiq.BufferUtils.toUtf8(seed), // seed is a bytestring
+            keySize: keyParams.keySize,
+        }, '*');
```
danimoh: Might want to add a comment that due to the iframe's sandboxed origin, we can't be more restrictive than `'*'`.

src/lib/Key.js
```diff
@@ -91,6 +161,124 @@ class Key {
...
+        });
+
+        /**
+         * @param {MessageEvent} event
+         */
+        function onMessage(event) {
+            if (event.source === event.target) {
+                // console.log("Ignored same-window event:", event);
+                return;
+            }
```
danimoh: Here, we absolutely must check that the message came from the iframe, otherwise the following attack is possible:
A malicious site starts a `Connect` request in an iframe (via Hub, which then redirects to the Keyguard). Then, when the malicious site suspects that the the Keyguard is waiting for the RSA key calculation, it itself sends a postmessage to the popup with an RSA key known to the attacker, which the Keyguard would willingly store and continue to use, also for other websites.

src/lib/Key.js
```diff
@@ -91,6 +161,124 @@ class Key {
...
+
+            /** @type {{privateKey: ArrayBuffer, publicKey: ArrayBuffer}} */
+            const data = event.data;
+            if (!('privateKey' in data) || !('publicKey' in data)) return;
+
+            window.removeEventListener('message', onMessage);
```
danimoh: At this point, also the iframe should be removed from the DOM, such that it can be garbage collected.

src/lib/KeyStore.js
```diff
@@ -147,12 +153,14 @@ class KeyStore {
...
 
-        const keyRecord = /** @type {KeyRecord} */ {
+        /** @type {KeyRecord} */
+        const keyRecord = {
             id: key.id,
             type: key.type,
             hasPin: key.hasPin,
             secret: buffer.subarray(0, buffer.byteLength),
             defaultAddress: key.defaultAddress.serialize(),
+            rsaKeyPair: key.rsaKeyPair,
```
danimoh: I think if the key is password encrypted, we should encrypt the rsa key as well. Same in setRsaKeypair.

src/lib/RequestParser.js
```diff
@@ -128,15 +128,16 @@ class RequestParser { // eslint-disable-line no-unused-vars
      * @returns {Nimiq.Transaction}
      */
     parseTransaction(object) {
+        if (!object || typeof object !== 'object' || object === null) {
+            throw new Errors.InvalidRequestError('Request must be an object');
```
danimoh:
```suggestion
        if (!object || typeof object !== 'object') {
            throw new Errors.InvalidRequestError('Transaction info must be an object');
```

- The `object === null` check is already covered by `!object`.
- The transaction is being parsed here, not the entire request.

tests/lib/RequestParser.spec.js
```diff
@@ -189,12 +189,12 @@ describe('RequestParser', () => {
...
-        expect(() => requestParser.parseShopLogoUrl({})).toThrow();
-        expect(() => requestParser.parseShopLogoUrl('isThisAUrl?')).toThrow();
-        expect(() => requestParser.parseShopLogoUrl('file:///home/maxmustermann/image.jpg')).toThrow();
+        expect(() => requestParser.parseLogoUrl(5, true, 'shopLogoUrl')).toThrow();
+        expect(() => requestParser.parseLogoUrl({}, true, 'shopLogoUrl')).toThrow();
+        expect(() => requestParser.parseLogoUrl('isThisAUrl?', true, 'shopLogoUrl')).toThrow();
+        expect(() => requestParser.parseLogoUrl('file:///home/maxmustermann/image.jpg', true, 'shopLogoUrl')).toThrow();
 
-        expect(requestParser.parseShopLogoUrl(undefined)).toBe(undefined);
+        expect(requestParser.parseLogoUrl(undefined, true, 'shopLogoUrl')).toBe(undefined);
```
danimoh, line 197: A test could be added here to test the `false` for the newly introduced `allowEmpty`.

tests/lib/Key.spec.js
```diff
@@ -56,4 +56,58 @@ describe('Key', () => {
...
+    //     const partialSignatureB = keyB.signPartially(
+    //         'm',
+    //         transaction.serializeContent(),
+    //         signerPublicKeys,
+    //         secretB,
+    //         aggregatedCommitment,
+    //     );
+
+    //     expect(partialSignatureB.toHex()).toEqual('caa6353261d250e2f1f67499f526c47503015e08d2a69169322fecae83cdf607');
+    // })
```
danimoh: TODO

src/request/connect/index.html
```diff
@@ -0,0 +1,133 @@
...
+        <div id="connect-account" class="page nq-card">
+            <div class="page-header nq-card-header">
+                <h1 class="nq-h1">
+                    <span data-i18n="connect-heading-prefix">Log in to</span>
+                    <span class="app-name"></span>
+                </h1>
+                <span class="more">
+                    Connect and use your account to<br>
+                    approve transactions.
+                </span>
```
danimoh: Missing i18n.

src/request/connect/index.html
```diff
@@ -0,0 +1,133 @@
...
+                <div class="app-icon-account-icon">
+                    <img class="app-icon">
+                    <div class="login-file-account-icon"></div>
+                </div>
+
+                <div class="flex-grow"></div>
+
+                <div class="bottom-explainer">
+                    The connection is for approving only.<br>
+                    Your funds remain unaffected.
```
danimoh: Missing i18n.

src/request/connect/index.html
```diff
@@ -0,0 +1,133 @@
...
+                                The first address of your account will be visible to all participants in your shared wallet.
+                            </p>
+                        </div>
+                    </div>
+                </div>
+            </div>
+
+            <div class="page-footer nq-card-footer">
+                <button data-i18n="passwordbox-connect-account" class="nq-button light-blue continue display-none">Connect your account</button>
+                <form id="password-box" class=""></form>
```
danimoh: Empty `class` can be removed.

```suggestion
                <form id="password-box"></form>
```

src/request/connect/ConnectApi.js
```diff
@@ -0,0 +1,75 @@
...
+    // /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-transaction'), // Already third-party whitelisted in Hub
+    /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-multisig-transaction'),
+    // /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-message'), // Already third-party whitelisted in Hub
+    /** @type {KeyguardCommand.KeyguardCommand} */ ('derive-address'),
+    // Bitcoin
+    /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-btc-transaction'),
+    /** @type {KeyguardCommand.KeyguardCommand} */ ('derive-btc-xpub'),
+    // Swap
+    /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-swap'),
+];
```
danimoh: Should some of the new commands like `sign-staking` or `sign-polygon-transaction` be added?

```suggestion
const PermissionKeyguardCommands = [
    /** @type {KeyguardCommand.KeyguardCommand} */ ('remove-key'),
    /** @type {KeyguardCommand.KeyguardCommand} */ ('export'),
    /** @type {KeyguardCommand.KeyguardCommand} */ ('change-password'),
    // /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-transaction'), // Already third-party whitelisted in Hub
    /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-multisig-transaction'),
    // /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-staking'), // Already third-party whitelisted in Hub
    // /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-message'), // Already third-party whitelisted in Hub
    /** @type {KeyguardCommand.KeyguardCommand} */ ('derive-address'),
    // Bitcoin
    /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-btc-transaction'),
    /** @type {KeyguardCommand.KeyguardCommand} */ ('derive-btc-xpub'),
    // Polygon
    /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-polygon-transaction'),
    /** @type {KeyguardCommand.KeyguardCommand} */ ('derive-polygon-address'),
    // Swap
    /** @type {KeyguardCommand.KeyguardCommand} */ ('sign-swap'),
];
```

types/KeyguardCommand.d.ts
```diff
@@ -0,0 +1,4 @@
+import * as _KeyguardCommand from '../client/src/KeyguardCommand';
+
+export as namespace KeyguardCommand;
+export = _KeyguardCommand;
```
danimoh: Could be nice to actually reimplement the enum from the client in the Keyguard's js.

src/request/connect/index.html
```diff
@@ -0,0 +1,133 @@
...
+                            <p data-i18n="connect-tooltip-multisig-3">
+                                The first address of your account will be visible to all participants in your shared wallet.
+                            </p>
+                        </div>
+                    </div>
+                </div>
+            </div>
+
+            <div class="page-footer nq-card-footer">
+                <button data-i18n="passwordbox-connect-account" class="nq-button light-blue continue display-none">Connect your account</button>
```
danimoh: Unused button

src/request/connect/Connect.js
```diff
@@ -0,0 +1,139 @@
...
+        this._passwordBox = new PasswordBox($passwordBox, {
+            hideInput: !request.keyInfo.encrypted,
+            buttonI18nTag: 'passwordbox-connect-account',
+            minLength: request.keyInfo.hasPin ? Key.PIN_LENGTH : undefined,
+        });
+
+        $button.addEventListener('click', () => {
+            $button.classList.add('display-none');
+            $passwordBox.classList.remove('display-none');
+        });
```
danimoh: Unused event handler for unused buton.

src/request/connect/Connect.js
```diff
@@ -0,0 +1,139 @@
...
+        try {
+            key = await KeyStore.instance.get(request.keyInfo.id, passwordBuf);
+        } catch (error) {
+            const errorMessage = error instanceof Error ? error.message : String(error);
+            if (errorMessage === 'Invalid key') {
+                TopLevelApi.setLoading(false);
+                this._passwordBox.onPasswordIncorrect();
+                return;
+            }
+            reject(new Errors.CoreError(errorMessage));
```
danimoh:
```suggestion
            reject(new Errors.CoreError(error instanceof Error ? error : errorMessage));
```
Passing the original error preserves the stack trace.

src/request/connect/Connect.js
```diff
@@ -0,0 +1,139 @@
...
+
+        /** @type {KeyguardRequest.SignatureResult[]} */
+        const signatures = [];
+
+        for (const keyPath of request.requestedKeyPaths) {
+            const publicKey = key.derivePublicKey(keyPath);
+
+            /** @type {Uint8Array} */
+            const messageBytes = Utf8Tools.stringToUtf8ByteArray(request.challenge);
+            const signature = key.signMessage(keyPath, messageBytes);
```
danimoh: This blindly signs any arbitrary text message as a valid Nimiq message signature. I.e. an attacker can trick the user into unknowingly signing messages with the only restriction enforced by request parsing being that the message must be textual, which is the most common case for messages to be approved by the user. This can for example be abused to authenticate at staking pool websites as the user.

To avoid this, a prefix separate to the Nimiq message signature prefix should be used, for example `\x18Nimiq Signed Challenge:\n` (in contrast to `\x16Nimiq Signed Message:\n` used for Nimiq messages).
Alternatively, the challenge can still be signed as a Nimiq message, but with an additional prefix enforced on the challenge. A separate prefix is arguably more elegant.

This is a breaking change, which also requires adaptions in apps using the Connect request.

src/request/connect/index.html
```diff
@@ -0,0 +1,133 @@
...
+                        <svg class="nq-icon"><use xlink:href="../../../node_modules/@nimiq/style/nimiq-style.icons.svg#nq-info-circle-small"/></svg>
+                        <div class="tooltip-box">
+                            <p data-i18n="connect-tooltip-multisig-1">
+                                Your account will be used to verify your identity only.
+                            </p>
+                            <p data-i18n="connect-tooltip-multisig-2">
+                                The shared wallet has its own address. It can receive transactions from anywhere just like a regular wallet.
+                            </p>
+                            <p data-i18n="connect-tooltip-multisig-3">
+                                The first address of your account will be visible to all participants in your shared wallet.
```
danimoh: Not the first address but the addresses for the `requestedKeyPaths`.

src/request/connect/ConnectApi.js
```diff
@@ -0,0 +1,75 @@
...
+            throw new Errors.InvalidRequestError('request is required');
+        }
+
+        const parsedRequest = {};
+        parsedRequest.appName = this.parseAppName(request.appName);
+        parsedRequest.keyInfo = await this.parseKeyId(request.keyId);
+        parsedRequest.keyLabel = this.parseLabel(request.keyLabel);
+        parsedRequest.requestedKeyPaths = request.requestedKeyPaths.map(
+            path => this.parsePath(path, 'requestedKeyPaths'),
+        );
```
danimoh, line 42: This can be abused by a malicious app to silently learn any number of the user's addresses. Those should probably be shown in the UI.

src/request/connect/Connect.js
```diff
@@ -0,0 +1,139 @@
...
+
+            signatures.push({
+                publicKey: publicKey.serialize(),
+                signature: signature.serialize(),
+            });
+        }
+
+        const rsaPublicCryptoKey = await key.getRsaPublicKey(Key.defaultEncryptionKeyParams);
+        const keyParams = /** @type {RsaKeyPairExport} */ (key.rsaKeyPair).keyParams;
```
danimoh:
```suggestion
        const keyParams = Key.defaultEncryptionKeyParams;
        const rsaPublicCryptoKey = await key.getRsaPublicKey(keyParams);
```

src/request/connect/Connect.js
```diff
@@ -0,0 +1,139 @@
...
+
+        /** @type {KeyguardRequest.ConnectResult} */
+        const result = {
+            signatures,
+            encryptionKey: {
+                format: 'spki',
+                keyData: new Uint8Array(await window.crypto.subtle.exportKey('spki', rsaPublicCryptoKey)),
+                algorithm: { name: 'RSA-OAEP', hash: 'SHA-256' },
```
danimoh: Can read algorithm from `rsaPublicCryptoKey.algorithm` instead of hard coding the constant here.

```suggestion
                algorithm: {
                    name: rsaPublicCryptoKey.algorithm.name,
                    hash: (/** @type {RsaHashedKeyAlgorithm} */ (rsaPublicCryptoKey.algorithm)).hash.name,
                },
```

client/src/PublicRequest.ts
```diff
@@ -492,6 +527,24 @@ export type DerivePolygonAddressResult = {
     }>,
 };
 
+export type ConnectRequest = SimpleRequest & {
+    appLogoUrl: string,
+    permissions: KeyguardCommand[],
```
danimoh: The permissions are not currently processed in a meaningful fashion. Neither are they displayed to the user, nor are they persisted or checked for. Therefore, this should be removed for now.

src/request/sign-multisig-transaction/index.html
```diff
@@ -0,0 +1,154 @@
...
+    <link rel="stylesheet" bundle-toplevel href="../../components/PasswordBox.css">
+
+    <link rel="stylesheet" href="../../components/Tooltip.css">
+    <link rel="stylesheet" href="../../components/Copyable.css">
+    <link rel="stylesheet" href="../../components/AddressInfo.css">
+    <link rel="stylesheet" href="../../components/AddressInfoDetailed.css">
+    <link rel="stylesheet" href="../../components/LoginFileAccountIcon.css">
+    <!-- <link rel="stylesheet" href="../../components/Timer.css"> -->
+    <!-- <link rel="stylesheet" href="../../components/PaymentInfoLine.css"> -->
+    <link rel="stylesheet" href="SignMultisigTransaction.css">
```
danimoh: `CurrencyInfo.js` and `FiatApi.js` are probably not necessary to be included.

src/request/sign-multisig-transaction/index.html
```diff
@@ -0,0 +1,154 @@
...
+                    </div>
+
+                    <div class="user-and-account-names">
+                        <span class="hide-with approving-as" data-i18n="sign-multisig-tx-approving-as">Approving as</span>
+                        <span class="hide-with user-name"></span>
+                        <span class="hide-with with" data-i18n="sign-multisig-tx-with">with</span>
+                        <span class="hide-as approving-with" data-i18n="sign-multisig-tx-approving-with">Approving with</span>
+                        <div class="login-file-account-icon"></div>
+                        <span class="account-name"></span>
+                    </div>
```
danimoh: The structure of these sentences could be entirely different in other languages, and also the little text segments do not provide enough context to the translators of what the complete sentence will be in the end. Therefore, this should be complete texts with placeholders, which are then later replaced in Javascript and assembled as DOM elements.

src/request/sign-multisig-transaction/index.html
```diff
@@ -0,0 +1,154 @@
...
+            <span id="global-close-text"></span>
+        </button>
+
+        <!-- Default and loading screen -->
+        <div id="loading" class="page">
+            <!-- Loading spinner SVG -->
+            <svg height="48" width="54" color="inherit" class="loading-spinner"><g>
+                <path class="big-hex" d="M51.9,21.9L41.3,3.6c-0.8-1.3-2.2-2.1-3.7-2.1H16.4c-1.5,0-2.9,0.8-3.7,2.1L2.1,21.9c-0.8,1.3-0.8,2.9,0,4.2 l10.6,18.3c0.8,1.3,2.2,2.1,3.7,2.1h21.3c1.5,0,2.9-0.8,3.7-2.1l10.6-18.3C52.7,24.8,52.7,23.2,51.9,21.9z" stroke="currentColor" stroke-width="3" fill="none" stroke-linecap="round" opacity="0.4" stroke-dasharray="92.5 60"/>
+                <path class="small-hex" d="M51.9,21.9L41.3,3.6c-0.8-1.3-2.2-2.1-3.7-2.1H16.4c-1.5,0-2.9,0.8-3.7,2.1L2.1,21.9c-0.8,1.3-0.8,2.9,0,4.2 l10.6,18.3c0.8,1.3,2.2,2.1,3.7,2.1h21.3c1.5,0,2.9-0.8,3.7-2.1l10.6-18.3C52.7,24.8,52.7,23.2,51.9,21.9z" stroke="currentColor" stroke-width="3" fill="none" stroke-linecap="round" stroke-dasharray="47.5 105"/>
+            </g></svg>
```
danimoh: Could apply the same little optimization from 3b69993e here, too. Equally in `connect/index.html`.

src/request/sign-multisig-transaction/SignMultisigTransaction.css
```diff
@@ -0,0 +1,210 @@
...
+    opacity: 0.5;
+    margin-bottom: 0.25rem;
+}
+
+#confirm-transaction .data-section {
+    margin: 1rem 3rem;
+    text-align: center;
+    max-width: 100%;
+    overflow-wrap: break-word;
+}
```
danimoh: This entire section is almost identical to `SignTransaction.css`. Thus, that one should simply be imported instead of duplicating everything.

src/request/sign-multisig-transaction/SignMultisigTransaction.css
```diff
@@ -0,0 +1,210 @@
...
+
+#confirm-transaction .user-and-account-names {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    white-space: nowrap;
+    font-size: 2rem;
+    gap: 0.5rem;
+    margin: 0 -2rem;
+    margin-bottom: -0.5rem;
```
danimoh:
```suggestion
    margin: 0 -2rem -0.5rem;
```

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -0,0 +1,284 @@
...
+
+        return parsedRequest;
+    }
+
+    /**
+     * @param {any} object
+     * @returns {MultisigConfig}
+     */
+    parseMultisigConfig(object) {
+        if (!object || typeof object !== 'object' || object === null) {
```
danimoh:
```suggestion
        if (!object || typeof object !== 'object') {
```

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -0,0 +1,284 @@
...
+                 * @returns {Nimiq.RandomSecret}
+                 */
+                secret => {
+                    if (!(secret instanceof Uint8Array)) {
+                        throw new Errors.InvalidRequestError('Invalid secrets: must be Uint8Arrays');
+                    }
+                    return new Nimiq.RandomSecret(secret);
+                },
+            );
+        } else if ('encrypted' in object.secrets) {
```
danimoh:
```suggestion
        } else if (object.secrets && typeof object.secrets === 'object' && 'encrypted' in object.secrets && 'keyParams' in object.secrets) {
```

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -0,0 +1,284 @@
...
+                 * @param {any} secret
+                 * @returns {Nimiq.RandomSecret}
+                 */
+                secret => {
+                    if (!(secret instanceof Uint8Array)) {
+                        throw new Errors.InvalidRequestError('Invalid secrets: must be Uint8Arrays');
+                    }
+                    return new Nimiq.RandomSecret(secret);
+                },
+            );
```
danimoh: If `secret` cannot be deserialized as a valid `RandomSecret`, the error is not mapped to a`InvalidRequestError`.
```suggestion
            try {
                for (const secret of object.secrets) {
                    if (!(secret instanceof Uint8Array)) throw new Error('must be Uint8Arrays');
                    secrets.push(new Nimiq.RandomSecret(secret));
                }
            } catch (error) {
                const errorMessage = error instanceof Error ? error.message : String(error);
                throw new Errors.InvalidRequestError(`Invalid secrets: ${errorMessage}`);
            }
```

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -0,0 +1,284 @@
...
+        } else if ('encrypted' in object.secrets) {
+            // Not checking fixed length here, to stay flexible for future increases of the number of commitments
+            if (!Array.isArray(object.secrets.encrypted) || object.secrets.encrypted.length < 2) {
+                throw new Errors.InvalidRequestError(
+                    'Invalid secrets.encrypted: must be an array with at least 2 elements',
+                );
+            }
```
danimoh: The length check is somewhat unnecessary and also not done above for unencrypted secrets.
Instead, a more sound check would be for `SignTransaction` to check that the number of my commitments, which it identifies after deriving my public key, matches the number of my secrets, which it currently doesn't do. If it would do that, no further check would be required that it should be at least two, because the commitments were all already checked to be at least two.

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -0,0 +1,284 @@
...
+                 */
+                array => array.length * 8 !== rsaKeySize,
+            )) {
+                throw new Errors.InvalidRequestError(
+                    'Invalid secrets.encrypted: all encrypted secrets must be the same length',
+                );
+            }
+            // Validate keyParams
+            if (!object.secrets.keyParams) {
+                throw new Errors.InvalidRequestError('Missing secrets.keyParams');
```
danimoh:
```suggestion
            if (!object.secrets.keyParams || typeof object.secrets.keyParams !== 'object') {
                throw new Errors.InvalidRequestError('Invalid secrets.keyParams');
```

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -0,0 +1,284 @@
...
+
+    /**
+     * @param {Nimiq.Transaction} transaction
+     * @param {MultisigConfig} multisig
+     */
+    verifyMultisigAddress(transaction, multisig) {
+        const multisigAddress = Nimiq.Address.fromPublicKeys(
+            multisig.publicKeys,
+            multisig.signers.length,
+        );
```
danimoh, line 226: This limits the Multisig support to k of n Multisigs, where each combination of signers which can sign together has identical size k and draws from the n public keys. More advanced combinations are not supported, for example of varying size, e.g. a Multisig which can be signed by single master key A, or keys B and C combined, or combinations with no overlap. But that's fine. If ever needed, support for such can be added in the future by instead of passing the `publicKeys`, passing something like `publicKeyCombinations`.
Maybe some documentation should be added to `PublicRequestTypes` explaining what the publicKeys represent (i.e. the n public keys of a k of n multisig, and not aggregated public keys calculated over pre-determined combinations), and that only k of n Multisigs are supported currently.

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -0,0 +1,284 @@
...
+            if (!transaction.recipient.equals(multisigAddress)) {
+                throw new Errors.InvalidRequestError(
+                    'Transaction recipient does not match calculated multisig address',
+                );
+            }
+        } else {
+            throw new Errors.InvalidRequestError(
+                'The multisig account must either be the sender or the recipient of the transaction',
+            );
+        }
```
danimoh: More concise and with documentation:
```suggestion
        if (!(transaction.senderType === Nimiq.AccountType.Basic && transaction.sender.equals(multisigAddress))
            // Vesting contracts owned by the multisig are allowed to send anywhere. The contract address can't be
            // verified here, it can however be assumed to be owned by the multisig, as only the owner can sign a valid
            // vesting withdrawal transaction.
            && transaction.senderType !== Nimiq.AccountType.Vesting
            // For other transactions which are not from a basic account, i.e. from a contract (other than a vesting
            // contract), we enforce the multisig to be the recipient. E.g. unstaking, HTLC refund, HTLC withdrawal etc.
            // are only allowed to the multisig address.
            && !(transaction.recipientType === Nimiq.AccountType.Basic && transaction.recipient.equals(multisigAddress))
        ) {
            throw new Errors.InvalidRequestError(
                'The transaction must either be send from or to the multisig, or from a vesting contract owned by it',
            );
        }
```

src/request/sign-multisig-transaction/SignMultisigTransaction.js
```diff
@@ -0,0 +1,309 @@
...
+        $accountName.textContent = request.keyLabel;
+
+        // Set up account icon
+        const $loginFileIcon = /** @type {HTMLDivElement} */ ($nameSection.querySelector('.login-file-account-icon'));
+        if (request.keyInfo.type === Nimiq.Secret.Type.ENTROPY) {
+            // eslint-disable-next-line no-new
+            new LoginFileAccountIcon(request.keyInfo.defaultAddress.toUserFriendlyAddress(), $loginFileIcon);
+        } else {
+            // Show identicon for legacy accounts (which must be supported to support Team Nimiq Multisig)
+            $loginFileIcon.innerHTML = ''; // Remove LoginFile icon
```
danimoh: The HTML element is already empty.

src/request/sign-multisig-transaction/SignMultisigTransaction.js
```diff
@@ -0,0 +1,309 @@
...
+                return;
+            }
+
+            for (let i = 0; i < secrets.length; i++) {
+                ownCommitmentPairs.push(new Nimiq.CommitmentPair(
+                    secrets[i],
+                    ownSigner.commitments[i],
+                ));
+            }
+        }
```
danimoh: The `for` loop can be deduplicated by first collecting only the `secrets` in the `if ... else ...` block, and doing the loop afterwards.
```suggestion
        /** @type {Nimiq.RandomSecret[]} */
        let ownCommitmentSecrets;
        if (Array.isArray(request.multisigConfig.secrets)) {
            ownCommitmentSecrets = request.multisigConfig.secrets;
        } else {
            // If we only have encrypted secrets, decrypt them first
            const rsaKey = await key.getRsaPrivateKey(request.multisigConfig.secrets.keyParams);

            try {
                ownCommitmentSecrets = await Promise.all(request.multisigConfig.secrets.encrypted.map(
                    async encrypted => new Nimiq.RandomSecret(new Uint8Array(
                        await window.crypto.subtle.decrypt(rsaKey.algorithm, rsaKey, encrypted),
                    )),
                ));
            } catch (error) {
                const errorMessage = error instanceof Error ? error.message : String(error);
                reject(new Errors.InvalidRequestError(`Cannot decrypt secrets: ${errorMessage}`));
                return;
            }
        }
        /** @type {Nimiq.CommitmentPair[]} */
        const ownCommitmentPairs = [];
        for (let i = 0; i < ownCommitmentSecrets.length; i++) {
            ownCommitmentPairs.push(new Nimiq.CommitmentPair(
                ownCommitmentSecrets[i],
                ownSigner.commitments[i],
            ));
        }
```

src/request/sign-multisig-transaction/SignMultisigTransaction.js
```diff
@@ -0,0 +1,309 @@
...
+        this._senderAddressInfo = new AddressInfo({
+            userFriendlyAddress: transaction.sender.toUserFriendlyAddress(),
+            label: request.senderLabel,
+            imageUrl: null,
+            accountLabel: null,
+            multisig: {
+                signers: request.multisigConfig.signers.length,
+                participants: request.multisigConfig.publicKeys.length,
+            },
+        });
```
danimoh: If the multisig is the recipient instead of the sender, the `multisig` property should be on the `_recipientAddressInfo`.

src/request/sign-multisig-transaction/SignMultisigTransaction.js
```diff
@@ -0,0 +1,309 @@
...
+     * @returns {string}
+     */
+    _formatData(transaction) {
+        if (Nimiq.BufferUtils.equals(transaction.data, Constants.CASHLINK_FUNDING_DATA)) {
+            return I18n.translatePhrase('funding-cashlink');
+        }
+
+        if (transaction.flags === Nimiq.TransactionFlag.ContractCreation) {
+            // TODO: Decode contract creation transactions
+            // return ...
```
danimoh: Even if not fully decoding the transaction data, it should at least hint at this being an interaction with a contract, be it contract creation, withdrawal, staking etc, based on sender and recipient type. Also even if there is no recipient data present.

src/components/AddressInfo.js
```diff
@@ -58,6 +69,18 @@ class AddressInfo { // eslint-disable-line no-unused-vars
             // eslint-disable-next-line no-new
             new Identicon(this._addressInfo.userFriendlyAddress, $identicon);
         }
+
+        if (this._addressInfo.multisig) {
+            const $badge = document.createElement('div');
+            $badge.classList.add('multisig-badge', 'nq-blue-bg');
+            $badge.innerHTML = TemplateTags.hasVars(2)`
+                ${this._addressInfo.multisig.signers}
+                <span class="of">of</span>
```
danimoh: Missing i18n.

src/components/AddressInfo.css
```diff
@@ -37,6 +38,28 @@
...
+    position: absolute;
+    right: -0.875rem;
+    bottom: -0.25rem;
+    border-radius: 5rem;
+    font-weight: bold;
+    font-size: 1.75rem;
+    padding: 0.375rem 0.75rem;
+    line-height: 1;
+    border: 0.5rem solid white;
+    word-spacing: -0.071428571em; // 1/14
```
danimoh: Invalid CSS comment (should be `/* 1/14 */`). But probably better as `calc(-1em / 14)` anyways.
```suggestion
    word-spacing: calc(-1em / 14);
```

src/components/LoginFileAccountIcon.js
```diff
@@ -0,0 +1,57 @@
+/* global LoginFileConfig */
+/* global IqonHash */
+
+class LoginFileAccountIcon { // eslint-disable-line no-unused-vars
+    /**
+     * @param {string} [address]
```
danimoh: Why is the address allowed to be `undefined`? It's nowhere used in such a fashion.

tools/build.sh
```diff
@@ -365,6 +366,7 @@ cp -v favicon.ico dist
 cp -rv src/assets/* dist/assets/
 cp -v src/lib/QrScannerWorker.js dist/lib/
 cp -v node_modules/@nimiq/style/nimiq-style.icons.svg dist/assets/
+cp -v src/lib/rsa/sandboxed/* dist/lib/rsa/sandboxed/
```
danimoh: Integrity hashes on the js files would be good.

demos/index.html
```diff
@@ -24,6 +24,9 @@
 <div><a href="/demos/RemoveKey.html">Remove</a></div>
 <div><a href="/demos/SignTransaction.html">Sign TX</a></div>
 <div><a href="/demos/DerivePolygonAddress.html">Derive Polygon Address</a></div>
+<div><a href="/demos/SignMultisigTransaction.html">Sign Multisig TX</a></div>
+<div><a href="/demos/RSAKeys.html">RSA Keys</a></div>
+<div><a href="/demos/ConnectAccount.html">Connect Account</a></div>
```
danimoh, line 29: These demos don't work currently.

danimoh (changes requested):
In my opinion, quite a bit too many shortcuts have been taken, for example regarding missing i18n, unencrypted storage of private keys and omission of of relevant information in the UI.

There were also several serious security issues:
- generated RSA keys are orders of magnitude weaker than they should be, due to the way they are generated.
- (almost) arbitrary Nimiq messages are blindly signed, which can be used to impersonate the user.
- malicious apps are able to trick the Keyguard into storing and using RSA private keys known to them.

Additionally, this PR points out opportunities for further security hardening, code cleanup, code deduplication, smaller fixes, stricter parsing etc..

danimoh (changes requested):
In my opinion, quite a bit too many shortcuts have been taken, for example regarding missing i18n, unencrypted storage of private keys and omission of of relevant information in the UI.

There were also several serious security issues:
- generated RSA keys are orders of magnitude weaker than they should be, due to the way they are generated.
- (almost) arbitrary Nimiq messages are blindly signed, which can be used to impersonate the user.
- malicious apps are able to trick the Keyguard into storing and using RSA private keys known to them.

Additionally, this PR points out opportunities for further security hardening, code cleanup, code deduplication, smaller fixes, stricter parsing etc..

danimoh:
> @danimoh Do you maybe have a better idea of how to solve the double KeyguardCommand.KeyguardCommand in this commit?
> https://github.com/nimiq/keyguard/commit/253219009034c528395025c8d42168deade52d9c

I tried multiple approaches.
Actually, the nested namespace can be avoided as easily as this:
```typescript
import { KeyguardCommand as _KeyguardCommand } from '../client/src/KeyguardCommand';

export as namespace KeyguardCommand;
export = _KeyguardCommand;
```

But that also occurred to me only later, and I think the approach that I took instead, which is replicating the client enums on the Javascript side of the Keyguard, is eventually nicer. This way, the enums are also usable in the Javascript in a type safe manner, without requiring any type assertions.

I first implemented the replicated enum in https://github.com/nimiq/keyguard/pull/524/commits/1f16c3d10dd181bedc6b164f07e6d747b03352a8, but then simplified / refined the approach in https://github.com/nimiq/keyguard/pull/524/commits/ea4bebed766669a1c949b38c813c7be3e7bbdcff. Both replication solutions are fully type checked and ensure equality with the original enum.

client/src/PublicRequest.ts
sisou: Resolved in #524

types/Keyguard.d.ts
sisou: Resolved in #524

package.json
sisou: Yes, and this PR has been updated to use a regular released package version.

client/src/PublicRequest.ts
sisou: Who knows? This just keeps the option open.

src/lib/rsa/sandboxed/forge.min.js.map
sisou: Resolved in #524

src/lib/rsa/sandboxed/forge.min.js
sisou: I added a check for exactly this hash in the build script in #532

src/lib/rsa/sandboxed/RSAKeysIframe.html
sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.js
sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.html
sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.js
sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.js
sisou: I think browser extensions could trigger these.

sisou: I re-added the check in #532, after it was removed in #524.

sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.js
sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.js
sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.js
sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.js
sisou: Resolved in #524

src/lib/rsa/sandboxed/RSAKeysIframe.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/Key.js
sisou: Resolved in #524

src/lib/KeyStore.js
sisou: Resolved in #524 & #526

src/lib/RequestParser.js
sisou: Resolved in #524

src/request/connect/index.html
sisou: Resolved in #524

src/request/connect/index.html
sisou: Resolved in #524

src/request/connect/index.html
sisou: Resolved in #524

src/request/connect/ConnectApi.js
sisou: Resolved in #524

types/KeyguardCommand.d.ts
sisou: Resolved in #524

src/request/connect/index.html
sisou: Resolved in #524

src/request/connect/Connect.js
sisou: Resolved in #524

src/request/connect/Connect.js
sisou: Resolved in #524

src/request/connect/Connect.js
sisou: Resolved in #524

src/request/connect/index.html
sisou: Resolved in #524

src/request/connect/ConnectApi.js
sisou: Resolved in #524

src/request/connect/Connect.js
sisou: Resolved in #524

src/request/connect/Connect.js
sisou: Resolved in #524

client/src/PublicRequest.ts
sisou: They are processed, stored, and checked in the Hub.

src/request/sign-multisig-transaction/index.html
sisou: Resolved in #524

src/request/sign-multisig-transaction/index.html
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransaction.css
sisou: Resolved in #524

src/request/sign-multisig-transaction/index.html
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransaction.css
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
sisou: But the check against the number of commitments only happens after password entering, whereas this check happens at the start of the request. The earlier the better.

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
sisou: I do not agree with the more concise version, its not understandable. The documentation can also be added to the more verbose layout.

src/request/sign-multisig-transaction/SignMultisigTransaction.js
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransaction.js
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransaction.js
sisou: Resolved in #524

src/request/sign-multisig-transaction/SignMultisigTransaction.js
sisou: Resolved in #524

src/components/AddressInfo.js
sisou: Resolved in #524

src/components/AddressInfo.css
sisou: Resolved in #524

src/components/LoginFileAccountIcon.js
sisou: Resolved in #524

tools/build.sh
sisou: Resolved in #524

client/src/PublicRequest.ts
sisou: No, instead I'll remove `SignMultisigTransactionResult` in #532 and type the result as `SignatureResult` instead to keep it simple.

tests/lib/RequestParser.spec.js
sisou: Added in #532

tests/lib/Key.spec.js
sisou: Fixed and re-enabled in #532

demos/index.html
sisou: Fixed in #532

client/src/PublicRequest.ts
danimoh: But if the Hub is doing it, still this code does not need to be in Keyguard, because the Keyguard is not doing any of it.

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
danimoh: There is not really an advantage in an early check. The check is to protect from invalid or malicious requests. In production, a working app should not be issuing invalid requests. During development, the dev might encounter the error during testing, and in that case indeed only after password entry, but for the dev that's fine. He is supposed to fix the issue after encountering it once. As for malicious requests, as the request is still eventually rejected, there is nothing to be gained, such that a malicious request for this error is also not to be expected in production.
I.e. for production, the additional, early check just bloats the code.

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
danimoh: I think the more concise version is actually better to understand, which is why I changed it. It clearly lists the acceptable cases, and does not mix acceptable and unacceptable cases.

PR #459, opened 2022-12-06
Safari sessionStorage partitioning fix

danimoh:
This pull request introduces a `NonPartitionedSessionStorage` class to circumvent the [issues that Safari introduced in 16.1. by partitioning `sessionStorage` between top-level and iframe contexts](https://bugs.webkit.org/show_bug.cgi?id=247565).
Under the hood, now cookies are used if the `sessionStorage` is detected as being partitioned.
For this, cookie handling has been unified in the `CookieJar` and generally been made more resilient and complete.
On top of the `CookieJar`, which handles plain cookies, a `CookieStorage` has been built which can encode arbitrary data of (practically) arbitrary size into cookies, and encrypts them by default to avoid sensitive data getting leaked to the server.
The cookie-stripping Cookie Monster service worker has now also been enabled as an additional measure to avoid cookies leaking to the server.
I did not observe the previous issues we had a few years ago in Safari with multiple service workers in the same tab (in the top-level window and the iframe). This was however on localhost (with wallet and hub built as production, to enable their service workers). Eventually, I would like to test this in a testnet deployment.

The login flow and swap flow have been adapted to now use the `NonPartitionedSessionStorage` instead of the `sessionStorage`.
These are accompanied with small changes to the Hub to forward the potentially created `CookieStorage` encryption keys.
I'll push those changes tomorrow. They'll also require a Keyguard client version update when released.

The `CookieStorage` and `NonPartitionedSessionStorage` have been implemented in a generic fashion, and not too specific to the login and swap use-cases, with the idea in mind to also be able to use them instead of the Safari keys cookie, to lift the limit on maximum accounts the user can be logged into. Also, the concept of the `NonPartitionedSessionStorage` can be extended to longer-lived storage to maybe be used as a replacement in Safari for the partitioned indexedDB, too.

I guess, I should also add some tests for the new libraries.

### DIFF
--- a/client/src/KeyguardClient.ts
+++ b/client/src/KeyguardClient.ts
@@ -148,9 +148,13 @@ export class KeyguardClient {
         return this._iframeRequest<EmptyRequest, SimpleResult>(KeyguardCommand.HAS_KEYS);
     }
 
-    public async deriveAddresses(keyId: string, paths: string[]): Promise<DerivedAddress[]> {
+    public async deriveAddresses(
+        keyId: string,
+        paths: string[],
+        tmpCookieEncryptionKey?: Uint8Array,
+    ): Promise<DerivedAddress[]> {
         return this._iframeRequest<DeriveAddressesRequest, DerivedAddress[]>
-            (KeyguardCommand.DERIVE_ADDRESSES, { keyId, paths });
+            (KeyguardCommand.DERIVE_ADDRESSES, { keyId, paths, tmpCookieEncryptionKey });
     }
 
     public async releaseKey(keyId: string, shouldBeRemoved = false): Promise<SimpleResult> {
--- a/src/lib/CookieJar.js
+++ b/src/lib/CookieJar.js
@@ -31,14 +31,14 @@ class CookieJar { // eslint-disable-line no-unused-vars
      * @param {KeyInfo[]} keys
      */
     static fillKeys(keys) {
-        this.writeCookie('k', this._encodeKeysCookie(keys));
+        this.writeCookie(CookieJar.Cookie.KEYS, this._encodeKeysCookie(keys));
     }
 
     /**
      * @returns {KeyInfo[]}
      */
     static eatKeys() {
-        const keysCookie = this.readCookie('k');
+        const keysCookie = this.readCookie(CookieJar.Cookie.KEYS);
         return keysCookie ? this._decodeKeysCookie(keysCookie) : [];
     }
 
@@ -47,7 +47,7 @@ class CookieJar { // eslint-disable-line no-unused-vars
      * @returns {AccountInfo[]}
      */
     static eatDeprecatedAccounts() {
-        const accountsCookie = this.readCookie('accounts');
+        const accountsCookie = this.readCookie(CookieJar.Cookie.DEPRECATED_ACCOUNTS);
         if (accountsCookie) {
             const decoded = decodeURIComponent(accountsCookie);
             const cookieAccounts = JSON.parse(decoded);
@@ -100,3 +100,23 @@ class CookieJar { // eslint-disable-line no-unused-vars
         });
     }
 }
+
+/**
+ * @readonly
+ * @enum { 'lang' | 'k' | 'removeKey' | 'accounts' | 'migrate' }
+ */
+CookieJar.Cookie = {
+    LANGUAGE: /** @type {'lang'} */ ('lang'),
+    KEYS: /** @type {'k'} */ ('k'),
+    REMOVE_KEY: /** @type {'removeKey'} */ ('removeKey'),
+    /**
+     * @deprecated
+     * @type {'accounts'}
+     */
+    DEPRECATED_ACCOUNTS: ('accounts'),
+    /**
+     * @deprecated
+     * @type {'migrate'}
+     */
+    DEPRECATED_MIGRATION_FLAG: ('migrate'),
+};
--- a/src/lib/CookieJar.js
+++ b/src/lib/CookieJar.js
@@ -1,14 +1,45 @@
 /* global KeyInfo */
+/* global Errors */
 
 class CookieJar { // eslint-disable-line no-unused-vars
     /**
      * @param {string} name
      * @param {string} value
-     * @param {number} [maxAge]
+     * @param {number} [maxAge] Time in seconds after which the cookie gets automatically deleted.
+     * @returns {string | null} Potential leftover data that did not fit the cookie.
      */
-    static writeCookie(name, value, maxAge = 31536000 /* 1 year */) {
-        const secure = window.location.protocol === 'https:' ? 'Secure;' : '';
-        document.cookie = `${name}=${value};max-age=${maxAge.toString()};${secure}SameSite=strict;Path=/`;
+    static writeCookie(name, value, maxAge = CookieJar.DEFAULT_MAX_AGE) {
+        const initialRawCookies = document.cookie;
+        const maxPayloadSize = CookieJar.MAX_COOKIE_SIZE - name.length;
+        const payload = value.substring(0, maxPayloadSize);
+        const cookieOptions = this._generateCookieOptions(maxAge); // do not count towards max size
+        document.cookie = `${name}=${payload}${cookieOptions}`;
+
+        // Check storage quota limits, in case we're not deleting a cookie.
+        if (maxAge > 0) {
+            if (this.readCookie(name) !== payload) {
+                throw new Errors.KeyguardError('Failed to write cookie. Quota exceeded?');
+            }
+            const initialCookies = this.readCookies(initialRawCookies);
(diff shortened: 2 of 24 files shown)


client/src/KeyguardClient.ts
```diff
@@ -148,9 +148,13 @@ export class KeyguardClient {
...
 
-    public async deriveAddresses(keyId: string, paths: string[]): Promise<DerivedAddress[]> {
+    public async deriveAddresses(
+        keyId: string,
+        paths: string[],
+        temporaryCookieStorageEncryptionKey?: Uint8Array,
+    ): Promise<DerivedAddress[]> {
         return this._iframeRequest<DeriveAddressesRequest, DerivedAddress[]>
-            (KeyguardCommand.DERIVE_ADDRESSES, { keyId, paths });
+            (KeyguardCommand.DERIVE_ADDRESSES, { keyId, paths, temporaryCookieStorageEncryptionKey });
```
sisou: Very long variable name :D Do we need the `temporary`? Can it be dropped?

src/lib/CookieJar.js
```diff
@@ -47,42 +137,36 @@ class CookieJar { // eslint-disable-line no-unused-vars
...
     /**
      * @param {string} str
      * @returns {KeyInfo[]}
      */
-    static _decodeCookie(str) {
+    static _decodeKeysCookie(str) {
         if (!str) return [];
 
-        const keys = str.split(',');
+        // previous versions used "," as separator which is strictly speaking not allowed in cookie values
```
sisou, line 168: :bulb:

src/lib/CookieStorage.js
```diff
@@ -0,0 +1,312 @@
...
+     * @param {CookieMetadata} metadata
+     * @returns {Uint8Array}
+     * @private
+     */
+    static _encodeMetadata(metadata) {
+        const encryptionInitVectorSize = metadata.encryptionInitVector ? metadata.encryptionInitVector.byteLength : 0;
+        if (metadata.chunkCount > 127
+            || (metadata.isEncrypted && !metadata.encryptionInitVector)
+            || (metadata.encryptionInitVector && encryptionInitVectorSize !== CookieStorage.ENCRYPTION_INIT_VECTOR_SIZE)
+        ) throw new Errors.KeyguardError('Invalid CookieStorage metadata.');
```
sisou, line 244: Suggestion to put the first condition already on a new line, indented, to match with the other lines.

```suggestion
        if (
            metadata.chunkCount > 127
            || (metadata.isEncrypted && !metadata.encryptionInitVector)
            || (metadata.encryptionInitVector && encryptionInitVectorSize !== CookieStorage.ENCRYPTION_INIT_VECTOR_SIZE)
        ) throw new Errors.KeyguardError('Invalid CookieStorage metadata.');
```

src/lib/CookieStorage.js
```diff
@@ -0,0 +1,312 @@
...
+     * @param {string} name
+     * @param {string} [namespace]
+     * @returns {CookieMetadata | null}
+     * @private
+     */
+    static _getMetaData(name, namespace) {
+        const firstChunk = CookieJar.readCookie(this._getChunkCookieName(name, 0, namespace));
+        if (!firstChunk) return null;
+        // Read enough base64 data to include the longest possible metadata, if the cookie is long enough.
+        return this._decodeMetadata(Nimiq.BufferUtils.fromBase64Url(firstChunk.substring(0, 23)));
```
sisou: The `23` is a bit arbitrary here, without explanation. Is it the length of IV_SIZE + 1 in base64 encoding? Can we maybe make this a formula?

src/lib/CookieStorage.js
```diff
@@ -0,0 +1,312 @@
...
+     * @param {string} name
+     * @param {number} chunk
+     * @param {string} [namespace = CookieJar.Cookie.NAMESPACE_COOKIE_STORAGE]
+     * @returns {string}
+     * @private
+     */
+    static _getChunkCookieName(name, chunk, namespace = CookieJar.Cookie.NAMESPACE_COOKIE_STORAGE) {
+        name = CookieJar.sanitizeCookieName(name, /* encodeInvalidChars */ false);
+        namespace = CookieJar.sanitizeCookieName(namespace, /* encodeInvalidChars */ false);
+        return `${namespace}${name}${chunk !== 0 ? `${chunk.toString(32)}` : ''}`;
```
sisou: `Number.prototype.toString()` supports a radix up to 36 according to Chrome DevTools, if you want to save all the space ;)

src/lib/NonPartitionedSessionStorage.js
```diff
@@ -0,0 +1,190 @@
...
+
+        if (isSessionStoragePartitioned) {
+            // Use CookieStorage. We currently only support encrypted CookieStorage entries, to encourage encryption.
+            return CookieStorage.set(name, data, {
+                encryptionKey,
+                maxAge: NonPartitionedSessionStorage.COOKIE_MAX_AGE,
+                namespace: CookieJar.Cookie.NAMESPACE_NON_PARTITIONED_SESSION_STORAGE,
+            });
+        }
+        sessionStorage.setItem(
```
sisou: Extra blank line to show this is the "else" block.
```suggestion
        }
        
        sessionStorage.setItem(
```

src/lib/CookieJar.js
```diff
@@ -95,3 +179,40 @@ class CookieJar { // eslint-disable-line no-unused-vars
...
+
+/**
+ * @readonly
+ * @enum { 'lang' | 'k' | 'removeKey' | 'ssnp' | 'cs.' | 'npss.' | 'accounts' | 'migrate' }
+ */
+CookieJar.Cookie = {
+    LANGUAGE: /** @type {'lang'} */ ('lang'),
+    KEYS: /** @type {'k'} */ ('k'),
+    REMOVE_KEY: /** @type {'removeKey'} */ ('removeKey'),
+    FLAG_SESSION_STORAGE_NON_PARTITIONED: /** @type {'ssnp'} */ ('ssnp'),
```
sisou:
```suggestion
    FLAG_SESSION_STORAGE_NOT_PARTITIONED: /** @type {'ssnp'} */ ('ssnp'),
```

src/lib/NonPartitionedSessionStorage.js
```diff
@@ -0,0 +1,190 @@
...
+    static isIframe() {
+        return window.top !== window;
+    }
+}
+// 7 minutes to simulate an ephemeral storage, similar to sessionStorage. Note that we set a max age and don't simply
+// use session cookies because session cookies live until the browser is closed, or even longer if the browser is
+// configured to persist sessions.
+NonPartitionedSessionStorage.COOKIE_MAX_AGE = 7 * 60; // 7 minutes; in seconds
+NonPartitionedSessionStorage.TOP_LEVEL_SESSION_STORAGE_ENTRY = 'topLevelEntry';
+NonPartitionedSessionStorage.IFRAME_SESSION_STORAGE_ENTRY = 'iframeEntry';
```
sisou:
```suggestion
NonPartitionedSessionStorage.TOP_LEVEL_SESSION_STORAGE_ENTRY = '_topLevelEntry';
NonPartitionedSessionStorage.IFRAME_SESSION_STORAGE_ENTRY = '_iframeEntry';
```

sisou (commented):
Wow, crazy stuff. Well done, but a lot of time spent on generalization :D

sisou (approved):

client/src/KeyguardClient.ts
mraveux: Could also be `tmpCookieStorageEncryptionKey` 🤓

danimoh: I went with `tmpCookieEncryptionKey`.

src/lib/CookieStorage.js
danimoh: Nah, that doesn't look nice :D

src/lib/CookieStorage.js
danimoh: I actually calculated it wrong, with the iv size as 16 bytes instead of 12 :D
(But higher numbers are not an issue, it was just base64 decoding more data than necessary.)

src/lib/CookieStorage.js
danimoh: Well, ok, why not.

danimoh:
Thanks for the review.
Yeah, I saw that the stored swap request will require support for cookie storage of arbitrary size, so that led me somewhat down a rabbit hole with the generalization. But now we can use it for the keys cookie too :)

I addressed your comments @sisou.
See here for the diff: https://github.com/nimiq/keyguard/compare/c2dc8334a6978d51e24bc7ec3287d92fe663026e..aa01a3187b3b3f83fa7b68f025994de2c3ca11dc

sisou (approved):

PR #460, opened 2022-12-06
Update GH actions & small tsconfig cleanup

sisou:
GH actions v2 were throwing deprecation warnings, v3 are the current versions.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,11 +14,11 @@
     // "declaration": true,                   /* Generates corresponding '.d.ts' file. */
     // "sourceMap": true,                     /* Generates corresponding '.map' file. */
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
-    "outDir": "./.ts-out",                    /* Redirect output structure to the directory. */
+    // "outDir": "./.ts-out",                 /* Redirect output structure to the directory. */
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -56,13 +56,7 @@
...
   },
   "include": [
-    "types/Nimiq.d.ts",
-    "types/Keyguard.d.ts",
-    "types/ErrorContainer.d.ts",
-    "types/MessagePrefix.d.ts",
-    "types/BarcodeScanner.d.ts",
-    "types/BitcoinJS.d.ts",
-    "types/UserAgentClientHints.d.ts",
+    "types/*",
```
danimoh, line 63: Note that this now also includes `KeyguardRequestNamespace.d.ts`.
Don't know whether that's any issue though.

danimoh (approved):

tsconfig.json
sisou: It's also included via `Keyguard.d.ts`, but the CI still works, so it doesn't seem to be a problem that it's doubled.

sisou: Actually removed the type reference in `Keyguard.d.ts` now, so it's no longer double.

mraveux (approved):

PR #461, opened 2022-12-06
Add autocomplete attribute to password inputs

sisou:
Chrome was always warning about missing autocomplete attribute for the password inputs. This adds it.

Defaults to `current-password`, so the regular PasswordBox gets that, as it doesn't set anything in the constructor. For the PasswordSetterBox, we select `new-password`, so that browsers and password managers can suggest a new safe password.

### DIFF
--- a/src/components/PasswordInput.js
+++ b/src/components/PasswordInput.js
@@ -6,19 +6,22 @@
 class PasswordInput extends Nimiq.Observable {
     /**
      * @param {?HTMLElement} $el
-     * @param {number} [maxLength]
-     * @param {string} [placeholder]
+     * @param {object} [options]
+     * @param {number} [options.maxLength]
+     * @param {string} [options.placeholder]
+     * @param {'current-password' | 'new-password'} [options.autocomplete]
      */
-    constructor($el, maxLength, placeholder = '••••••••') {
+    constructor($el, options = {}) {
         super();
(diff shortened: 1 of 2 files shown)


src/components/PasswordInput.js
```diff
@@ -6,19 +6,22 @@
 class PasswordInput extends Nimiq.Observable {
     /**
      * @param {?HTMLElement} $el
-     * @param {number} [maxLength]
-     * @param {string} [placeholder]
+     * @param {object} [options]
+     * @param {number} [options.maxLength]
+     * @param {string} [options.placeholder]
+     * @param {'current-password' | 'new-password'} [options.autocomplete]
```
danimoh: Two optional improvements:
- The default value can be specified for documentation purposes.
- In my PR #459 I noticed that typescript only allows a property to be omitted if it's marked as optional via `[...]`. It does not allow it to be passed as undefined. We don't use that here, but if we want to reflect it in the types, they can be marked as undefineable by annotating them with `=` or `| undefined`.

```suggestion
     * @param {number=} [options.maxLength = Infinity]
     * @param {string=} [options.placeholder = '••••••••']
     * @param {'current-password' | 'new-password' | undefined} [options.autocomplete = 'current-password']
```

danimoh (approved):

mraveux (approved):
👍

PR #463, opened 2022-12-16
Fixes to PR#459 (Safari sessionStorage partitioning fix)

danimoh:
This PR includes 4 fixes to previous [PR#459](https://github.com/nimiq/keyguard/pull/459) (Safari sessionStorage partitioning fix):
- Regex special chars in cookie names lead to this cookie not being found in `document.cookie` and not being readable with `CookieJar.readCookie`.
- On iOS, all browsers use Safari's WebKit enginge, however browsers other than Safari were not detected as partitioning the sessionStorage. I think it should be working now, but unfortunately, I can't test this with BrowserStack.
- The previous Safari version check did not check the version parts individually and would for example wrongly have identified 17.0. as not being newer than 16.1.
- Firefox does not forward the page referrer to `document.referrer` if a service worker is involved. A workaround has been implemented to still make it work.

Additionally, the feature detection whether the sessionStorage is partitioned has been improved by detecting cases as being partitioned where the Keyguard is known to be open as top level in the same tab before, but the top level sessionStorage entries are not readable.

### DIFF
--- a/src/lib/NonPartitionedSessionStorage.js
+++ b/src/lib/NonPartitionedSessionStorage.js
@@ -151,9 +151,23 @@ class NonPartitionedSessionStorage {
         // instead of hub.nimiq.com). If this change happens, the Keyguard's sessionStorage would not be partitioned
         // anymore between the Keyguard top-level and Keyguard iframe within the Hub, because the Keyguard and Hub live
         // on the same eTLD+1. This issue is tracked here: https://bugs.webkit.org/show_bug.cgi?id=247565
-        if (!BrowserDetection.isSafari()) return false;
-        const safariVersion = BrowserDetection.safariVersion();
-        return safariVersion[0] >= 16 && safariVersion[1] >= 1;
+        // On iOS, we have to include all browsers, because they all use Safari's WebKit engine, the version of which is
+        // determined by the iOS version.
+        /**
+         * @param {number[]} version
+         * @param {number[]} referenceVersion
+         * @returns {boolean}
+         */
+        const isVersionAtLeast = (version, referenceVersion) => {
+            for (let i = 0; i < Math.min(version.length, referenceVersion.length); ++i) {
+                if (version[i] > referenceVersion[i]) return true;
+                if (version[i] < referenceVersion[i]) return false;
+            }
+            return version.length >= referenceVersion.length;
+        };
+        const firstBadSafari = [16, 1];
+        return (BrowserDetection.isSafari() && isVersionAtLeast(BrowserDetection.safariVersion(), firstBadSafari))
+            || (BrowserDetection.isIOS() && isVersionAtLeast(BrowserDetection.iOSVersion(), firstBadSafari));
     }
 
     static hasSessionStorageAccess() {
--- a/src/lib/NonPartitionedSessionStorage.js
+++ b/src/lib/NonPartitionedSessionStorage.js
@@ -113,35 +113,55 @@ class NonPartitionedSessionStorage {
         }
 
         // Create a hash of the user agent to consume less space in the flag cookie than the complete user agent.
+        // We store a hash of the user agent string, to re-run the check on browser update because the partitioning
+        // behavior can change over time.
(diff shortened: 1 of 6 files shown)


src/lib/NonPartitionedSessionStorage.js
```diff
@@ -113,35 +113,47 @@ class NonPartitionedSessionStorage {
...
             // opened as top-level and then as iframe, or vice-versa, in the same tab, because sessionStorage always
             // exists per tab only. Note that this check would be inaccurate if the iframe would be embedded into a
             // Keyguard top level window, because no partitioning occurs if the top-level window and the iframe are on
             // the same origin. As the Keyguard is only ever embedded as iframe into the Hub, this check is accurate.
-            // Also note, that in the opposite case, where an entry is not readable in the other context, nothing is
-            // proven because the Keyguard might not have been opened in the other context before in the same tab.
-            // We store a hash of the user agent string, to re-run the check on browser update because the partitioning
-            // behavior can change over time.
-            CookieJar.writeCookie(CookieJar.Cookie.FLAG_SESSION_STORAGE_NOT_PARTITIONED, userAgentHashString);
+            CookieJar.writeCookie(CookieJar.Cookie.FLAG_SESSION_STORAGE_PARTITIONED, `0${userAgentHashString}`);
```
sisou: Can you make the `0` and `1` prefixes into variables? E.g. an "enum": 
```
const PREFIXES = {
    WHATEVER_0_MEANS: '0',
    WHATEVER_1_MEANS: '1',
}
```

Because I don't understand what they mean.

src/lib/NonPartitionedSessionStorage.js
```diff
@@ -113,35 +113,47 @@ class NonPartitionedSessionStorage {
...
+            CookieJar.writeCookie(CookieJar.Cookie.FLAG_SESSION_STORAGE_PARTITIONED, `1${userAgentHashString}`);
+            return true;
         }
 
-        const flagNonPartitioned = CookieJar.readCookie(CookieJar.Cookie.FLAG_SESSION_STORAGE_NOT_PARTITIONED);
-        if (flagNonPartitioned) {
-            if (flagNonPartitioned === userAgentHashString) return false;
+        const flagPartitioned = CookieJar.readCookie(CookieJar.Cookie.FLAG_SESSION_STORAGE_PARTITIONED);
+        if (flagPartitioned) {
+            if (flagPartitioned.substring(1) === userAgentHashString) return flagPartitioned[0] === '1';
```
sisou: Then you can test here against the variable, giving the `=== '1'` more meaning.

sisou (approved):
Wild stuff.

Just one suggestion to name the hard-coded constants.

Also, the referrer fix for Firefox could have been just a disabling of the serviceworker for non-iOS/Safari environments, no? Seems easier.

src/lib/NonPartitionedSessionStorage.js
```diff
@@ -182,6 +216,14 @@ class NonPartitionedSessionStorage {
 NonPartitionedSessionStorage.COOKIE_MAX_AGE = 7 * 60; // 7 minutes; in seconds
 NonPartitionedSessionStorage.TOP_LEVEL_SESSION_STORAGE_ENTRY = '_topLevelEntry';
 NonPartitionedSessionStorage.IFRAME_SESSION_STORAGE_ENTRY = '_iframeEntry';
+/**
+ * @readonly
+ * @enum { '1' | '0' }
+ */
+NonPartitionedSessionStorage.PREFIX_FLAG_SESSION_STORAGE_PARTITIONED = {
```
sisou, line 224: The word session storage is twice in this. Suggestion to rename:
```suggestion
NonPartitionedSessionStorage.PREFIX_FLAG_IS_PARTITIONED = {
```

danimoh: The name `PREFIX_FLAG_SESSION_STORAGE_PARTITIONED` is meant as "prefix" of the "FLAG_SESSION_STORAGE_PARTITIONED" cookie.
-> Added a comment.

PR #465, opened 2023-03-03
USDC-on-Polygon integration

sisou:
Generate Polygon address(es) when creating a new account or logging in, and add request to sign OpenGSN relay requests for transferring USDC via Nimiq's USDC-transfer-contract. Also add USDC support to sign-swap request(s).

PR #467, opened 2023-03-19
Upgrade Typescript to v5

sisou:
v4 -> v5 was no changes actually, see the commit history.

~This is based on the `usdc` branch, so can only be merged after #465 is merged.~


src/components/RecoveryWords.js
```diff
@@ -157,9 +165,10 @@ class RecoveryWords extends Nimiq.Observable {
     _setFocusToNextInput(index, paste) {
         index = Math.max(index, 0);
         if (index < this.$fields.length) {
-            this.$fields[index].focus();
-            if (paste) {
-                this.$fields[index].fillValueFrom(paste);
+            const field = this.$fields[index];
+            field.focus();
+            if (paste && 'fillValueFrom' in field) {
```
onmax, line 170:
```suggestion
            if (paste && 'fillValueFrom' in field && typeof field.fillValueFrom === 'function') {
```

not sure if this is relevant

src/components/RecoveryWords.js
```diff
@@ -123,18 +130,19 @@ class RecoveryWords extends Nimiq.Observable {
...
-            const mnemonic = this.$fields.map(field => field.value);
+            const mnemonic = this.$fields.map(field => ('value' in field ? field.value : '-'));
             const type = Nimiq.MnemonicUtils.getMnemonicType(mnemonic); // throws on invalid mnemonic
             this._mnemonic = { words: mnemonic, type };
             this.fire(RecoveryWords.Events.COMPLETE, mnemonic, type);
         } catch (e) {
-            if (e.message !== 'Invalid checksum') {
+            const err = /** @type {Error} */ (e);
+            if (err.message !== 'Invalid checksum') {
                 console.error(e); // eslint-disable-line no-console
```
onmax:
```suggestion
                console.error(err); // eslint-disable-line no-console
```

_Super picky comment_

I wouldn't use `e` after defining `err` with a type

tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     /* Basic Options */
-    "target": "ES2015",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
-    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "target": "ES2020",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
```
onmax:
```suggestion
    "target": "ES2020",                       /* Specify ECMAScript target version. */
```

tsconfig.json
```diff
@@ -1,8 +1,8 @@
...
+    "module": "CommonJS",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
```
onmax, line 6:
```suggestion
    "module": "CommonJS",                     /* Specify module code generation. */
```

onmax (approved):
LGTM

tsconfig.json
onmax: People can learn what values are valid from the docs so we don't have these comments out of sync ?

danimoh (approved):
Cool, thanks for this.
I added my improvements, feel free to have a look.
The change of static properties of classes to actual js `static`s requires an eslint update, so this will be in a separate PR.

@onmax please also review my changes, if you want.

danimoh:
The force-push is for rebase onto master, with resolved conflicts in the `FiatApi`.
However, some additional type errors seem to have surfaced on the latest changes from master.
Taking a look now.

src/lib/LoginFile.js
```diff
@@ -65,9 +65,12 @@ class LoginFile {
...
         await this._drawPromise;
 
-        return new Promise(resolve => {
+        return new Promise((resolve, reject) => {
             this.$canvas.toBlob(blob => {
-                if (!blob) throw new Error('Cannot generate URL');
+                if (!blob) {
+                    reject(new Error('Cannot generate URL'));
+                    return;
+                }
```
sisou: TIL throwing in the `new Promise()` callback doesn't reject the promise. The more you know...

tsconfig.json
```diff
@@ -4,11 +4,6 @@
     "target": "ES2020",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
     "module": "CommonJS",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
     "lib": ["DOM"],                           /* Specify library files to be included in the compilation. */
-    "skipLibCheck": true,                     /* Skip checking .d.ts files. We need this currently, because types in */
-                                              /* lib.webworker.d.ts referenced in ServiceWorker.js clash with types */
-                                              /* in lib.dom.d.ts, and also because our typescript version is too old */
-                                              /* to handle the types of ethers. */
-                                              /* TODO once typescript has been updated, this can be removed. */
```
sisou, line 11: Well spotted! :+1:

src/common.js
```diff
@@ -88,8 +88,7 @@ async function runKeyguard(RequestApiClass, opts) { // eslint-disable-line no-un
 
     // Back arrow functionality
     document.body.addEventListener('click', event => {
-        // @ts-ignore (Property 'matches' does not exist on type 'EventTarget'.)
-        if (!event.target || !event.target.matches('a.page-header-back-button')) return;
+        if (!(event.target instanceof HTMLElement && event.target.matches('a.page-header-back-button'))) return;
```
sisou: I liked the previous double-negative OR syntax better.

src/components/PasswordBox.js
```diff
@@ -41,7 +41,7 @@ class PasswordBox extends Nimiq.Observable {
 
         this.$el.classList.toggle('hide-input', this.options.hideInput);
 
-        this._passwordInput = new PasswordInput(this.$el.querySelector('[password-input]'));
+        this._passwordInput = new PasswordInput(/** @type {HTMLElement} */(this.$el.querySelector('[password-input]')));
```
sisou, line 44: Why is this change necessary? I thought with your changes `null | undefined` are also allowed?

src/components/PasswordSetterBox.js
```diff
@@ -27,7 +27,7 @@ class PasswordSetterBox extends Nimiq.Observable {
         this.$el = PasswordSetterBox._createElement($el, this.options);
 
         this._passwordInput = new PasswordInput(
-            this.$el.querySelector('[password-input]'),
+            /** @type {HTMLElement} */ (this.$el.querySelector('[password-input]')),
```
sisou, line 30: Same here.

src/components/PasswordSetterBox.js
```diff
@@ -52,7 +52,7 @@ class PasswordSetterBox extends Nimiq.Observable {
     }
 
     /**
-     * @param {?HTMLFormElement} $el
+     * @param {?HTMLFormElement | undefined} $el
```
sisou, line 55: What's the difference between `| undefined` in the type and putting `$el` in square brackets?

src/common.js
danimoh: Me theoretically too, I only changed it to make it fit into a single line.

src/components/PasswordBox.js
danimoh: I think that it worked previously, is actually a typescript bug. `querySelector` returns `Element | null`, `PasswordInput` previously accepted `?HTMLElement` which is equivalent to `HTMLElement | null`, which should refuse a passed value of `Element | null` but didn't. Now it's `{?HTMLElement} [$el]`, which is equivalent to `HTMLElement | null | undefined`, which somehow trigger typscript to actually complain (i.e. it's now actually more restrictive), which is why a type assertion is now required. This is the way it should be imo.

src/components/PasswordSetterBox.js
danimoh: https://github.com/nimiq/keyguard/pull/467#discussion_r1746968333

src/components/PasswordSetterBox.js
danimoh: `[$el]` is an optional parameter which can be omitted. A parameter with `| undefined` can be set to `undefined` but not omitted. An optional parameter can not precede a non-optional parameter, which is why here it needs to be declared as `| undefined`.

In the old ts version, there was also another difference, which I now didn't observe anymore with the newer version: https://github.com/nimiq/keyguard/pull/461#discussion_r1045729111

src/common.js
sisou: Changed it back to double negative, that version is only one character longer and still fits into one line.

src/components/RecoveryWords.js
sisou: I don't think this is necessary.

src/common.js
danimoh: Oh, right :+1:
Sometimes in my IDE lines look longer when it shows names of the passed parameters.

PR #470, opened 2023-10-12
Fix bug in iOS>=16.4 where the right side of a rotation-container was…

sisou:
… not interactive

By making the rotation container not interactive, but the children yes, everything on the page can be clicked (tapped) again.


Eligioo (approved):
Tested this locally on iPhone 13 iOS 16.6.1 and can confirm that this solves the issue.

PR #484, opened 2023-12-06
Support sending new native USDC

sisou:
Add the necessary config, handlers, validators and implementations to allow signing new native USDC transactions, which use a slightly different contract ABI, but which works the same. Approval is now called Permit, and the permit is signed with different data.

The Hub changes that go with this are minimal: https://github.com/nimiq/hub/commit/3504e772bbab30ddbec7722e765b4412b6c024f2


src/lib/polygon/PolygonKey.js
```diff
@@ -97,6 +97,55 @@ class PolygonKey { // eslint-disable-line no-unused-vars
...
+                { name: 'deadline', type: 'uint256' },
+            ],
+        };
+
+        const message = {
+            owner: ownerAddress,
+            spender: forwarderContractAddress,
+            value: approvalAmount,
+            nonce: tokenNonce,
+            deadline: ethers.constants.MaxUint256,
```
mraveux, line 128: Should the permit never expire? I don't know much but couldn't that be an issue?

src/lib/polygon/PolygonKey.js
```diff
@@ -97,6 +97,55 @@ class PolygonKey { // eslint-disable-line no-unused-vars
...
     }
 
+    /**
+     * @param {string} path
+     * @param {string} forwarderContractAddress
+     * @param {ethers.BigNumber} approvalAmount
+     * @param {number} tokenNonce
+     * @param {string} ownerAddress
+     * @returns {Promise<{sigR: string, sigS: string, sigV: number}>}
+     */
```
mraveux, line 103:
```suggestion
    /**
     * Signs a USDC permit for a specific transaction on the Polygon network.
     *
     * @param {string} path - The derivation path for the key.
     * @param {string} forwarderContractAddress - The address of the forwarder contract.
     * @param {ethers.BigNumber} approvalAmount - The amount of USDC to approve.
     * @param {number} tokenNonce - The nonce of the token.
     * @param {string} ownerAddress - The address of the owner.
     *
     * @returns {Promise<{sigR: string, sigS: string, sigV: number}>} - A promise that resolves to an object containing the components of the signature:
     * - sigR: The R component of the signature, a 32-byte number.
     * - sigS: The S component of the signature, a 32-byte number.
     * - sigV: The V component of the signature, a recovery value.
     */
```

mraveux (approved):

src/lib/polygon/PolygonKey.js
sisou: This is a requirement by the transfer contract. If given another `deadline`, the generated permit will not be valid for the contract. Maybe @mar-v-in can elaborate why the infinite deadline is required?

src/lib/polygon/PolygonKey.js
sisou: I feel the descriptions of the parameters just repeat the same words already included in the parameter name itself: `forwarderContractAddress - The address of the forwarder contract`, `tokenNonce - The nonce of the token`. So I don't think this is necessary. The only argument for which the description expands on the name is for `path`, which could be called `derivationPath` instead to fix it more easily.

src/lib/polygon/PolygonKey.js
mar-v-in: The deadline is referring to the validity of the transaction. A transaction is also invalidated by having a new transaction with the same nonce, so if you really wanted to invalidate a transaction, one could just send a new transaction. This is the same model used by normal EVM transactions, so it seems reasonable safe to me to use it here as well.

The reason why we set deadline to infinity is that a) we didn't have that in the previous version and b) putting any actual value increases complexity everywhere (that value would also need to be part of the transaction, parsed by the smart contract, ...).

src/lib/polygon/PolygonContractABIs.full.js
```diff
@@ -0,0 +1,299 @@
...
+        'function verify(tuple(address from, address to, uint256 value, uint256 gas, uint256 nonce, bytes data, uint256 validUntil) forwardRequest, bytes32 domainSeparator, bytes32 requestTypeHash, bytes suffixData, bytes signature) view',
+        'function versionPaymaster() view returns (string)',
+        'function versionRecipient() view returns (string)',
+        'function withdraw(uint256 amount, address target)',
+        'function withdrawRelayHubDeposit(uint256 amount, address target)',
+        'function withdrawToken(address token, uint256 amount, address target)',
+        'function wrappedChainToken() view returns (address)',
+    ],
+};
+/* eslint-enable max-len */
```
danimoh: This is unused code, right?
If it's just for documentation/backup purposes, it would better be a txt or md file.

src/lib/polygon/PolygonContractABIs.js
```diff
@@ -1,181 +1,27 @@
...
-        // 'function withdrawRelayHubDeposit(uint256 amount, address target)',
-        // 'function withdrawToken(address token, uint256 amount, address target)',
-        // 'function wrappedChainToken() view returns (address)',
+    ],
+
+    NATIVE_USDC_CONTRACT_ABI: [],
+
+    NATIVE_USDC_TRANSFER_CONTRACT_ABI: [
+        'function transfer(address token, uint256 amount, address target, uint256 fee)',
+        'function transferWithPermit(address token, uint256 amount, address target, uint256 fee, uint256 value, bytes32 sigR, bytes32 sigS, uint8 sigV)',
```
danimoh, line 24: Is `value` the permit?
Can this name be changed?

src/lib/polygon/PolygonKey.js
```diff
@@ -97,6 +97,55 @@ class PolygonKey { // eslint-disable-line no-unused-vars
...
+    async signUsdcPermit(path, forwarderContractAddress, approvalAmount, tokenNonce, ownerAddress) {
+        // TODO: Make the domain parameters configurable in the request?
+        const domain = {
+            name: 'USD Coin', // This is currently the same for testnet and mainnet
+            version: '2', // This is currently the same for testnet and mainnet
+            verifyingContract: CONFIG.NATIVE_USDC_CONTRACT_ADDRESS,
```
danimoh, line 109: I wonder whether the old contract should actually also become `LEGACY_USDC_CONTRACT_ADDRESS` or `BRIDGED_USDC_CONTRACT_ADDRESS`, now that native usdc is meant to become the main usdc.

src/lib/polygon/PolygonKey.js
```diff
@@ -97,6 +97,55 @@ class PolygonKey { // eslint-disable-line no-unused-vars
...
+            domain,
+            types,
+            message,
+        );
+
+        const sigR = signature.slice(0, 66); // 0x prefix plus 32 bytes = 66 characters
+        const sigS = `0x${signature.slice(66, 130)}`; // 32 bytes = 64 characters
+        const sigV = parseInt(signature.slice(130, 132), 16); // last byte = 2 characters
+
+        return { sigR, sigS, sigV };
```
danimoh: The conversion from `signature` to `sigR`, `sigS`, `sigV` could be moved into a separate method to avoid this little bit of code duplication.

client/src/PublicRequest.ts
```diff
@@ -226,6 +226,14 @@ export type PolygonTransactionInfo = {
...
     },
+
+    /**
+     * The sender's nonce in the token contract, required when calling the
+     * contract function `transferWithPermit`.
+     */
+    permit?: {
+        tokenNonce: number,
+    },
 };
```
danimoh: Add a comment that permit is for native USDC and approval for bridged USDC.e

danimoh (approved):
Nice and clean.

src/lib/polygon/PolygonContractABIs.full.js
sisou: yeah, that makes sense :+1:

src/lib/polygon/PolygonContractABIs.js
sisou: It is the permit amount yes. Unfortunately, it cannot be changed, as the hash of the function signature is used for it's identification, meaning changing any argument name changes the ID.

src/lib/polygon/PolygonKey.js
sisou: Can be done, yes. I will do it in a follow up PR to not litter this one :+1:

PR #485, opened 2024-01-06
Correctly count length of password with multibyte characters

sisou:
This PR fixes the way password character length is counted to use the [recommended method from MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#strings_with_length_not_equal_to_the_number_of_characters) for strings with potential multibyte characters.

Additionally, this method is applied to the password strength calculator to always handle full characters, not ["lone surrogates"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt#using_charat).

### DIFF
--- a/src/lib/PasswordStrength.js
+++ b/src/lib/PasswordStrength.js
@@ -45,13 +45,21 @@ class PasswordStrength {
 
         const baseScore = 30;
 
-        for (let i = 0; i < password.length; i++) {
-            if (password.charAt(i).match(/[A-Z]/g)) { count.upperCase += 1; }
-            if (password.charAt(i).match(/[0-9]/g)) { count.numbers += 1; }
-            if (password.charAt(i).match(/(.*[!,@,#,$,%,^,&,*,?,_,~])/)) { count.symbols += 1; }
+        const characters = [...password];
+        const length = characters.length;
+
+        for (let i = 0; i < length; i++) {
+            if (characters[i].match(/[A-Z]/g)) {
+                count.upperCase += 1;
+            } else if (characters[i].match(/[0-9]/g)) {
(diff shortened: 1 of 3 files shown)


src/lib/PasswordStrength.js
```diff
@@ -45,13 +45,21 @@ class PasswordStrength {
...
-            if (password.charAt(i).match(/(.*[!,@,#,$,%,^,&,*,?,_,~])/)) { count.symbols += 1; }
+        const characters = [...password];
+        const length = characters.length;
+
+        for (let i = 0; i < length; i++) {
+            if (characters[i].match(/[A-Z]/g)) {
+                count.upperCase += 1;
+            } else if (characters[i].match(/[0-9]/g)) {
+                count.numbers += 1;
+            } else if (characters[i].match(/[^a-z]/)) {
```
onmax: Subjective take: Easier to read 
`!characters[i].match(/[a-z]/)`, instead of
`characters[i].match(/[^a-z]/)`

Probably not worth changing

onmax (approved):
Looks good to me. Now, I have one password with emojis 🥳

Eligioo (approved):

Eligioo:
LGTM

Albermonte:
I would include some comments on why using the spread operator makes a difference in the lenght.

Otherwise, LGTM 👍🏼

Albermonte (approved):

src/lib/PasswordStrength.js
sisou: Thanks, this is a great suggestion! Code readability is a priority in the security-relevant Keyguard.

PR #488, opened 2024-01-31
Allow swapping bridged USDC.e to native USDC via new OpenGSN-enabled swap contract

sisou:
Removes the possibility of sending USDC.e, adds the ability to swap it to USDC.

The combined diff might be confusing, as the requests all look similar and I removed the `transferWithApproval` from the top of the handler and added `swapWithApproval` to the bottom of the handler, which messes up Github's diff calculations. I suggest to view the two commit diffs individually.


src/request/sign-polygon-transaction/SignPolygonTransaction.js
```diff
@@ -34,18 +34,24 @@ class SignPolygonTransaction {
...
-            request.description.name === 'refund' ? 'usdc' : 'none',
-        ).renderTo($recipient);
+        if (request.description.name === 'swap' || request.description.name === 'swapWithApproval') {
+            new PolygonAddressInfo(relayRequest.from, 'USDC', 'usdc').renderTo($recipient);
+        } else {
+            const recipientAddress = /** @type {string} */ (request.description.args.target);
+            new PolygonAddressInfo(
+                recipientAddress,
+                request.description.name === 'refund' ? request.keyLabel : request.recipientLabel,
+                request.description.name === 'refund' ? 'usdc' : 'none',
```
danimoh: I know this was like this before, but now that this code has been split into multiple `if` cases, I would split the refund case into a separate `if`, too, same as is done for the sender.

src/request/sign-polygon-transaction/SignPolygonTransaction.js
```diff
@@ -149,6 +155,38 @@ class SignPolygonTransaction {
             ]);
         }
 
+        if (request.description.name === 'swapWithApproval') {
+            const { sigR, sigS, sigV } = await polygonKey.signUsdcApproval(
+                request.keyPath,
+                new ethers.Contract(
+                    CONFIG.USDC_CONTRACT_ADDRESS,
+                    PolygonContractABIs.USDC_CONTRACT_ABI,
```
danimoh: Now would be a good time to rename those to `LEGACY_...`, see https://github.com/nimiq/keyguard/pull/484#discussion_r1442952587.
Same for `USDC_HTLC_CONTRACT_ADDRESS`.

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
```diff
@@ -106,19 +110,44 @@ class SignPolygonTransactionApi extends PolygonRequestParserMixin(TopLevelApi) {
...
+                value: forwardRequest.value,
+            }));
+
+            if (description.args.token !== CONFIG.USDC_CONTRACT_ADDRESS) {
+                throw new Errors.InvalidRequestError('Invalid USDC token contract in request data');
+            }
+
+            if (!['swap', 'swapWithApproval'].includes(description.name)) {
+                throw new Errors.InvalidRequestError('Requested Polygon contract method is invalid');
+            }
```
danimoh, line 131: Should check that `targetAmount` is about the same as `amount`, or display the value in the UI. Is it supposed to be `amount - fee`?

danimoh (changes requested):
Adding a check for the swap `targetAmount` would be a good idea.
The other comments are optional and not blocking for this PR, but would still be nice.

So for complete native USDC support, the mainnet swap contract would be missing, and then also a native USDC HTLC contract for atomic swaps?

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
sisou: The `targetAmount` is for making sure that the swap doesn't loose too much value due to slippage/unbalanced pool. If the result of the swap would be less than `targetAmount`, the swap is not executed (fees are still paid until then, though).

The `targetAmount` is likely 1% less (or even 0.5%, that's what the official UniSwap UI uses) than the input amount, as the conversion UniSwap pool is very stable.

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
```diff
@@ -127,25 +110,57 @@ class SignPolygonTransactionApi extends PolygonRequestParserMixin(TopLevelApi) {
...
+
+            // Ensure swap `targetAmount` is not too low
+            const inputAmount = /** @type {PolygonSwapDescription | PolygonSwapWithApprovalDescription} */ (description)
+                .args
+                .amount;
+            const targetAmount = /** @type {PolygonSwapDescription | PolygonSwapWithApprovalDescription} */ (description) // eslint-disable-line max-len
+                .args
+                .targetAmount;
```
danimoh, line 139: No need to cast here as `description` already has the desired type.
```suggestion
            const { inputAmount, targetAmount } = description.args;
```

danimoh (approved):

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
sisou: Does it for you? In my editor, it does not (the type annotation in line 119 doesn't actually work to narrow `description`).

sisou: What I could do is 
```js
const { inputAmount, targetAmount } = /** @type {PolygonSwapDescription | PolygonSwapWithApprovalDescription} */ (description).args;
```

danimoh: `yarn typecheck` has no complaints.

sisou: Yes, because the variables are happily assigned `any`. Which works, but is not helpful.

types/Keyguard.d.ts
```diff
@@ -152,6 +145,26 @@ type PolygonRefundDescription = ethers.utils.TransactionDescription & {
     readonly args: PolygonRefundArgs,
 };
 
+interface PolygonSwapArgs extends ReadonlyArray<any> {
```
mraveux, line 148: Is it intended for `PolygonSwapArgs` to extend an array to then have object props? Or am I missing/misunderstanding something?

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
mraveux: Yeah I have the same issue as Sisou, `parseTransaction` seems to return a `TransactionDescription` which is apparently not assignable to `description`'s type, thus giving me an error line 120.

mraveux (approved):
Nice & clean 👍

types/Keyguard.d.ts
sisou: That is indeed on purpose. The `args` can be accessed in two ways on the object returned from the `ethers` parser: via array access by index (`args[0]`) and also by name like an object (`args.token`), if the ABI has the argument name information, which ours does.

PR #490, opened 2024-02-26
Switch to native USDC HTLC for swaps

sisou:
Native USDC is represented as `USDC_MATIC` in Fastspot, so this is the symbol we use here, too.

Since we used `SwapAsset.USDC` in some places literally, e.g. for CSS classes or text, I introduced the `CryptoUtils.assetToCurrency()` method to convert between an asset and its currency ticker.


client/src/PublicRequest.ts
```diff
@@ -285,8 +285,8 @@ export type SignSwapRequestCommon = SimpleRequest & {
             refundKeyPath: string, // To validate that we own the HTLC script's refund address
         }>
     ) | (
-        {type: 'USDC'}
-        & Omit<PolygonTransactionInfo, 'amount'>
+        {type: 'USDC_MATIC'}
+        & Omit<PolygonTransactionInfo, 'approval' | 'amount'>
```
sisou: `approval` is still part of `PolygonTransactionInfo`, because it is used for swapping bridged USDC (which needs `approval`) to native USDC. It is not used for swapping, because that now uses `permit`.

client/src/PublicRequest.ts
```diff
@@ -320,8 +320,8 @@ export type SignSwapRequestCommon = SimpleRequest & {
             output: BitcoinTransactionChangeOutput,
         }
     ) | (
-        {type: 'USDC'}
-        & Omit<PolygonTransactionInfo, 'approval' | 'amount'>
+        {type: 'USDC_MATIC'}
+        & Omit<PolygonTransactionInfo, 'approval' | 'permit' | 'amount'>
```
sisou: For the redeem transaction, we need neither approval nor permit.

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
```diff
@@ -137,9 +137,11 @@ class SignPolygonTransactionApi extends PolygonRequestParserMixin(TopLevelApi) {
             const targetAmount = /** @type {PolygonSwapDescription | PolygonSwapWithApprovalDescription} */ (description) // eslint-disable-line max-len
                 .args
                 .targetAmount;
-            if (targetAmount.lt(inputAmount.mul(99).div(100))) {
+            // Allow 1% slippage for swaps on Polygon mainnet, but up to 5% for testnet
+            const minTargetAmountPercentile = CONFIG.POLYGON_CHAIN_ID === 137 ? 99 : 95;
+            if (targetAmount.lt(inputAmount.mul(minTargetAmountPercentile).div(100))) {
                 throw new Errors.InvalidRequestError(
-                    'Requested Polygon swap `targetAmount` more than 1% lower than the input `amount`',
+                    'Requested USDC swap `targetAmount` is too low',
```
sisou: This change technically belongs to #488 still.

client/src/PublicRequest.ts
danimoh: Could also be commented in the code. NIM and BTC above have such nice comments for every property that's omitted.

client/src/PublicRequest.ts
danimoh: Same as above putting the comment into the code would make sense.

src/lib/swap/CryptoUtils.js
```diff
@@ -8,31 +8,50 @@
...
+            case 'NIM':
+                return 'nim';
+            case 'BTC':
+                return 'btc';
+            case 'USDC_MATIC':
+                return 'usdc';
+            case 'EUR':
+                return 'eur';
+            default:
+                throw new Error('Invalid asset');
```
danimoh: Very nitpicking but could include the passed asset like the other methods above :smile: 
```suggestion
                throw new Error(`Invalid asset ${asset}`);
```

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
```diff
@@ -137,9 +137,11 @@ class SignPolygonTransactionApi extends PolygonRequestParserMixin(TopLevelApi) {
             const targetAmount = /** @type {PolygonSwapDescription | PolygonSwapWithApprovalDescription} */ (description) // eslint-disable-line max-len
                 .args
                 .targetAmount;
-            if (targetAmount.lt(inputAmount.mul(99).div(100))) {
+            // Allow 1% slippage for swaps on Polygon mainnet, but up to 5% for testnet
+            const minTargetAmountPercentile = CONFIG.POLYGON_CHAIN_ID === 137 ? 99 : 95;
+            if (targetAmount.lt(inputAmount.mul(minTargetAmountPercentile).div(100))) {
```
danimoh: Not really a percentile.
Maybe like this?
```suggestion
            const maxInputAmountSlippagePercentage = CONFIG.POLYGON_CHAIN_ID === 137 ? 1 : 5;
            if (targetAmount.lt(inputAmount.mul(100 - maxInputAmountSlippagePercentage).div(100))) {
```

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
```diff
@@ -95,10 +95,10 @@ class SignPolygonTransactionApi extends PolygonRequestParserMixin(TopLevelApi) {
             if (!['transfer', 'transferWithPermit'].includes(description.name)) {
                 throw new Errors.InvalidRequestError('Requested Polygon contract method is invalid');
             }
-        } else if (forwardRequest.to === CONFIG.BRIDGED_USDC_HTLC_CONTRACT_ADDRESS) {
+        } else if (forwardRequest.to === CONFIG.NATIVE_USDC_HTLC_CONTRACT_ADDRESS) {
             const usdcHtlcContract = new ethers.Contract(
-                CONFIG.BRIDGED_USDC_HTLC_CONTRACT_ADDRESS,
-                PolygonContractABIs.BRIDGED_USDC_HTLC_CONTRACT_ABI,
+                CONFIG.NATIVE_USDC_HTLC_CONTRACT_ADDRESS,
+                PolygonContractABIs.NATIVE_USDC_HTLC_CONTRACT_ABI,
```
danimoh: We should be aware that users can now not refund bridged USDC HTLCs anymore should they still have funds being stuck in such. But in that case fastspot claims the funds anyways and users can request it from there?

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
```diff
@@ -95,10 +95,10 @@ class SignPolygonTransactionApi extends PolygonRequestParserMixin(TopLevelApi) {
...
             );
 
             /** @type {PolygonRefundDescription} */
```
danimoh, line 104: Could potentially check here that the `target` address is the user's address.
Not sure whether it's checked somewhere.
The UI displays the address but also the key label (without checking that it's actually the correct address for that label). Thus users that only look at the label would miss an address mismatch.
Obviously users should check the address, but adding a check here would probably also not hurt.

danimoh (approved):
Approved with a few suggestions/comments.

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
sisou: Only HTLCs created by the user but not claimed by Fastspot can be refunded by the user. HTLCs that the user should have claimed but didn't, go back to Fastspot after 2 hours and need to be "refunded" from there.

So your comment is valid for the first case. No idea if these cases exist. I'll check if I can still allow to refund those.

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
sisou: we can only check the address after the password input, as only then it's derived.

I have added a check for this in the password confirm handler.

src/request/sign-polygon-transaction/SignPolygonTransactionApi.js
sisou: It was very easy, see latest force-push changeset: https://github.com/nimiq/keyguard/compare/583488fb61c901c403b6bbf12b01723d7b6a09f7..8dd1d00d530c5b385b0e37e4689942efc1fb0ff6

danimoh: Nice

mraveux (approved):

PR #491, opened 2024-02-27
Allow use of sign-polygon-transaction flow for re-signing USDC redeem transactions

sisou:
To enable resigning a USDC swap redeem transaction, we are going to re-use the regular sign-polygon-transaction flow. This PR extends the types and checks to allow that.

### DIFF
--- a/src/request/sign-polygon-transaction/SignPolygonTransaction.js
+++ b/src/request/sign-polygon-transaction/SignPolygonTransaction.js
@@ -32,7 +32,7 @@ class SignPolygonTransaction {
 
         /** @type {HTMLLinkElement} */
         const $sender = (this.$el.querySelector('.accounts .sender'));
-        if (request.description.name === 'refund') {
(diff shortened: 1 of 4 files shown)


src/request/sign-polygon-transaction/SignPolygonTransaction.js
```diff
@@ -42,20 +42,13 @@ class SignPolygonTransaction {
 
         /** @type {HTMLLinkElement} */
         const $recipient = (this.$el.querySelector('.accounts .recipient'));
-        if (request.description.name === 'refund') {
-            new PolygonAddressInfo(
-                /** @type {string} */ (request.description.args.target),
-                request.keyLabel,
-                'usdc',
-            ).renderTo($recipient);
+        const recipientAddress = /** @type {string} */ (request.description.args.target);
```
danimoh: Only used as `recipientAddress` in two of the three cases but ok.

danimoh (approved):

mraveux (approved):

PR #503, opened 2024-07-12
Fiat api refactor

danimoh:
- Switch back to the public CoinGecko API as we shut down our custom API endpoint. The public API is accessible from the keyguard domain with no issues and the usage limits are sufficient for the Keyguard's needs.
- Update the list of currencies supported by CoinGecko.
- Refactor FiatApi and translations mentioning the provider, to possibly support other providers in the future. Initially, I thought this would be more work, as I thought the implementation would be close to the one in `FiatApi` in `@nimiq/utils`, with dynamic providers, currency support depending on the chosen provider, type generics for the chosen provider etc. However, in the Keyguard, it can be quite a bit easier, as the provider can be considered a fixed value. Based on that value (which currently is only CoinGecko), I'm simply setting, for example, the appropriate list of supported currencies and the api endpoint, also as fixed values. Typescript seems to be happy with this. Maybe a more modern typescript version might complain, or if a second provider is actually added, but in that case all code and types for the inactive provider can simply be commented out to circumvent such issues.
- This PR is probably easier reviewed with whitespace changes ignored.

### DIFF
--- a/src/lib/FiatApi.js
+++ b/src/lib/FiatApi.js
@@ -48,9 +48,9 @@ FiatApi.SupportedCryptoCurrency = {
 /**
  * @readonly
  * @enum {'aed' | 'ars' | 'aud' | 'bdt' | 'bhd' | 'bmd' | 'brl' | 'cad' | 'chf' | 'clp' | 'cny' | 'czk' | 'dkk' | 'eur'
- *     | 'gbp' | 'hkd' | 'huf' | 'idr' | 'ils' | 'inr' | 'jpy' | 'krw' | 'kwd' | 'lkr' | 'mmk' | 'mxn' | 'myr' | 'nok'
- *     | 'ngn' | 'nzd' | 'php' | 'pkr' | 'pln' | 'rub' | 'sar' | 'sek' | 'sgd' | 'thb' | 'try' | 'twd' | 'uah' | 'usd'
- *     | 'vnd' | 'zar'}
+ *     | 'gbp' | 'gel' | 'hkd' | 'huf' | 'idr' | 'ils' | 'inr' | 'jpy' | 'krw' | 'kwd' | 'lkr' | 'mmk' | 'mxn' | 'myr'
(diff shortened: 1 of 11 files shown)


src/lib/FiatApi.js
```diff
@@ -11,109 +11,138 @@ class FiatApi {
...
+                    }), {});
+                    return {
+                        ...result,
+                        [cryptoCurrency]: sanitizedRecord,
+                    };
+                }, {});
+            }
+
+            default:
+                throw new Error(`FiatApi.Provider ${FiatApi.Provider} not supported yet.`);
```
sisou: This `default` should not be necessary. I think it might even be detrimental, as with it there won't be a TS compiler error when we add a new Provider, it'll only error at runtime.
My suggestion is to remove it.

src/lib/FiatApi.js
```diff
@@ -11,109 +11,138 @@ class FiatApi {
...
+     * @readonly
+     * Crypto currency tickers mapped to coingecko coin ids.
+     */
+    FiatApi.COINGECKO_COIN_IDS = {
+        [FiatApi.SupportedCryptoCurrency.NIM]: 'nimiq-2',
+        [FiatApi.SupportedCryptoCurrency.BTC]: 'bitcoin',
+    };
+} else {
+    throw new Error(`FiatApi.Provider ${FiatApi.Provider} not supported yet.`);
+}
```
sisou: Can we also convert this `if-else` statement into a switch statement without a `default` case, so TS catches missing providers?

mraveux (approved):
Simpler than I would've expected, but clean nonetheless.

src/lib/FiatApi.js
danimoh: Our TS version does in fact not warn if a provider is missing in the switch statement, but as a newer version might, I changed it.

sisou: See https://github.com/nimiq/keyguard/pull/467, which you are welcome to review ;)

sisou (approved):

PR #515, opened 2024-10-23
Support USDT

sisou:
USDT on Polygon is an older smart contract like bridged USDC was, thus I am re-introducing code to handle the EVM method `transferWithApproval`.


src/lib/polygon/PolygonContractABIs.full.js.txt
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-const PolygonContractABIsFull = { // eslint-disable-line no-unused-vars
+const PolygonContractABIsFull = { eslint-disable-line no-unused-vars
```
sisou: Accidentally removed

PR #523, opened 2025-01-13
Transaction signing fixes

sisou:
See the individual commit descriptions for change details.

### DIFF
--- a/src/request/sign-transaction/SignTransaction.js
+++ b/src/request/sign-transaction/SignTransaction.js
@@ -183,37 +183,18 @@ class SignTransaction {
         }
 
         const privateKey = key.derivePrivateKey(request.keyPath);
-        const keyPair = Nimiq.KeyPair.derive(privateKey);
+        const publicKey = Nimiq.PublicKey.derive(privateKey);
 
-        /** @type {Nimiq.Transaction} */
-        let tx;
+        const tx = request.transaction;
 
-        if (!request.transaction.data.length) {
-            tx = Nimiq.TransactionBuilder.newBasic(
-                keyPair.toAddress(),
-                Nimiq.Address.fromString(request.transaction.recipient.toHex()),
-                BigInt(request.transaction.value),
-                BigInt(request.transaction.fee),
-                request.transaction.validityStartHeight,
-                request.transaction.networkId,
-            );
-        } else {
-            tx = Nimiq.TransactionBuilder.newBasicWithData(
-                keyPair.toAddress(),
-                Nimiq.Address.fromString(request.transaction.recipient.toHex()),
-                request.transaction.data,
-                BigInt(request.transaction.value),
-                BigInt(request.transaction.fee),
-                request.transaction.validityStartHeight,
-                request.transaction.networkId,
-            );
-        }
-
-        tx.sign(keyPair);
+        // Manually create signature proof, because tx.sign(keyPair) does not support HTLC redeeming transactions.
+        const signature = Nimiq.Signature.create(privateKey, publicKey, tx.serializeContent());
+        const proof = Nimiq.SignatureProof.singleSig(publicKey, signature);
+        tx.proof = proof.serialize();
 
         /** @type {KeyguardRequest.SignTransactionResult} */
         const result = {
-            publicKey: keyPair.publicKey.serialize(),
+            publicKey: publicKey.serialize(),
             signature: tx.proof.subarray(tx.proof.length - 64),
             serializedTx: tx.serialize(),
         };
--- a/src/request/sign-transaction/SignTransaction.js
+++ b/src/request/sign-transaction/SignTransaction.js
@@ -182,21 +182,16 @@ class SignTransaction {
             return;
         }
 
-        const privateKey = key.derivePrivateKey(request.keyPath);
-        const publicKey = Nimiq.PublicKey.derive(privateKey);
+        const publicKey = key.derivePublicKey(request.keyPath);
+        const signature = key.sign(request.keyPath, request.transaction.serializeContent());
 
-        const tx = request.transaction;
-
-        // Manually create signature proof, because tx.sign(keyPair) does not support HTLC redeeming transactions.
-        const signature = Nimiq.Signature.create(privateKey, publicKey, tx.serializeContent());
-        const proof = Nimiq.SignatureProof.singleSig(publicKey, signature);
-        tx.proof = proof.serialize();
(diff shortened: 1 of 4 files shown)


src/request/sign-transaction/SignTransaction.js
```diff
@@ -183,37 +183,18 @@ class SignTransaction {
...
+        // Manually create signature proof, because tx.sign(keyPair) does not support HTLC redeeming transactions.
+        const signature = Nimiq.Signature.create(privateKey, publicKey, tx.serializeContent());
+        const proof = Nimiq.SignatureProof.singleSig(publicKey, signature);
+        tx.proof = proof.serialize();
 
         /** @type {KeyguardRequest.SignTransactionResult} */
         const result = {
-            publicKey: keyPair.publicKey.serialize(),
+            publicKey: publicKey.serialize(),
             signature: tx.proof.subarray(tx.proof.length - 64),
```
danimoh: Can return `signature.serialize()` instead of extracting it from the proof.

src/request/sign-transaction/SignTransaction.js
```diff
@@ -183,37 +183,18 @@ class SignTransaction {
...
-                request.transaction.networkId,
-            );
-        }
+        const publicKey = Nimiq.PublicKey.derive(privateKey);
+
+        const tx = request.transaction;
 
-        tx.sign(keyPair);
+        // Manually create signature proof, because tx.sign(keyPair) does not support HTLC redeeming transactions.
+        const signature = Nimiq.Signature.create(privateKey, publicKey, tx.serializeContent());
```
danimoh: Alternatively, there's also `key.sign(path, data)`. But it's pretty much the same.

src/request/sign-transaction/SignTransaction.js
```diff
@@ -183,37 +183,18 @@ class SignTransaction {
...
+        const proof = Nimiq.SignatureProof.singleSig(publicKey, signature);
+        tx.proof = proof.serialize();
```
danimoh: A single sig proof is a sensible default, but one could also argue that the Keyguard should not even make any assumptions about what proof to create and only return the signature. That was also the case before bb576462.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -445,8 +445,9 @@ class SwapIFrameApi extends BitcoinRequestParserMixin(RequestParser) { // eslint
                     transaction.recipient, Nimiq.AccountType.HTLC, new Uint8Array(0),
                     transaction.sender, Nimiq.AccountType.Basic, new Uint8Array(0),
                     transaction.value - fee, fee,
```
danimoh, line 457: The fee calculation needs to be updated as transaction sizes changed with PoS.

Additionally, enforcing 0 as minimum value for the transaction amount would also make sense.

src/request/swap-iframe/SwapIFrameApi.js
```diff
@@ -445,8 +445,9 @@ class SwapIFrameApi extends BitcoinRequestParserMixin(RequestParser) { // eslint
...
-                    0 /* Nimiq.Transaction.Flag.NONE */,
-                    parsedRequest.fund.htlcDetails.timeoutTimestamp,
+                    Nimiq.TransactionFlag.None,
+                    transaction.validityStartHeight
+                        + (parsedRequest.fund.htlcDetails.timeoutTimestamp - Math.round(Date.now() / 1e3)),
```
danimoh: Could add a comment here with a similar explanation as in the commit message. Additionally, the calculated vsh can be wrong, if the user's system clock is off. The comment could also state, that an incorrect vsh that renders the refund transaction unusable is not much of an issue actually, as the user can manually refund then.

Enforcing `transaction.validityStartHeight` as the minimum allowed value would also make sense.

danimoh (approved):
Makes sense generally, and approved. Nonetheless, there are some smaller issues which should be addressed.

src/request/sign-transaction/SignTransaction.js
sisou: You are right. But the `serializedTx` return field is now used in Hub and probably Wallet, so removing that again will be a separate task.

src/request/swap-iframe/SwapIFrameApi.js
sisou: 167 bytes for refund txs is actually accurate. No idea if I updated this at some point, or it didn't change :shrug:

sisou: As an example see a refund tx from Fastspot: `098f74db2d247e05b3175933dbcd6027e21e36f58e06e94957f31ad14696bfd2`

```js
{
  hash: '098f74db2d247e05b3175933dbcd6027e21e36f58e06e94957f31ad14696bfd2',
  blockNumber: 6815039,
  timestamp: 1735439672089,
  confirmations: 1448037,
  size: 167,
  relatedAddresses: [
    'NQ79 JPT1 FFTJ U9TD HNRU N07T 4FGR 0PJJ VJ1D',
    'NQ56 VE9K PV0Y TK4T QBMY 8E9G B85J QR6M MSHL'
  ],
  from: 'NQ79 JPT1 FFTJ U9TD HNRU N07T 4FGR 0PJJ VJ1D',
  fromType: 2,
  to: 'NQ56 VE9K PV0Y TK4T QBMY 8E9G B85J QR6M MSHL',
  toType: 0,
  value: 8301585962,
  fee: 0,
  senderData: '',
  recipientData: '',
  flags: 0,
  validityStartHeight: 6815037,
  proof: '020008600ec9f0d44dc8d43275c705d7780caa31497d2620da4d7838d10574a6dfa100de3165b68453935067076f726018c60aaaa5ca0a128d0d0dc9882d309c63b359b2a229ca3b95b648c8abf97cf0d9fdba60440c7de57c192b9f5a23a5e59b4800',
  networkId: 24,
  executionResult: true
}
```

sisou:
Thanks @danimoh, I addressed your comments.

mraveux (approved):
Looks good to me 👍 

I was wondering at first why you replaced the `contractAddress` by `new Nimiq.Address(new Uint8Array(20))` in [SwapIFrameApi.js](https://github.com/nimiq/keyguard/pull/523/files#diff-0b688f04fa7bc83c3e63e37bfce2b39d660dc3112d0582d5c7de6212390ab45f) but @danimoh explained it to me and we went to check it in the web client, and it looks good 😁

sisou:
Rebased for merging

PR #524, opened 2025-03-17
Fixes for PR #458 (Multisig: Connect request and SignMultisigTransaction request)

danimoh:
Addresses most of the comments on PR #458.


src/request/sign-multisig-transaction/SignMultisigTransaction.js
```diff
@@ -229,50 +224,44 @@ class SignMultisigTransaction {
...
+                ownSigner.commitments[i],
+            ));
+        }
 
         const signature = key.signPartially(
             request.keyPath,
             request.transaction.serializeContent(),
             ownCommitmentPairs,
-            request.multisigConfig.signers.filter(signer => !signer.publicKey.equals(publicKey)),
+            request.multisigConfig.signers.filter(signer => signer !== ownSigner),
```
sisou: This may be a problem, because `Nimiq.KeyPair`s cannot be compared as objects, as they are references into Rust WASM memory. So the same KeyPair can be represented as different objects, with different pointers, altough they represent the same bytes.

I think the previous comparison was more accurate.

src/lib/TransactionDataFormatting.js
```diff
@@ -0,0 +1,122 @@
...
+
+        // Contract withdrawal without any data
+        // If a transaction has no transaction data but a special sender type, try to give it a useful label nonetheless
+        if (transaction.senderData.length === 0
+            && transaction.data.length === 0
+            && transaction.senderType !== Nimiq.AccountType.Basic) {
+            return I18n.translatePhrase('tx-data-contract-withdrawal', {
+                contractType: TransactionDataFormatting._capitalize(plainTransaction.senderType),
+            });
+        }
```
sisou: Vesting withdrawal transactions can have a UTF8 message as the only data. How is that represented by this? I think only as hex, which is not useful.

src/lib/TransactionDataFormatting.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * @private
+     * @param {Nimiq.PlainTransactionRecipientData} plainData
+     * @param {boolean} includeDataType
+     * @returns {string}
+     */
+    static _formatPlainData(plainData, includeDataType) {
+        if (plainData.type === 'raw') return plainData.raw; // Return plain hex. Empty if transaction has no data.
```
sisou, line 77: I think this should attempt to decode as UTF8 and only return the raw HEX when decoding fails.

src/lib/TransactionDataFormatting.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        // Staking transaction
+        if (transaction.senderType === Nimiq.AccountType.Staking) {
+            // A staking withdrawal transaction. Those are never signaling transactions.
+            prefix = I18n.translatePhrase('tx-data-contract-withdrawal', { contractType: 'Staking' });
+        } else if (transaction.recipientType === Nimiq.AccountType.Staking) {
+            if (!(transaction.flags & Nimiq.TransactionFlag.Signaling)) { // eslint-disable-line no-bitwise
+                // A staking transaction which creates/adds actual stake by moving funds to the staking contract
+                prefix = I18n.translatePhrase('tx-data-contract-creation', { contractType: 'Staking' });
```
sisou: This will translate to "Staking creation", even when users just _add_ stake to their existing stake. That is misleading.

I think the staking contract should have a separate translation, because the staking contract is never _created_, it already exists since the PoS genesis block.

src/request/sign-transaction/SignTransactionApi.js
```diff
@@ -57,6 +58,11 @@ class SignTransactionApi extends TopLevelApi {
             parsedRequest.cashlinkMessage = /** @type {string} */(this.parseMessage(request.cashlinkMessage));
         }
 
+        if (parsedRequest.transaction.senderType === Nimiq.AccountType.Staking
+            || parsedRequest.transaction.recipientType === Nimiq.AccountType.Staking) {
+            throw new Errors.InvalidRequestError('For staking transactions, use a Keyguard sign-staking request');
```
sisou: Reword: `'For staking transactions, use the Keyguard request "sign-staking"'`

src/components/AddressInfo.css
```diff
@@ -49,7 +49,7 @@
     padding: 0.375rem 0.75rem;
     line-height: 1;
     border: 0.5rem solid white;
-    word-spacing: -0.071428571em; // 1/14
+    word-spacing: calc(-1em / 14);
```
sisou, line 52: The comment was not invalid.

src/lib/rsa/sandboxed/RSAKeysIframe.html
```diff
@@ -3,7 +3,8 @@
 <head>
     <meta charset="UTF-8">
     <!-- Note: sandbox and frame-ancestors policies can not actually be specified via a meta tag. You have to send them
-    as response headers. -->
+    as response headers. Additionally, the server needs to include the sandboxed iframe's null origin in its cors header
+    for the script files imported here, in order to pass integrity checks added by the build process. -->
```
sisou, line 7: What does that mean? What must the server send now? Can you list that here so we can tell IT?

tools/build.sh
```diff
@@ -360,13 +360,31 @@ awk '
 # make redirect file available at /request/ too
 cp dist/index.html dist/request
 
+# Build RSA iframe
+output "🔑  Building RSA Iframe"
+cp src/lib/rsa/sandboxed/forge.min.js dist/lib/rsa/sandboxed/forge.min.HASH.js
+# Integirty check that forge.min.js is from node-forge 1.3.1
```
sisou: Spelling: "Integrity"

tools/build.sh
```diff
@@ -360,13 +360,31 @@ awk '
...
+RSA_IFRAME_FORGE_HASH="3Gf9EyQnrZbJZmyESzlWVBPEDdsfLQY8U1Evv204ff0="
```
sisou: Why hard-code the hash here but not verify it? That just means it will fail at runtime if we update the script, instead of at build time. 

Either:
1. Calculate the hash from the script that is copied to the build folder
2. Verify this hash against the file it is supposed to represent

tools/build.sh
```diff
@@ -360,13 +360,31 @@ awk '
...
+RSA_IFRAME_FORGE_NAME=$(add_hash_to_file_name dist/lib/rsa/sandboxed/forge.min.HASH.js)
+cp src/lib/rsa/sandboxed/RSAKeysIframe.js dist/lib/rsa/sandboxed/RSAKeysIframe.HASH.js
+RSA_IFRAME_SCRIPT_HASH=$(make_file_hash dist/lib/rsa/sandboxed/RSAKeysIframe.HASH.js)
+RSA_IFRAME_SCRIPT_NAME=$(add_hash_to_file_name dist/lib/rsa/sandboxed/RSAKeysIframe.HASH.js)
+cp src/lib/rsa/sandboxed/RSAKeysIframe.html dist/lib/rsa/sandboxed/
+# Note: requests in a sandboxed iframe are considered cross origin requests, for which, in addition to integrity checks,
+# also cors headers must be checked, see https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity.
```
sisou, line 378: What CORS headers must be sent with what values?

src/request/sign-multisig-transaction/SignMultisigTransaction.js
```diff
@@ -263,6 +263,10 @@ class SignMultisigTransaction {
                 ));
             }
         }
+        if (ownCommitmentPairs.length !== ownSigner.commitments.length) {
+            reject(new Errors.InvalidRequestError('The number of secrets does not match the number of my commitments'));
```
sisou: Reword: 'The number of secrets does not match the number of this signer\'s commitments'

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -141,10 +141,11 @@ class SignMultisigTransactionApi extends TopLevelApi {
             }
         } else if (object.secrets && typeof object.secrets === 'object'
             && 'encrypted' in object.secrets && 'keyParams' in object.secrets) {
-            // Not checking fixed length here, to stay flexible for future increases of the number of commitments
-            if (!Array.isArray(object.secrets.encrypted) || object.secrets.encrypted.length < 2) {
+            if (!Array.isArray(object.secrets.encrypted)) {
+                // No need to check object.secrets.encrypted.length here, as the number will later be checked against
+                // the number of my provided commitments, after we identify them in SignMultisigTransaction.
```
sisou: Reword: the number of this signer's provided commitments

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
```diff
@@ -141,10 +141,11 @@ class SignMultisigTransactionApi extends TopLevelApi {
             }
         } else if (object.secrets && typeof object.secrets === 'object'
             && 'encrypted' in object.secrets && 'keyParams' in object.secrets) {
-            // Not checking fixed length here, to stay flexible for future increases of the number of commitments
-            if (!Array.isArray(object.secrets.encrypted) || object.secrets.encrypted.length < 2) {
+            if (!Array.isArray(object.secrets.encrypted)) {
+                // No need to check object.secrets.encrypted.length here, as the number will later be checked against
```
sisou, line 147: Although failing here would fail in the beginning of the request, instead of _after_ password entry.

src/lib/Key.js
```diff
@@ -30,11 +30,11 @@ class Key {
...
     /**
      * @private
-     * @param {EncryptionKeyParams} paramsA
-     * @param {EncryptionKeyParams} paramsB
+     * @param {RsaKeyParams} paramsA
+     * @param {RsaKeyParams} paramsB
      * @returns {boolean}
      */
-    static _areEncryptionKeyParamsEqual(paramsA, paramsB) {
+    static _areEqualRsaKeyParams(paramsA, paramsB) {
```
sisou: Suggestion for reword: `_areRsaKeyParamsEqual`

src/lib/Key.js
```diff
@@ -43,11 +43,11 @@ class Key {
 
     /**
      * @private
-     * @param {EncryptionKeyParams} params
+     * @param {RsaKeyParams} params
      * @returns {boolean}
      */
-    static _areDefaultEncryptionKeyParams(params) {
-        return Key._areEncryptionKeyParamsEqual(params, Key.defaultEncryptionKeyParams);
+    static _areDefaultRsaKeyParams(params) {
```
sisou: Suggestion for reword: `_areRsaKeyParamsDefault`

src/lib/TransactionDataFormatting.js
sisou: With my changes to the `formatTransactionData` function in my branch this is now no longer necessary, as this method is only called when the recipient is a contract or the data is not UTF8 anyway, so UTF8 decoding is not relevant.

client/src/PublicRequest.ts
```diff
@@ -530,9 +530,9 @@ export type DerivePolygonAddressResult = {
 
 export type ConnectRequest = SimpleRequest & {
     appLogoUrl: string,
-    permissions: KeyguardCommand[],
     requestedKeyPaths: string[],
     challenge: string,
+    // permissions: KeyguardCommand[],
```
sisou, line 547: Permissions are stored and verified in the Hub.

src/request/sign-multisig-transaction/SignMultisigTransaction.js
danimoh: The `ownSigner` is assigned one of the objects from `request.multisigConfig.signers`, i.e. the strict equality check by reference works as expected.

sisou: Yeah, I noticed that later and changed it back to your version, just in a different place and with an explainer in my latest PR. Sorry, forgot to comment here.

src/lib/TransactionDataFormatting.js
danimoh: Yes, hex, I believe. Decoding the Utf8 message makes sense then for this case :+1:.
It's not only vesting withdrawals though, HTLC withdrawals can have additional data as well, apart from the required redeeming information, which is in the signature proof. For `RemoveStake` and `DeleteValidator` transactions, I think additional recipient data might also be allowed, apart from the required sender data, but I have not confirmed.

Your proposed changes handle this correctly though.

src/lib/TransactionDataFormatting.js
danimoh: Yes, I like interpreting data as UTF8 if the recipient is a basic account. I'm not a fan of trying to interpret other data as UTF8. So I like your latest approach.

src/lib/TransactionDataFormatting.js
danimoh: Yeah, I was just re-using the translation, but a separate, more specific translation might make sense.

src/lib/rsa/sandboxed/RSAKeysIframe.html
danimoh: - [`sandbox` and `frame-ancestors` headers](https://github.com/nimiq/keyguard/pull/458#discussion_r1939751795)
- The `forge.min.js` and `RSAKeysIframe.js` (or rather the deployed versions with filename hashes) need to be send with `Access-Control-Allow-Origin: null`. Mdn has a [warning regarding doing this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin#null), but as far as I can see, there is no actual damage in doing so in this case.

CC @curdbecker

tools/build.sh
danimoh: - Checking it at runtime is the important check for detecting files tampered with, checking it also at build time adds little to that.
- Having a hard coded hash in the code means that it needs to be updated manually in the code, which will show up in pull requests on changes, and serves as a nice reminder on code review to cross-check that file hash with the `forge.min.js` obtained from the original source. If the hash is calculated automatically, a reviewer has to not forget himself to cross check the file `forge.min.js` against the original source, if it's changed in the keyguard code base.

tools/build.sh
danimoh: https://github.com/nimiq/keyguard/pull/524#discussion_r2182339196

src/request/sign-multisig-transaction/SignMultisigTransactionApi.js
danimoh: See https://github.com/nimiq/keyguard/pull/458#discussion_r2181074120

PR #526, opened 2025-06-02
Multisig: Use AES for RSA key encryption

danimoh:
An alternative approach to encrypting the RSA private key on storage, which is used in the multisig feature.
This alternative approach has a couple of advantages to the [one previously taken](https://github.com/nimiq/keyguard/pull/524/commits/8213c40b2616e9e58a5d5a22a24d271a83ded576):
- it uses AES encryption via browser native crypto APIs
- RSA keys are represented as native `CryptoKey`s at runtime
- the implementation is more versatile and the AES encryption can easily be used for other use cases in the future.

### DIFF
--- a/src/lib/Key.js
+++ b/src/lib/Key.js
@@ -6,7 +6,7 @@
 /* global Utf8Tools */
 
 /**
- * @typedef {{hasPin?: boolean, rsaKeyPair?: RsaKeyPairExport}} KeyConfig
+ * @typedef {{hasPin?: boolean, rsaKeyPair?: RsaKeyPair}} KeyConfig
  */
 
 class Key {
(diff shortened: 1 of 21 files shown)


src/lib/Key.js
```diff
@@ -185,13 +251,7 @@ class Key {
      */
     async getRsaPrivateKey(keyParams) {
         const rsaKeyPair = await this.setRsaKeyPair(keyParams);
```
sisou: I don't like that a method that is called `set*` is used to _get_ the key.

sisou (approved):

src/lib/Key.js
danimoh: Calculating and setting/caching the RSA key is actually the main functionality of that method but I see where you're coming from.
I addressed it in https://github.com/nimiq/keyguard/pull/526/commits/e2e4af41c3e907bf23a7803577195f1d2de744e2.

PR #532, opened 2025-07-01
Multisig-fixes Fixes

sisou:
Some changes for the review I made on #524:

- Change transaction data formatting logic to always show a message when the tx is to a BASIC account (and also the contract withdrawal prefix if it is a withdrawal)
- Minor reword of some error messages, comments and method names
- Verify `forge.min.js` file integrity during build, not only at runtime
- Move filtering of other signers to right behind the search for the own signer, to make it clearer that object comparison can be used there.

### DIFF
--- a/demos/ConnectAccount.html
+++ b/demos/ConnectAccount.html
@@ -35,14 +35,14 @@
 
     <div class="row">
         <label>Challenge</label>
-		<input id="challenge" value="some random characters">
+        <input id="challenge" value="some random characters">
     </div>
 
     <div class="row">
         <label>Permissions</label>
-		<label>
-			<input type="checkbox" id="permission-sign-multisig-transaction" checked> sign-multisig-transaction
-		</label>
+        <label>
+            <input type="checkbox" id="permission-sign-multisig-transaction" checked> sign-multisig-transaction
+        </label>
     </div>
 
     <!-- <br><button id="redirect" type="button" class="small" disabled>Connect Account (redirect)</button> -->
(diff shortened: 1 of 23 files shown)


demos/ConnectAccount.html
```diff
@@ -52,6 +52,11 @@
 </form>
 
 <script>
+    window.addEventListener('load', async () => {
+        await Nimiq.default();
+	    loadAccounts();
```
danimoh: Indentation is off.
Not that it matters, though :smiley: 
You don't have to change it.

demos/ConnectAccount.html
```diff
@@ -52,6 +52,11 @@
 </form>
 
 <script>
+    window.addEventListener('load', async () => {
```
danimoh: I believe, if this script itself would also be `type="module"`, it wouldn't be required to run in a `load` event listener.

src/lib/TransactionDataFormatting.js
```diff
@@ -19,29 +19,29 @@ class TransactionDataFormatting { // eslint-disable-line no-unused-vars
...
+
+        // Contract withdrawal
+        if (transaction.senderType !== Nimiq.AccountType.Basic) {
+            prefix = I18n.translatePhrase('tx-data-contract-withdrawal', {
                 contractType: TransactionDataFormatting._capitalize(plainTransaction.senderType),
             });
         }
 
-        let prefix = '';
         let includeDataType = true;
```
danimoh: I'd also move the definition of `includeDataType` up to where `prefix` is now defined.

src/lib/TransactionDataFormatting.js
```diff
@@ -19,29 +19,29 @@ class TransactionDataFormatting { // eslint-disable-line no-unused-vars
...
-            && transaction.senderType !== Nimiq.AccountType.Basic) {
-            return I18n.translatePhrase('tx-data-contract-withdrawal', {
+        let prefix = '';
+
+        // Contract withdrawal
+        if (transaction.senderType !== Nimiq.AccountType.Basic) {
+            prefix = I18n.translatePhrase('tx-data-contract-withdrawal', {
                 contractType: TransactionDataFormatting._capitalize(plainTransaction.senderType),
             });
         }
```
danimoh: With this change, these lines below can be removed:

```js
        if (transaction.senderType === Nimiq.AccountType.Staking) {
            // A staking withdrawal transaction. Those are never signaling transactions.
            prefix = I18n.translatePhrase('tx-data-contract-withdrawal', { contractType: 'Staking' });
        } else
```

danimoh (approved):

demos/ConnectAccount.html
sisou: It's because it's a mix of tabs and spaces and tabs are rendered as 8 spaces by Github by default.

demos/ConnectAccount.html
sisou: And you are right! :muscle:

danimoh (approved):

PR #542, opened 2026-01-07
Backup Codes

danimoh:
This PR contains the Backup Codes feature, which serves as an alternative backup option to the Login File and Recovery Words.

So far, I pushed the export flow, which is ready for review. The import flow will follow soon after in the coming days.
Before final release, the link to the instructions will have to be updated with the link to the actual blog post.

@paberr, from you a review of the technical parts should suffice. Please have a look at the changes in  `src/lib/BackupCodes.js` and `src/lib/Key.js`.


src/request/export/index.html
```diff
@@ -319,6 +324,172 @@ <h1 class="nq-h1 confirm-download " data-i18n="download-loginfile-successful">Do
...
+                <div class="page-body nq-card-body backup-codes-illustration"></div>
+
+                <div class="page-footer nq-card-footer">
+                    <p class="nq-notice show-loading" data-i18n="export-backup-codes-send-code-loading">
+                        Generating codes...
+                    </p>
+                    <button data-i18n="export-backup-codes-send-code-copy" class="nq-button light-blue hide-loading">
+                        Copy code {n}
+                    </button>
+                    <a href="#TODO" target="_blank" rel="noopener" class="skip nq-link hide-loading">
```
danimoh: TODO: replace with the link to the actual blog post when ready.

src/lib/BackupCodes.js
```diff
@@ -0,0 +1,106 @@
...
+        // basically derived from the secret and metadata by hashing.
+        const plainText = new Nimiq.SerialBuffer(/* versionAndFlags */ 1 + secretBytes.length);
+        plainText.writeUint8(versionAndFlags);
+        plainText.write(secretBytes);
+
+        // We generate one code from the key via a key derivation function (kdf), which is then applied to the plaintext
+        // as a one-time-pad (otp) to yield the second code. We use Argon2id as kdf (as opposed to, for example, pbkdf2)
+        // due to its memory-hardness.
+        const derivationUseCase = `BackupCodes - ${versionAndFlags}`; // include metadata in derivation / checksum
+        const code1Bytes = await key.deriveSecret(derivationUseCase, 'Argon2id', 8, plainText.length);
```
danimoh: @paberr what do you think, which kdf we should use and what parameters? I chose rather cautious options, but in favor of shorter computation time, less expensive options could also suffice, imo. However, sticking to the cautious  options is also possible. The UI includes appropriate loading states.

paberr: Since the seed is already high-entropy data, I would go for a more standard KDF such as HKDF.

src/lib/BackupCodes.js
```diff
@@ -0,0 +1,106 @@
...
+        if (version > BackupCodes.VERSION) throw new Error('Invalid Backup Codes: unsupported version');
+        const isLegacyPrivateKey = !!(flags & BackupCodes.FLAG_IS_LEGACY_PRIVATE_KEY);
+
+        const secretBytes = plainText.slice(/* versionAndFlags */ 1);
+        const secret = isLegacyPrivateKey ? new Nimiq.PrivateKey(secretBytes) : new Nimiq.Entropy(secretBytes);
+        const key = new Key(secret);
+
+        // Check whether our recovered key would derive the same codes, as a checksum check.
+        const [checksum1, checksum2] = await BackupCodes.generate(key);
+        if (checksum1 !== code1 || checksum2 !== code2) throw new Error('Invalid Backup Codes: checksum mismatch');
```
paberr: If we care about side-channel leakage: These are string comparisons, which are not constant time.
An attacker could learn timings. Doing it on byte arrays would mitigate this, but not sure if it's a realistic threat.

paberr (commented):
Also, I would explicitly state that it is ok for us to leak equality (because it's deterministic).
- If an attacker ever sees a backup code pair (or even one of them), they can test whether another pair belongs to the same key by recomputing/relating (depending on what they have). It is strictly more information than in a random case.
- If the user stores codes in multiple places, compromise of one place can enable correlation across other leaks.

src/lib/BackupCodes.js
danimoh: As you're mentioning that by the nature of our unique, deterministic codes, one code can be unambiguously matched to the other, do you see some value in making this matching more expensive, by choosing an expensive kdf? This way, an attacker who manages to get access to a large set of backup codes, for example via database leaks, has a harder time matching backup codes and reconstructing keys. However, as we're treating with crypto keys here, I guess for an attacker with access to backup codes, it would always be worth attempting to reconstruct keys, regardless of a more expensive kdf.

src/lib/BackupCodes.js
danimoh: Yeah, not sure how this could be leveraged here, but as changing it to a constant time comparison is easy, I'll do that.

danimoh (commented):
Thanks for the review. Do you think the issues with code equality and code correlation are worth switching to truly random codes? I do like the elegance of deterministic codes and the fact that no separate checksum needs to be included.

src/lib/BackupCodes.js
paberr: There are scenarios where it adds minimal value, but I don't think it's creating any actual security. So, I think it's better to keep it simple as the case is outside our threat model anyway.

paberr:
> Thanks for the review. Do you think the issues with code equality and code correlation are worth switching to truly random codes? I do like the elegance of deterministic codes and the fact that no separate checksum needs to be included.

Adding a checksum has a similar effect: you can check if two codes match with high likelihood. So, as long as we want to have some sort of check, it's not worth it changing this.

danimoh:
> Adding a checksum has a similar effect: you can check if two codes match with high likelihood. So, as long as we want to have some sort of check, it's not worth it changing this.

Ah right, of course :smile:
Multiple different exports of the same key can not be related to each other anymore with random codes, but a matching code can still be identified with high likelihood. I wouldn't omit having some sort of checksum, though, and I think the benefits of the deterministic codes outweigh the small drawback regarding code equality.

mraveux:
It looks good to me 👍 
I couldn't find anything I was unhappy about, and it worked pretty well on the browsers I tested it on 👍

mraveux (approved):

src/lib/Key.js
```diff
@@ -291,18 +344,45 @@ class Key {
...
+            name: 'HKDF',
+            // Use SHA-512 for cheaper derivation of longer derivedSecretLengths and better quantum resistance.
+            hash: 'SHA-512',
+            salt,
+            info: Utf8Tools.stringToUtf8ByteArray([
+                useCase,
+                // Derive different secrets for legacy PrivateKey based accounts and modern Entropy based accounts, even
+                // if their underlying secret bytes are the same.
+                this.secret instanceof Nimiq.PrivateKey ? 'PrivateKey' : 'Entropy',
+            ].join()),
```
paberr, line 355: It's important to make sure a user can't create collisions of the info string. Since all use-case strings are hardcoded (not user-controlled), I don't see a practical risk for collision. But as a defensive measure, using an unambiguous separator (e.g. a length-prefixed encoding) would make this robust by construction rather than by convention.

paberr (approved):
Generally seems good now.

PR #552, opened 2026-08-18
Add Github Actions workflow to deploy to testnet

jeffesquivels:
Deploys the testnet Keyguard to an S3-backed CloudFront distribution from CI, next to the existing
`deploy.sh --testnet` SSH path. It publishes to a separate hostname (`vars.DEPLOY_URL`), so both
paths can run side by side until DNS for keyguard.nimiq-testnet.com is cut over.

## Trigger

Publishing a **pre-release** deploys, and nothing else does. A pushed tag on its own never deploys:
`deploy.sh` pushes two tags per run (`v$VERSION` and `v$VERSION-$ENV_TAG-$DEPLOYER`) and tags
mainnet deployments the same way, so a tag trigger would deploy twice over, and for the wrong
network. Full releases stay free for the mainnet workflow to claim later via `types: [released]`.

There is no manual trigger. An earlier revision had `workflow_dispatch` with an opt-out for each of
the gates below, and a deploy path that can be told to skip its own checks is the deploy path an
attacker uses; both inputs and the trigger are gone. Every run is a published pre-release, and every
gate is fatal.

## What the run does

**Gates, before anything is installed or built**

- every deployment variable must be set — a missing one stops the run rather than silently
  skipping the step that needs it
- `verify-tag.js`: the release's tag must be annotated and carry a signature GitHub verified,
  optionally by someone in `vars.ALLOWED_TAGGERS`. Runs on the runner's preinstalled Node before
  `yarn install`, so nothing from the ref has been executed while the ref is still being judged
- `verify-checks.js`: the commit must have passed `Complete Check` — concluded `success`, and
  nothing else: `skipped` and `neutral` say the tests reached no verdict. The deploy itself runs no
  test, and a release can be published from any commit, so this is where that evidence comes from. A
  check still in flight is waited out rather than failed, and a check-runs request that fails is
  retried on the next poll instead of ending the deploy

**Build**

- `yarn install --frozen-lockfile --ignore-scripts`, then `yarn build testnet`
- `yarn checkdist testnet` (new, `tools/distValidator.js`): asserts dist/ is complete, that every
  build rewrite was applied, that the bundle carries *this* network's config and no other, and that
  every file is covered by an upload rule. Also added to `yarn pr`, so it runs on every PR
- writes `dist/build-info.json` (ref, sha, config, run URL, timestamp)

**Publish**

- OIDC credentials, requested only after the build and only for 15 minutes
- `s3-sync.sh`: one upload pass per content type with an explicit `content-type`, HTML last since it
  carries the SRI hashes pinning everything else. A dryrun then proves the typed passes were
  exhaustive, reports what the prune will delete, and refuses a bucket that does not look like a
  Keyguard deployment or a prune larger than `MAX_PRUNE`
- CloudFront invalidation, waited to completion
- `tools/deploymentValidator.js` smoke test: security headers on all four `frame-ancestors`
  variants, and every SRI hash re-derived from the bytes the CDN serves

## Also usable outside CI

- `yarn checkdist <testnet|mainnet>` after a build
- `yarn checkdeployment <url>` against any origin, including today's nginx one

## Setup this expects, none of which lives in the repo

- Environment `testnet` with a required reviewer, holding `AWS_REGION`, `S3_BUCKET`,
  `CLOUDFRONT_DISTRIBUTION_ID`, `DEPLOY_URL`, optional `ALLOWED_TAGGERS`, and
  `secrets.AWS_DEPLOY_ROLE_ARN` for a role whose trust policy is scoped to the OIDC subject
  `repo:nimiq@33815469/keyguard@139560932:environment:testnet`
- A bucket used by nothing else — the sync prunes with `--delete` at the bucket root
- **An index rewrite for directory paths.** With an S3 REST origin, CloudFront resolves
  `DefaultRootObject` only at `/`, so `/request/create/` 404s unless the distribution rewrites
  directory paths to `index.html` in a CloudFront Function or fronts the S3 *website* endpoint.
  This is the most likely first failure; the smoke test catches it, but only after a full sync
- Response-headers policies reproducing the nginx headers, including the per-path `frame-ancestors`.
  `tools/deploymentValidator.js` is the specification for them

## Verification

- `tools/deploymentValidator.js https://keyguard.nimiq-testnet.com` passes against the live nginx
  origin on every check, with 8 SRI hashes re-derived; only `/build-info.json` 404s there, because
  only this workflow writes it
- `tools/distValidator.js` correctly rejects a mainnet build asked to validate as testnet, and the
  s3-sync include patterns were shown to cover every file in a real dist/ exactly once
- `verify-tag.js`, `verify-checks.js` and `s3-sync.sh` were exercised against a stub GitHub API and
  a stub `aws` CLI: unsigned and lightweight tags, a tag name containing a slash, a missing tag,
  taggers outside the allowlist, failing, skipped and in-flight checks, a re-run whose newest
  attempt is the one that counts, a run of API failures, an uncovered file type, a foreign bucket,
  and an oversized prune

## Reviewing

The first commit adds the workflow. Each commit after it addresses one review finding on its own,
so they can be read, questioned or dropped individually.

## Left for later, on purpose

- The content-hashed bundles under `request/` are still served `no-cache, must-revalidate` to match
  nginx exactly. Switching them to `immutable` is a behaviour change and should land as its own
  revertible commit
- The mainnet workflow, on `types: [released]`


.github/workflows/deploy-testnet.yml
```diff
@@ -35,6 +39,9 @@ jobs:
     name: Build and deploy to S3 + CloudFront
     runs-on: ubuntu-latest
 
+    # `published` fires for full releases too, and those are the mainnet channel.
+    if: ${{ github.event_name != 'release' || github.event.release.prerelease }}
```
jeffesquivels: Since this should only run on a prerelease or a manual trigger, we could make this closed to those cases instead of open to anything that is not a release, like this:

`if: ${{ github.event_name == 'workflow_dispatch' || github.event.release.prerelease }}`

.github/scripts/verify-checks.js
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * The check runs that must have passed, by the name GitHub reports them under: a check run is named
+ * after the job that produced it, so this is complete-check.yml's `build` job. Rename that job and
+ * this has to follow -- which is the point, since a check nobody can find is a check nobody runs.
+ */
+const REQUIRED = ['build'];
+
+/** What a check run may conclude with and still count as passed. */
+const ACCEPTED = new Set(['success', 'skipped', 'neutral']);
```
jeffesquivels: Does it make sense to have 'skipped' here?

.github/scripts/verify-checks.js
```diff
@@ -0,0 +1,132 @@
...
+ * @returns {Promise<Map<string, any>>}
+ */
+async function latestCheckRuns() {
+    const response = await api(`repos/${REPOSITORY}/commits/${SHA}/check-runs?per_page=100`);
+
+    /** @type {Map<string, any>} */
+    const latest = new Map();
+    for (const run of response.check_runs || []) {
+        const previous = latest.get(run.name);
+        if (!previous || (run.started_at || '') >= (previous.started_at || '')) latest.set(run.name, run);
```
jeffesquivels: IIUC, run IDs are monotonically increasing, which means we could simplify the checks here by using them, like this: `if (!previous || run.id > previous.id) latest.set(run.name, run);`

.github/scripts/verify-checks.js
```diff
@@ -0,0 +1,132 @@
...
+    let runs = new Map();
+
+    do {
+        // eslint-disable-next-line no-await-in-loop
+        runs = await latestCheckRuns();
+        const unfinished = unfinishedChecks(runs);
+        if (!unfinished.length) break;
+        console.log(`waiting for ${unfinished.join(', ')} on ${SHA}`);
+        // eslint-disable-next-line no-await-in-loop
+        if (Date.now() < deadline) await sleep(POLL_INTERVAL_MS);
```
jeffesquivels: Do we want to enclose this within a `try {} catch {}` block so that transient API errors don't abort the whole loop?

.github/workflows/deploy-testnet.yml
```diff
@@ -104,8 +104,8 @@ jobs:
             exit 1
           fi
 
-      # For a release event the checkout defaults to the commit the release was cut from, which is
-      # the same commit verify-tag.js resolves the tag to.
+      # A release event checks out the commit the release was cut from, which is also the commit
+      # verify-checks.js judges below.
```
jeffesquivels: I think we should add the comment about `verify-check.js` without removing the part about `verify-tag.js` as it's important to understand how is it that both checks are being performed on the same commit that is checked out the release event.

.github/workflows/deploy-testnet.yml
sisou: Through the `workflow_dispatch` property, the workflow retains the option to be manually triggered on any commit. That's why this is open I think.

Happy to remove the manual trigger, would reduce other things as well.

.github/scripts/verify-checks.js
sisou: No, this makes no sense :D Only `success` should be allowed.

.github/scripts/verify-checks.js
sisou: Ok

.github/scripts/verify-checks.js
sisou: Yes, but with a maximum retry count in case of errors.

.github/workflows/deploy-testnet.yml
sisou: In commit 80a2d0a96ffb92eca26717e635265e075b0d4f35 this is changed to `github.event_name == 'workflow_dispatch' || github.event.release.prerelease`, and in commit d1ce1dc74ff36951934134aeab6a0d6cfc7fc8ae I removed the manual dispatch completely (that is the last commit and can be easily dropped if we want to keep manual dispatch).

jeffesquivels: I really like the simplification that occurs when dropping manual dispatch, let's leave it like that and if there's ever a need in the future to have manual dispatch, we can either add it back or, most likely, create a different workflow (to deploy a different dev environment).

Btw, I also like the use of atomic commits for reviewing PRs, I'm glad to see that's how things are done here :muscle:

sisou (approved):

PR #567, opened 2026-09-03
Deployment improvements

jeffesquivels:


tools/deploymentValidator.js
```diff
@@ -59,6 +67,47 @@ const {
...
+    ['default-src', ["'self'", "'unsafe-eval'"]],
+    ['connect-src', ["'self'", 'https://api.coingecko.com']],
+    ['img-src', ['http:', 'https:', 'blob:', 'data:']],
+    ['child-src', ["'self'", 'blob:']],
+    ['worker-src', ["'self'", 'blob:']],
+    // `'self'`, not `'none'`: src/lib/Key.js frames /lib/rsa/sandboxed/RSAKeysIframe.html to derive
+    // the RSA key sign-multisig-transaction needs. A header CSP is enforced alongside the page's
+    // meta CSP rather than in place of it, so `'none'` here vetoes what the page permits and the
+    // iframe never loads -- the request hangs rather than failing.
+    ['frame-src', ["'self'"]],
```
sisou, line 88: Would it be possible to configure `frame-src: none` for all pages, but only for two paths overwrite it with `frame-src: self`?

